Issue #398, opened 2026-07-23
E1 — Human-in-the-loop proactive connection requests


gitchrisqueen:
## Why
`invite_to_connect` (`run_automation.py:2572`) fires **reactively only** (profile viewers). No proactive ICP-targeted connect flow. 2026 penalizes mass automation — keep this approval-gated and capped.

## Scope
- [ ] Approval-gated, daily-capped proactive connect flow driven by an ICP/target list.
- [ ] Reuse `invite_to_connect`; respect rate-limit/kill-switch (`rate_limit.py`). **No volume prospecting.**

## Files
`app/run_automation.py` (~L2572), `utilities/db.py` (target list), `api/main.py`, UI.

## Acceptance
Requests only send after approval and within the daily cap; throttle/kill-switch honored.

---
_Part of the LEM 2026 Growth Roadmap (gap analysis vs. 2026 LinkedIn algorithm authorities). Follows repo norms: feature branch `feature/claude-<task>`, ≥80% patch coverage, migrations advance from V56, CI gates green before merge._

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #515, opened 2026-07-25
feat(outreach): smart connection targeting from content engagers (closes #486)

### LINKED ISSUE #486: feat(outreach): smart connection targeting from content engagers (extends #398)
## What
**Smart connection targeting** (extends #398): proactively send connection requests to **ICP-fit people who engage with the user's content** — and with adjacent thought-leaders'/competitors' content — i.e. warm 2nd-degree prospects, rather than cold lists.

## Why
The fastest qualified network growth is people already showing interest in the topic: anyone who liked/commented on the user's posts (or on a relevant influencer's post) is a warm, on-topic connection candidate. Growing the network with *qualified* prospects compounds every other feature (more feed reach, more catch-up moments, more leads).

## How
- Source candidates from: reactors/commenters on the user's posts (`post_engagers` + feed scrape) and reactors/commenters on a configurable set of adjacent authors' recent posts.
- ICP-score (title/industry/company vs `focus_topics`/target profile); dedup against existing connections + prior requests (`connection_requests`).
- Send an approval-gated, personalized connect note (reuse #398's `invite_to_connect` + note machinery); per-day cap; 429-safe.
- Accepted connections drop into the catch-up/nurture flows.

## Acceptance
- Candidate sourcing + ICP scoring + approval-gated personalized connect note; extends #398's machinery + `connection_requests` dedup. Unit tests ≥90% patch.

**Risk:** product-decision (outbound at scale) — build, then hold with a Decision Comment.

gitchrisqueen:
## What & why

Extends #398 with the **sourcing** half of proactive connecting: instead of an operator hand-typing prospects, LEM now finds **warm, ICP-fit people who engage with content** and files them into #398's existing approval-gated, daily-capped connect path. Nothing about the send path changes — approval, the combined `max_invites_per_day` cap, and the 429 / kill-switch backoff all still decide what actually goes out.

## Candidate sourcing
- **Own-post engagers** — read from `post_engagers` (already populated by the reply sweep), so this source costs **no scraping**.
- **Adjacent authors' engagers** — commenters on the recent posts of a configurable set of thought-leaders/competitors (`connection_target_authors`), harvested with the same SDUI comment-thread walker (`_comment_items_from_thread`) the reply sweep uses. Per-scan caps: **3 authors × 2 recent posts × 15 engagers**, each author/post best-effort so one dead profile can't lose the rest.

## ICP scoring & dedup (`utilities/connection_targeting.py` — pure, no DB/Selenium/LLM)
- ICP fit reuses `lead_scoring.icp_fit` against the user's `focus_topics` / include topics+keywords; warmth blends **recency + frequency + source weight** (own post `1.0` > adjacent post `0.65`), with diminishing returns per repeat. Score = `0.6·ICP + 0.4·warmth`.
- **The ICP floor applies only to people we have profile facts on.** Most engagers are never scraped and score `ICP_UNKNOWN` (50) — a blanket floor would reject nearly everyone and throw away the warm-engager signal this feature exists to use. They're kept, ranked on warmth, and their `reasons` says *"ICP fit unknown (profile not scraped)"* so the approver sees it.
- Deduped against **every `connection_requests` row the user has ever had, any status** — so a canceled/failed target doesn't come back tomorrow night. Name-only sightings are dropped (nobody to invite without a `/in/` URL).

## Approval-gated personalized note
- The note is grounded in the actual shared context (*"I saw your take on <author>'s post about <topic>"* / *"thanks for engaging with my posts"*), refined into the user's voice via `get_ai_message_refinement`, with a template fallback when the LLM is down, hard-capped at LinkedIn's 300 chars.
- `connection_targeting_mode`: **`off`** | **`suggest`** (default) | **`auto_queue`**. `suggest` always files `pending` drafts — **enabling targeting can never send anything on its own**; `auto_queue` defers to the existing `connection_request_mode`.
- Per-scan budget = `max_invites_per_day` − invites sent today − targets already queued, further capped by `MAX_NEW_CONNECT_TARGETS_PER_SCAN` (default **5**), so sourcing can never build a backlog that spends tomorrow's cap the instant it opens.
- Daily **4:00 AM** beat scan (after the 3:30 lead re-score), dispatched per active user and gated on the 429 breaker / manual pause; runs on the `se_outreach` lane.

## Also
- `connection_requests.source` / `icp_score` / `reasons` provenance, surfaced on each card in the Connections review UI so the approver can see *where this person came from and why*.
- Account → Engagement Targeting: targeting mode, min ICP fit, adjacent-author list.
- Migration `V20260725004908__add_connection_targeting.sql` (timestamp version, additive only).

## Scope note
Candidate sourcing uses **commenters**, not reactors: `post_engagers` only records commenters, and enumerating a post's reaction list means driving the reactions modal — new Selenium surface that can't be grounded headless. Commenting is also the higher-intent signal. Reactor sourcing is a follow-up if wanted. Already-connected people aren't pre-filtered either (we have no connections table) — those simply come back as `failed` from `invite_to_connect_now`, exactly as a hand-added target does today.

## Testing
- `tests/unit/utilities/test_connection_targeting.py` — 31 tests, **100%** of the new module (warmth/source/recency weighting, grouping, ICP floor semantics, note templates + 300-char limit, tz-aware `now`).
- `tests/unit/app/test_connection_targeting_scan.py` — 26 tests (mode gating, budget math, suggest vs auto_queue status, dedup, adjacent harvest + degradation when the session fails, LLM fallback, beat + queue lane).
- `tests/unit/utilities/test_db_connection_targeting.py` — 22 tests (provenance insert, open-request count, dedup keys across all statuses, engager sourcing, prefs defaults/clamps — the V52 "one bad value rolls back the whole prefs row" lesson).
- 3 added to `tests/unit/api/test_engagement_prefs_api.py`. Full unit suite: **3582 passed**. UI `tsc -b` clean.

Closes #486

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -136,6 +136,10 @@ DM_NURTURE_AUTO_APPROVE=false
 # Allow one cheap lem-simple call to classify replies no keyword settled (interested / objection /
 # not-now / disinterest). DM replies are low volume, so this defaults ON.
 DM_NURTURE_INTENT_LLM_ENABLED=true
+# Smart connection targeting (issue #486): hard ceiling on how many NEW connection targets one
+# sourcing scan may file per user, independent of their max_invites_per_day. Keeps sourcing a
+# trickle of high-fit people rather than a queue that looks like list-blasting.
+MAX_NEW_CONNECT_TARGETS_PER_SCAN=5
 
 # --- Ollama (local + cloud model hosting) ---
 OLLAMA_API_KEY=your_ollama_api_key
--- a/compose/local/database/migrations/V20260725004908__add_connection_targeting.sql
+++ b/compose/local/database/migrations/V20260725004908__add_connection_targeting.sql
@@ -1,0 +1,31 @@
+-- Smart connection targeting from content engagers (issue #486, extends #398). #398 gave us the
+-- approval-gated, daily-capped SEND path (connection_requests + invite_to_connect); this adds the
+-- SOURCING half: people who engaged with the user's own posts, and people who engaged with a
+-- configurable set of adjacent authors' recent posts, ICP-scored and deduped before they ever
+-- become a request. Nothing here bypasses #398's gates — every target still lands in
+-- connection_requests and still costs one slot of max_invites_per_day.
+
+-- Provenance on the request itself, so the operator approving in the Connections review UI can see
+-- WHERE this person came from and WHY they scored well. Nullable: rows added by hand (or by #398's
+-- API) carry no source and must keep working unchanged.
+ALTER TABLE connection_requests
+    ADD COLUMN source    VARCHAR(32)  NULL AFTER message,
+    ADD COLUMN icp_score TINYINT UNSIGNED NULL AFTER source,
+    ADD COLUMN reasons   VARCHAR(512) NULL AFTER icp_score;
+
+-- Targeting configuration.
+--   connection_targeting_mode — 'off' disables sourcing entirely; 'suggest' (default) ALWAYS files
+--     candidates as 'pending' drafts needing a human approve, regardless of connection_request_mode;
+--     'auto_queue' defers to connection_request_mode (so auto_approve users get the capped drip).
+--     Default 'suggest' means turning this on can never send outbound on its own.
+--   connection_target_authors — adjacent thought-leaders'/competitors' profile URLs whose recent
+--     posts we harvest engagers from (warm 2nd-degree prospects). Empty = own posts only.
+--   min_connection_icp_score — ICP floor for people we HAVE profile facts on. People we've never
+--     scraped score neutral (lead_scoring.ICP_UNKNOWN) and are judged on engagement instead, since
+--     most engagers are never scraped and a floor would otherwise reject everyone.
+-- VARCHAR (not ENUM) for the mode: adding an ENUM value needs a migration (see CLAUDE.md), and this
+-- list will grow.
+ALTER TABLE engagement_preferences
+    ADD COLUMN connection_targeting_mode VARCHAR(16)      NOT NULL DEFAULT 'suggest',
+    ADD COLUMN connection_target_authors JSON             NULL,
+    ADD COLUMN min_connection_icp_score  TINYINT UNSIGNED NOT NULL DEFAULT 55;
--- a/src/cqc_lem/api/main.py
+++ b/src/cqc_lem/api/main.py
@@ -471,6 +471,10 @@ class EngagementPreferencesRequest(BaseModel):
     max_dms_per_day: int = 20
     max_invites_per_day: int = 10
     connection_request_mode: str = "auto_approve"  # 'auto_approve' (default) | 'pre_review'
+    # Smart connection targeting (issue #486): 'off' | 'suggest' (default) | 'auto_queue'
+    connection_targeting_mode: str = "suggest"
+    connection_target_authors: List[str] = []
+    min_connection_icp_score: int = 55
     default_buyer_stage: Optional[str] = Field(default=None, max_length=_LEN_BUYER_STAGE)
     default_video_quality: str = "standard"
     reply_check_mode: str = "event"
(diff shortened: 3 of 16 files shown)


gitchrisqueen:
Held (`needs-human`, risk: product-decision — this is outbound connection requests at some volume). The code is built, tested and CI-ready; what needs you is **how aggressive the defaults should be** before it ever runs against a real account.

### 1. What should the default posture be for a user who never touches the setting?
- **A. `suggest` (as built)** — sourcing runs, targets land as `pending` drafts in Review → Connections; nothing is ever sent until you click Approve, even for users on `auto_approve`.  ✅ *recommended*
- **B. `auto_queue`** — targets follow the existing Connection approval setting, so `auto_approve` users get up to 5 personalized invites/day sent automatically with no human in the loop.
- **C. `off`** — sourcing is opt-in per user; the feature ships dormant and does nothing until someone enables it in Account settings.

### 2. Should people we've never scraped be suggested at all?
Most engagers have no scraped profile, so they score neutral ICP (50) — a real floor (default 55) would reject nearly all of them.
- **A. Keep them (as built)** — they pass the floor, rank on warmth (they *did* engage with your on-topic content), and their card reads "ICP fit unknown (profile not scraped)".  ✅ *recommended*
- **B. Known-fit only** — suggest only people whose title/company/industry we actually have. Much higher precision, but in practice this yields very few candidates until profile scraping has run broadly.

### 3. How many new targets per user per day?
Today: `MAX_NEW_CONNECT_TARGETS_PER_SCAN=5` per daily scan, inside the existing combined `max_invites_per_day=10` (which reactive profile-viewer invites also spend), and the scan subtracts anything already queued.
- **A. Keep 5/day** — at most ~5 new drafts a day (~35/week), comfortably under LinkedIn's weekly invite ceiling even if every one is approved.  ✅ *recommended*
- **B. Lower to 2/day** — a very slow trickle; easiest to review by hand, slowest network growth.
- **C. Raise to 10/day** — the scan can fill the whole daily invite cap; fastest growth, leaves no cap headroom for reactive profile-viewer invites.

**My recommendation: `1A 2A 3A`.** `suggest` + unknown-fit-kept + 5/day gives you the warm-candidate feed this issue asked for with a human on every send, and you can move to `auto_queue` (or raise the cap) per user once you've seen a week of the notes it actually drafts.

One scope call worth flagging, not a decision unless you disagree: sourcing uses **commenters**, not reactors — `post_engagers` only records commenters and enumerating a reactions modal is new Selenium surface I can't ground headless. Say the word and I'll add reactor sourcing as a follow-up.

gitchrisqueen:
1A 2A 3A

gitchrisqueen:
Addressed your review — thanks for the `1A 2A 3A`.

**Your decisions: all three are already what the PR ships**, so nothing changed to apply them. Verified each against the code rather than the description:

| Your call | Where it lives | Verified |
|---|---|---|
| **1A** — default posture `suggest` | `V20260725004908` column default + `db.py:2703` prefs default + `_target_status_for_mode` (`run_automation.py:3815`) forces `PENDING` for anything that is not `auto_queue` | ✅ enabling targeting alone can never send — even for `auto_approve` users |
| **2A** — keep unknown-ICP engagers | `rank_candidates` (`connection_targeting.py:208`) applies the floor only when `candidate.known_fit` | ✅ unscraped people rank on warmth, card reads "ICP fit unknown (profile not scraped)" |
| **3A** — 5 new targets/day | `MAX_NEW_CONNECT_TARGETS_PER_SCAN=5`, inside the combined `max_invites_per_day` minus sent-today minus already-queued | ✅ |

**Also cleared the two Copilot threads** while I was in here (both were real, both now resolved):
- `rank_candidates(limit=0)` returned the **full** list instead of an empty one — `0` is falsy, so `out[:limit] if limit else out` fell through to "unlimited". Not reachable today (the scan returns early when `budget <= 0`), but exactly the wrong way for that to fail. Now only `limit=None` means unlimited.
- `MAX_NEW_CONNECT_TARGETS_PER_SCAN` was parsed with a bare `int(os.getenv(...))` at import — a typo in `.env` would have crashed every worker on startup rather than degrading. Now falls back to the default of 5 and logs a warning, matching the other env parsing in that file.

Tests: +3 (`limit=0`, `limit=None`, env fallback). Full unit suite **3585 passed** locally.

Handing back to the runner for CI + re-review.

gitchrisqueen:
✅ CI green, Copilot review addressed & resolved — merging (fast path).

Issue #474, opened 2026-07-24
bug(automation): feed commenting posts duplicate comments on the same post (unstable dedup key + no concurrency gate)


gitchrisqueen:
## Summary
Feed commenting (`comment_on_feed_inline`, on **other people's** posts) is posting **two comments on the same post**. On 2026-07-24, user 1's 8 feed comments were actually **4 posts each commented twice**, ~2–3 min apart.

## Evidence (prod, 2026-07-24)
8 `action_type=comment / success` logs 12:01–12:26, all keyed `feedpost://<hash>` (no permalink). Message previews pair up cleanly onto 4 posts:
| post | log ids | times |
|---|---|---|
| "great models can't fix bad data" | 75, 76 | 12:04, 12:06 |
| "users only tap ~5% of Claude" | 77, 78 | 12:11, 12:13 |
| "Health-Aware Visual AI / real-world" | 79, 80 | 12:17, 12:19 |
| "team composition matters" | 81, 82 | 12:23, 12:26 |

The `commented_posts` ledger has **8 distinct `feedpost://` keys** — so the dedup layer believed these were 8 different posts.

## Root cause
The dedup key is a **hash of the post's rendered text**: `_feed_post_key(author, content)` = `sha1(author|content[:200])`. The stable LinkedIn activity URN is not captured — `_post_permalink_from_card` looks only for an `a[href*='/feed/update/']` anchor, which this feed's SDUI does not expose, so **every** post falls back to the content hash.

When the same card re-renders during the scroll loop (a "…see more" toggle, or our just-posted comment/reaction mutating the text node), `content[:200]` changes → the hash changes → a **new key** that evades every dedup layer at once:
- the in-memory `seen` set (keyed on the hash),
- `has_commented_post()` / the `commented_posts` ledger (keyed on the hash),
- the atomic `claim_post_for_comment()` (the `UNIQUE(user_id, post_key)` constraint is fine, but it's protecting an unstable key).

So the post is re-scored, still ranks top, gets a fresh claim under the new key, and is commented again.

## Contributing: no concurrency gate
`automate_commenting` self-requeues (`apply_async(countdown=future_forward)`) **and** is triggered by both the pre-post (~15 min before each scheduled post) and the daily golden-hour beat. Nothing prevents two runs for the same user walking the feed concurrently — which, combined with the unstable key, compounds the duplication.

## Fix (this issue)
[...]

PR #582, opened 2026-07-25
fix(automation): key feed comments on the activity URN from the card's ancestors (closes #580)

### LINKED ISSUE #580: bug(automation): feed commenting duplicates RECURRED — URN extraction fails on live feed cards (feedpost fallback)
## Problem (HIGH — recurrence of #474)
Feed commenting posted **duplicate comments on the same post again** for user 1 on 2026-07-25: 8 comments = **4 posts each commented twice** ~2–3 min apart (ids 89/90, 91/92, 93/94, 95/96).

## Why the #474 fix didn't hold
#474 added URN-based stable keys (`_stable_feed_post_key` → `feedurn://urn:li:...`) with a content-hash fallback. But **every comment today used a `feedpost://` (content-hash) key** — never `feedurn://` — even though the fix is deployed (prod v0.83.0). So on the LIVE feed cards, BOTH `_post_permalink_from_card` (looks for `a[href*='/feed/update/']`) and `_feed_post_urn_from_card` (regex `urn:li:(activity|ugcPost|share):\d+` over `card.get_attribute("outerHTML")`) return `None`, and it falls back to the normalized content hash — which is still unstable enough across the post-comment re-render to mint a second key → a second comment. The `commented_posts` UNIQUE constraint + atomic claim are working; they're just protecting an unstable key.

## Fix (needs live feed-DOM grounding — see note)
The activity URN IS present on LinkedIn feed cards (in a `data-*` attribute on the feed-update container or a nearby ancestor), just not where `_feed_post_urn_from_card` currently looks (the `card` returned by `_card_for_textbox` may not include the URN-bearing ancestor). Make URN capture on the feed reliable:
- Search a **higher ancestor** / the full feed-update container's HTML and its `data-urn`/`data-id`/`data-finite-scroll-hotkey-item` attributes for `urn:li:(activity|ugcPost|share):\d+`, not just the comment-button ancestor's `outerHTML`.
- If still no URN, make the content-hash fallback genuinely re-render-stable (the current normalization isn't enough — the two passes still differ; consider hashing only the post author + a longer normalized body slice captured BEFORE we react/comment, and add a per-run in-memory fingerprint of every post we've acted on so a re-render can't re-key it).
- Add a log line recording which key TYPE was used (urn vs permalink vs hash) so we can confirm in PostHog that feedurn:// is actually being used post-fix.

**NOTE:** this requires a **supervised live feed-DOM investigation** (like the #478 comment-thread work) to find exactly where the URN lives on a real feed card — a headless run can't reach the authenticated feed. The owner/maintainer is running diagnostic scripts against the live session to locate it; the fix should use those findings.

## Acceptance
- Feed comments key on the stable activity URN (`feedurn://`) on the live feed — verified in PostHog/logs that new comments use urn keys, not `feedpost://`; no same-post duplicates in a live run. Unit tests for the widened URN extraction + the stronger fallback ≥90% patch; live-validated on user 1's account.

gitchrisqueen:
## What & why

Feed commenting posted **duplicate comments on the same post again** for user 1 (4 posts × 2 comments, 2026-07-25) — a recurrence of #474. Every comment that day used a `feedpost://` content-hash key, never the `feedurn://` key #474 introduced, so the `commented_posts` UNIQUE constraint and the atomic claim were protecting an unstable key.

Two root causes, both fixed in `src/cqc_lem/app/run_automation.py`:

### 1. The URN was never found on live feed cards
`_feed_post_urn_from_card` regexed only the `outerHTML` of the card returned by `_card_for_textbox` — the *nearest* ancestor holding the Comment button. The activity URN lives in a `data-*` attribute on the **feed-update container above that**, so the scan came back empty every time.

The scan now (in this order):
1. the card's own attribute values,
2. each **ancestor's own attribute values**, walking up — and stopping at the first ancestor that contains more than one `button[aria-label='Comment']`, so a **sibling post's URN can never be picked up**,
3. descendant attribute values (containers outrank children: a reshare embeds the *original* post's URN),
4. the previous `outerHTML` regex as a last resort.

It runs as one JS pass through the driver (`_URN_SCAN_JS`); `_feed_post_urn_from_card(card, driver=None)` falls back to `card.parent` when no driver is passed, and any scripting failure degrades to the old HTML regex.

### 2. The content-hash fallback still wasn't re-render-stable
It hashed 240 chars of the normalized body, but the **collapsed card truncates a long post** while the expanded render carries the whole thing → two keys for one post. The fallback now hashes the author plus a **120-char word-boundary prefix** (`_norm_prefix`); LinkedIn truncates well past that, so both renders produce the identical prefix. The per-run in-memory guard was widened from one fingerprint to a **set at 60 and 120 chars** (`_feed_content_fingerprints`), so even a render that truncates earlier than the canonical prefix still matches a post we already acted on — and so does the case where a URN is found on one pass but not the next.

### 3. Key-source telemetry (acceptance criterion)
`_feed_post_identity()` returns `(key, source)` where source is `permalink` / `card` / `hash`. The run records it:
- per posted comment: `log_info("Feed comment keyed by <source> (<key>)", …)`
- run summary `log_info` with the source counts,
- `log_warning` when **any** comment fell back to a hash key (the duplicate-prone path),
- and `key_sources` / `commented_key_sources` on the feed funnel in Redis (already surfaced via `/api` → `feed_reach`; extra JSON fields are additive, no UI change needed).

That's how the live run gets verified: after deploy, a feed run should report `commented_key_sources: {"card": N}` (or `permalink`) and **no** `hash`.

## Testing

- `tests/unit/app/test_feed_dedup_key.py` — widened URN extraction (ancestor scan wins over card HTML, URN extracted+lowercased out of a `urn:li:comment:(…)` wrapper, explicit driver, scan exception / non-string result both fall back to the HTML regex), `_feed_post_identity` sources, truncation-proof fallback key, `_norm_prefix` word-boundary cut, multi-length fingerprints.
- `tests/unit/app/test_comment_dedup.py` — loop-level regressions: a truncated + expanded render of the same post yields **one** comment; an ancestor-scan URN becomes the claim key (`feedurn://…`); the funnel records the key source.
- Full suite: `5056 passed`.

## Note on live validation

The issue flags that pinpointing the exact attribute needs a supervised live-feed DOM session. This change doesn't guess a single selector — it scans **all** attribute values on the card, its ancestors and its descendants for the URN pattern, which is robust to wherever LinkedIn puts it, and the key-source telemetry is what confirms it on the real feed. **The live run on user 1's account is still required** to close the acceptance criteria; if `commented_key_sources` still shows `hash` after this ships, the logged warning plus the funnel will say so immediately.

Closes #580

### DIFF
--- a/src/cqc_lem/app/run_automation.py
+++ b/src/cqc_lem/app/run_automation.py
@@ -506,17 +506,90 @@ def _normalize_post_text(content: str) -> str:
     return t
 
 
+# The URN-less fallback hashes a PREFIX of the body, not all of it: LinkedIn's collapsed card
+# truncates a long post (~3 lines) while the expanded render carries the whole thing, so hashing
+# everything mints TWO keys for one post across a re-render — which is how #474 recurred as #580.
+# Truncation happens well past 120 characters, so a 120-char prefix is byte-identical in both.
+_FEED_KEY_PREFIX_CHARS = 120
+# Extra shorter prefix for the per-run fingerprint set: it still matches when a render truncates
+# even earlier than the canonical prefix (narrow window, aggressive "…more" collapse).
+_FEED_FP_PREFIX_CHARS = (60, _FEED_KEY_PREFIX_CHARS)
+
+
+def _norm_prefix(content: str, limit: int) -> str:
+    """Normalized post text cut to `limit` chars on a word boundary — the same prefix for the
+    collapsed and the expanded render of one post."""
+    norm = _normalize_post_text(content)
+    if len(norm) <= limit:
+        return norm
+    head, sep, _tail = norm[:limit].rpartition(" ")
+    return head if sep else norm[:limit]
+
+
+def _content_digest(author: str, content: str, limit: int) -> str:
+    """Short sha1 over the author + a truncation-proof normalized body prefix."""
+    payload = f"{(author or '').strip().lower()}|{_norm_prefix(content, limit)}"
+    return hashlib.sha1(payload.encode("utf-8", "ignore")).hexdigest()[:20]
+
+
 def _feed_post_key(author: str, content: str) -> str:
-    """Last-resort dedup key when no stable URN/permalink is available. Hashes the NORMALIZED
-    text (see _normalize_post_text) so it's stable across see-more/whitespace re-renders."""
-    norm = _normalize_post_text(content)[:240]
-    digest = hashlib.sha1(f"{(author or '').strip().lower()}|{norm}".encode("utf-8", "ignore")).hexdigest()[:20]
-    return f"feedpost://{digest}"
+    """Last-resort dedup key when no stable URN/permalink is available. Hashes the author plus a
+    NORMALIZED, truncation-proof body prefix (see _norm_prefix) so the collapsed and the expanded
+    render of one post produce ONE key."""
+    return f"feedpost://{_content_digest(author, content, _FEED_KEY_PREFIX_CHARS)}"
 
 
-def _feed_post_urn_from_card(card) -> "str | None":
-    """The canonical urn:li:(activity|ugcPost|share):<id> for a feed card, read from any data
-    attribute or anywhere in the card's HTML. Returns the lowercased URN or None."""
+def _feed_content_fingerprints(author: str, content: str) -> "set[str]":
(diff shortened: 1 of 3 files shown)


gitchrisqueen:
✅ CI green, Copilot review addressed & resolved — merging (fast path).

Issue #617, opened 2026-07-26
G2 — Comment quality contract v2 + comment-side similarity gate


gitchrisqueen:
## Context
Sampled production comments (logs 89–96) open with 'I totally agree that…', 'You raise an excellent point…', 'Spot on—', follow one [validate]+[generic insight]+[question] template, and near-duplicate each other across different posts the same day. LinkedIn's May 2026 crackdown explicitly demotes automated/AI comments out of 'Most Relevant' using semantic sameness + pattern signals. Posts already have an embedding-similarity gate (POST_SIMILARITY_MAX, content_framework.py); comments have nothing.

## Scope
- Comment quality contract enforced at generation review: must (a) reference a SPECIFIC claim from the target post (quote/paraphrase), (b) add exactly one of: own experience, data point, respectful disagreement, or genuine question, (c) be ≥2 sentences, (d) contain zero validation-filler openers — regex-ban list: 'I totally agree', 'You raise an excellent', 'Great post', 'Spot on', 'So insightful', 'Love this', 'Well said' (config-extendable).
- 'Tip #8' framing guidance in the comment prompt: write for the post's READERS, add what the post missed — not flattery for the author.
- Comment-side embedding dedup via lem-embedding: reject a draft whose cosine similarity vs the user's last N=50 posted comments exceeds threshold (env COMMENT_SIMILARITY_MAX, default mirroring POST_SIMILARITY_MAX); store comment embeddings or recompute on the fly from logs.
- Rotate comment structural blueprints (extend content_framework shared menus — do NOT create a parallel comment-only helper).

## Acceptance
- Deterministic tests: filler openers rejected; near-duplicate drafts rejected; contract checklist scored.
- Failed drafts regenerate (bounded retries) then skip the post with a logged reason, never ship a failing comment.

Issue #616, opened 2026-07-26
G1 — Target-creator engagement roster (50/30/20) + on-topic-only commenting


gitchrisqueen:
## Context (docs/engagement-growth-analysis-2026-07.md)
The feed funnel (Redis linkedin:feed_funnel:1, 2026-07-25) examined 21 posts, 0 matched include_topics, and the empty-filter fallback commented on off-ICP posts (AI-in-HR/hiring). engagement_preferences.include_authors has never been populated. Under 2026 Topic Authority ranking, off-topic commenting actively damages distribution. Every fast-growing native creator studied (Alić, Hassid, Hills) grew via high-volume substantive comments on a CURATED list — Hills: 50–100 accounts split 50% peers / 30% ICP / 20% large creators.

## Scope
- New table engagement_targets (migration): user_id, profile_url, name, category ENUM('peer','icp','creator'), max_comments_per_week (default 2), active, last_engaged_at, source ('user'|'suggested').
- Commenting task sources posts from roster authors' recent activity pages FIRST (Selenium: profile → recent posts), home feed second; respect per-author weekly cap (anti-pod: never repeatedly engage the same author).
- Change fallback semantics: when filters empty the funnel, widen filters but NEVER comment on posts failing LLM topic-relevance vs focus_topics. Off-topic = skip, log reason.
- SPA (Account.tsx): roster editor with category + suggested seeds; API CRUD in api/main.py; db.py accessors.
- Extend feed_funnel Redis diagnostics with roster-sourced vs feed-sourced counts.

## Acceptance
- Unit tests for roster selection, per-author cap, on-topic gate. ≥80% patch coverage.
- With an empty roster, behavior degrades to current feed flow minus off-topic fallback commenting.
- Live selector work for author-activity pages goes through the #403/#404 validation pattern (best-effort selectors + fail-closed).

Issue #485, opened 2026-07-24
feat(outreach): DM conversation auto-nurture (approval-gated reply handling)


gitchrisqueen:
## What
**DM conversation auto-nurture**: when a lead replies to one of our DMs (appreciation, profile-viewer, outreach-funnel, catch-up, lead-signal), auto-draft the **context-aware next message** — approval-gated — so opened conversations progress toward a discovery call instead of going cold.

## Why
Top-of-funnel outreach is wasted if replies stall. Most conversions are lost in the gap between "they replied" and "we followed up well." Keeping every live thread warm with a good, human, value-first next message is where outreach turns into pipeline.

## How
- Extend the existing DM inbound handling to detect a **lead reply** to one of our threads, pull the thread context + the lead's profile, and generate a voice-aligned next message via the DM machinery (guest/peer framing, not salesy) — routed through the existing `dm_followups` sequence with intent-aware branching (interested → propose call; objection → address; not-now → light nurture).
- Approval-gated by default; per-day cap; dedup per thread; stops on explicit disinterest.

## Acceptance
- Thread-aware reply detection + drafted next-step DM (approval-gated); reuses `dm_templates`/`dm_followups`/`build_dm_from_template`. Live-validated. Unit tests ≥90% patch.

**Risk:** product-decision (automated DM conversations) — build, then hold with a Decision Comment for owner sign-off on defaults.

Issue #482, opened 2026-07-24
feat(outreach): automate LinkedIn Catch-up (trigger-event congrats → warm BD touchpoints)


gitchrisqueen:
## What
Automate LinkedIn's **Catch-up** feed (`linkedin.com/mynetwork/catch-up/all/`) — the stream of network "moments": **new jobs / promotions, work anniversaries, birthdays, education milestones, "in the news"** — to send **personalized, voice-aligned congratulations** to connections, and route the high-value ones into the existing follow-up / outreach machinery.

## Why this is high-impact for inbound BD/leads
Catch-up moments are the **warmest, lowest-friction touchpoints on LinkedIn** and map directly to classic **trigger-event selling**:
- **New job / promotion** = a connection entering (or newly empowered in) a buying role at a company — a *reason to reconnect* that isn't salesy. This subset alone is the BD goldmine.
- Congratulations get **high reply rates** (people like being acknowledged), which reopens dormant relationships and starts conversations → discovery → leads.
- It's a **steady daily stream** with zero prospecting effort — the network generates moments continuously.
- It reactivates the **long tail of dormant connections** the user never talks to, at scale, warmly.

Our baseline shows **0 DMs and near-zero reciprocity** in the last 14 days — this is exactly the kind of consistent warm outbound that's currently missing. It composes cleanly with what we already have (appreciation DMs, `dm_templates`/`dm_followups`, the outreach funnel #399, connection requests #398).

## How (reuse existing machinery, approval-gated)
1. **Scrape** the catch-up page (Selenium/SDUI — needs a live-grounding pass like the #478 comment thread) → parse each moment: `{person, profile_url, event_type, event_detail}`.
2. **Prioritize / ICP-score** each moment: `new_job`/`promotion` > `work_anniversary` > `education` > `birthday`, weighted by fit to the user's `focus_topics` / target profile (title, industry, company). Skip low-value/low-fit moments so we're not spraying "Congrats!".
3. **Generate** a milestone-specific, voice-aligned message via the existing DM template machinery — new `event_type`s (`job_change`, `promotion`, `work_anniversary`, `birthday`, `education`) on `dm_templates`/`dm_followups` (migration) + `build_dm_from_template`. Reference the specific event; congrats is the opener, not a pitch.
[...]

PR #665, opened 2026-07-26
fix(outreach): activate the network flywheel — clean names, skip 1st-degree, source the funnel, unblock DM nurture (closes #623)

### LINKED ISSUE #623: G8 — Network activation: fix connection requests, feed outreach targets, unblock DM nurture
## Context (the flywheel has no input)
Production audit 2026-07-26: connection_requests has 1 row EVER — it targeted an existing 1st-degree connection (scraped recipient_name 'Harshal Karanpuriya Verified Profile 1st' shows degree-badge text polluting the name and no already-connected check), attempted despite icp_score 50 < min_connection_icp_score 55, and failed. outreach_funnel_targets: 0 rows ever. scheduled_dms: 0 rows ever. post_engagers: 1 row. The entire E1/E2 (#398/#399) outreach layer is idle for lack of inputs and correctness.

## Scope
- Fix name scraping: strip 'Verified Profile', connection-degree suffixes ('1st','2nd','3rd+') and other badge text from scraped names (helper in linkedin/helper.py + tests with real SDUI name strings).
- Skip candidates already 1st-degree (detect degree badge) and enforce min_connection_icp_score BEFORE creating the request row.
- Populate outreach_funnel_targets from: engagement-target roster (G1), post_engagers, and commenters on roster posts — approval-gated per existing connection_request_mode/E2 flow.
- Diagnose why the DM nurture/scheduler path never queues (scheduled_dms empty since V53): trace automate flow → fix the broken precondition; add a log_warning when a nurture/DM path exits early so silence is impossible.
- Surface failure_reason for failed connection requests (column if missing + UI display).

## Acceptance
- Unit tests: name cleaning, 1st-degree skip, ICP threshold, target population, early-exit logging.
- Respect all existing caps (max_invites_per_day, max_dms_per_day) and approval gates.

gitchrisqueen:
## Why

The 2026-07-26 production audit found the entire outbound layer idle, and every symptom traced to a missing input or a silent early exit:

| Table | Rows ever | Why |
|---|---|---|
| `connection_requests` | 1 (failed) | recipient name was `Harshal Karanpuriya Verified Profile 1st` (SDUI badge text scraped into the name), the target was an existing **1st-degree** connection, and it stored `icp_score 50` against a floor of `55` |
| `outreach_funnel_targets` | 0 | #399 shipped a processor; **nothing ever produced a target** — the only way in was an API nobody calls |
| `scheduled_dms` | 0 | the stock DM templates are **step-0 only**, so `enqueue_next_followup` found no step 1 and queued nothing → `dm_followups` stayed empty → `process_user_followups` never ran → no reply was ever read → #485 nurture could never draft |
| `post_engagers` | 1 | — |

## What changed

**Name + degree hygiene** (`utilities/linkedin/helper.py`)
`clean_person_name` / `connection_degree` / `is_first_degree` strip LinkedIn's badge, status and degree text out of a scraped display name and keep the degree as its own signal. Applied at every capture site: the reply sweep, adjacent-post commenter harvest, appreciation DMs, profile-viewer DMs.

**Never invite someone we're already connected to**
New nullable `post_engagers.connection_degree` captures the badge at scrape time. `rank_candidates` drops 1st-degree candidates and `scan_connection_candidates` re-checks before filing. `invite_to_connect_now` also aborts on a 1st-degree profile page rather than hunting for a Connect button that cannot exist. **Fail-open throughout**: a missing badge is *unknown*, not "not connected" — failing closed on a selector drift would stop all outreach.

**ICP floor honoured at file time**
Candidates with no profile facts now file with **no `icp_score`** instead of `ICP_UNKNOWN` (50), so no row ever reads as below the user's own floor; `reasons` still says the fit is unverified. Scored candidates below the floor are skipped before insert.

**`failure_reason`**
New nullable `connection_requests.failure_reason`. `invite_to_connect_now` returns `(sent, reason)`; `send_connection_request` stores it; the Connections review UI shows *"Why it failed: …"* on failed rows and *"ICP fit unverified"* when there's no score.

**Outreach funnel sourcing** — new `scan_outreach_funnel_targets` (daily 04:20 UTC, `se_outreach` lane)
Fills #399's funnel from the three warm surfaces we already track:
- the **engagement roster** (G1 / #616) — the author, paired with their most recent post
- **commenters on roster posts** — paired with that same post
- **`post_engagers`** — start at the *connect* stage (they already engaged with us, so there's no third-party post to comment on first)

Comment-stage drafts come from the same grounded generator + #617 quality/similarity gate the feed uses; a gated draft is left blank for the operator rather than shipping a template comment. Deduped against `connection_requests` *and* existing funnel rows, capped per scan (`MAX_NEW_FUNNEL_TARGETS_PER_SCAN=5`) and by approval-backlog depth (`MAX_OPEN_FUNNEL_TARGETS=25`). Nothing sends: targets file as drafts under the user's existing `connection_request_mode`, and each fired stage still drops back to `pending` for a fresh approval.

**Reply check unblocks nurture**
When a sequence has no next-step template, `enqueue_next_followup` now queues a **reply check** at that step (`DM_REPLY_CHECK_DELAY_HOURS=48`) instead of ending the thread. A reply becomes an approval-gated nurture draft; silence hits the existing "no template for this step" branch and stops the sequence — it can never send an extra DM. This is the shape `_schedule_catchup_followup` already used for catch-up touches (#482), generalized.

**Silence is no longer possible**
Every DM / nurture / targeting early exit now logs *why* (`log_warning` for "the pipeline did nothing", `log_info` for a deliberate per-candidate skip), so an idle flywheel shows up in PostHog instead of returning a string nobody reads.

## Migration

`V20260726074650__add_network_activation_fields.sql` — two nullable columns (`post_engagers.connection_degree`, `connection_requests.failure_reason`). Additive only, no backfill, no data loss.

## Testing

- `tests/unit/utilities/linkedin/test_person_name_cleaning.py` — 22 tests over the real SDUI name shapes, including the exact production value; degree parsing, aria-label extraction, and the fail-open contract.
- `tests/unit/app/test_network_activation.py` — 1st-degree skip (source + send), ICP floor at file time, failure-reason recording, funnel sourcing (all three surfaces, dedup, caps, degraded roster, gated draft), reply-check enqueue, and the early-exit log assertions.
- `tests/unit/utilities/test_db_network_activation.py` — degree COALESCE, failure-reason write/clear/truncate, backlog counter.
- `tests/unit/app/test_dm_followups.py` — the old "no next template → enqueue nothing" test asserted the bug; replaced with the reply-check behaviour plus nurture-off and no-double-check cases.

`poetry run pytest tests/unit -q` → **5797 passed**.

## Notes for review

- The one judgement call: an engager we have **no profile facts** for is still filed (ranked on warmth, `icp_score` NULL) rather than rejected by the floor. Rejecting them would reject nearly every engager — most are never scraped — and throw away the warm-engager signal the feature exists to use. The fix is that such a row no longer *claims* a score below the user's threshold.
- Funnel sourcing reuses `connection_targeting_mode=off` as the "don't source outbound" switch rather than adding a new preference.

Closes #623

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -189,6 +189,15 @@ DM_NURTURE_INTENT_LLM_ENABLED=true
 # sourcing scan may file per user, independent of their max_invites_per_day. Keeps sourcing a
 # trickle of high-fit people rather than a queue that looks like list-blasting.
 MAX_NEW_CONNECT_TARGETS_PER_SCAN=5
+# Network activation (issue #623). The stock DM templates are step-0 only, so a sequence used to end
+# the moment the first DM went out and nobody's reply was ever read. When a step has no next
+# template we now come back after this many hours to look for a reply — a reply becomes an
+# approval-gated nurture draft, silence ends the sequence as before.
+DM_REPLY_CHECK_DELAY_HOURS=48
+# Outreach-funnel sourcing: how many NEW comment->connect->DM targets one scan may file per user,
+# and how deep the approval backlog may get before sourcing stops adding to it.
+MAX_NEW_FUNNEL_TARGETS_PER_SCAN=5
+MAX_OPEN_FUNNEL_TARGETS=25
 
 # --- Ollama (local + cloud model hosting) ---
 OLLAMA_API_KEY=your_ollama_api_key
--- a/compose/local/database/migrations/V20260726074650__add_network_activation_fields.sql
+++ b/compose/local/database/migrations/V20260726074650__add_network_activation_fields.sql
@@ -1,0 +1,18 @@
+-- Network activation (issue #623). The 2026-07-26 production audit found the whole outbound layer
+-- idle: connection_requests had ONE row ever (it targeted an existing 1st-degree connection, with
+-- LinkedIn badge text scraped into the name), outreach_funnel_targets and scheduled_dms had none.
+-- Two columns close the two data gaps behind that.
+
+-- The connection-degree badge LinkedIn renders next to an engager's name, captured at scrape time.
+-- Without it the connection-targeting scan cannot tell a 2nd-degree prospect (worth an invite) from
+-- someone we are ALREADY connected to (an invite that can only fail). VARCHAR, not ENUM: LinkedIn
+-- has renamed these badges before and adding an ENUM value costs a migration (see CLAUDE.md).
+-- NULL means "LinkedIn didn't render a badge" — unknown, NOT "not connected".
+ALTER TABLE post_engagers
+    ADD COLUMN connection_degree VARCHAR(8) NULL AFTER engager_profile_url;
+
+-- Why a send failed, so a 'failed' row in the Connections review UI says something more useful than
+-- FAILED. invite_to_connect_now already produces this text ("Already connected (1st-degree)",
+-- "Failed to find more or connect button: …") — it just had nowhere to live.
+ALTER TABLE connection_requests
+    ADD COLUMN failure_reason VARCHAR(512) NULL AFTER reasons;
--- a/src/cqc_lem/app/my_celery.py
+++ b/src/cqc_lem/app/my_celery.py
@@ -187,6 +187,12 @@ app.conf.update(
             # is actually engaging. Files a few targets a day; sending stays daily-capped + approved.
             'schedule': crontab(hour='4', minute='0')
         },
+        'scan-outreach-funnel-targets': {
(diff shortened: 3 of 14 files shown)


gitchrisqueen:
⚠️ Merging with CI green but WITHOUT a Copilot review — review was requested repeatedly and never arrived (quota exhausted?). Head stable > 3600s. Set REVIEW_FALLBACK=hold in agent-pipeline/config.env to stall instead.

PR #662, opened 2026-07-26
feat(content): 70/20/10 content mix governor + artifact CTAs (closes #618)

gitchrisqueen:
## What & why

The July 2026 audit (#616–630) found nearly every planned post selling the diagnostic conversation. Salesy content loses up to 70% of its reach under the 2026 algorithm, and business-goal-first posting is the owner's own top complaint. This adds the Hills **70/20/10 content-mix governor** and replaces meeting-ask CTAs with **artifact CTAs**.

### Governor (`content_alignment.assign_content_mix` + `run_content_plan.plan_content_for_user`)
- Every planned post gets a class: `value` (70%, audience value, sells nothing) / `authority` (20%, expertise education, sells nothing) / `promo` (10%).
- Deterministic (no LLM), and `offset` continues the rotation from the user's existing post count so a new plan can't restart the promo cadence and land two promo posts back to back.
- Promo cadence is `PROMO_EVERY_N_POSTS`, **clamped to 10–30**. The audit's band is "≤1 per 7–10 posts", but 1-in-7 is ~14% promo — over the 10% ceiling the same audit sets — so the sparse end is the floor and ops can only make promo rarer. A 30-day plan is exactly 21/6/3.
- The promo slot claims a **text** post (its case-study shaping is steered in the text-post prompt; carousels/videos run their own generators) and is forced into the `case_snapshot` blueprint. Its prompt requires client-as-hero, one real outcome number, and explicit no-pressure framing. Falls back to whatever type is left if no text post remains, so the slot is never dropped.

### CTA policy (`ARTIFACT_CTA_POLICY`, one definition in `content_alignment.py`)
- **Prompt-side ban**: injected by `content_framework.cta_policy_directive()` (so newsletters, posts, and comments all inherit it) and into each post's mix directive. Also fixed the prompt bodies that literally asked for a meeting — the video script prompt said "such as scheduling a demo", and thought-leadership said "connect or learn more".
- **Deterministic repair**: `replace_meeting_ask_cta` drops the offending sentences after all LLM passes and closes on an artifact the user actually has — their configured lead magnet (the comment-keyword mechanic the automation already delivers on), else their newsletter, else nothing. It never invents an asset and never doubles an existing keyword ask.
- **Review lint**: a surviving meeting ask HOLDS the post at PENDING through a new `meeting_cta` quality gate — running for *every* post type (carousel/video captions included) and for user-edited drafts via `rescore_post`, with the offending phrase shown in the review queue.
[...]

PR #661, opened 2026-07-26
feat(content): story bank & fact intake — human-sourced specifics as the content anchor (closes #620)

gitchrisqueen:
## What & why

Every credible AI-using creator grounds drafts in their **own** material; LEM had no such source. `profiles.synthesis` (V48) tells a generator how the user *sounds* — nothing told it what the user has actually *done*, so every "personal proof" specific was either lifted from a profile blurb or quietly invented. That is the defining signature of the generic content LinkedIn's 2026 authenticity ranking demotes.

This adds the FACT half of the shared content core.

### Data
- **Migration `V20260726071321__add_story_bank.sql`** — `story_bank`: `kind` ENUM(`anecdote`,`number`,`opinion`,`client_win`,`mistake`,`artifact`), `title`, `body`, `happened_at`, plus the rotation counters `used_count` / `last_used_at` and a soft-retire `active` flag. Timestamp-versioned, additive, FK-cascaded on `users`.
- `utilities/db.py`: `get_story_bank_entries` (LRU order), `count_story_bank_entries`, `upsert_story_bank_entries` (insert new / update by id — never resets the rotation counters), `delete_story_bank_entry`, `record_story_bank_use`.

### Generation
- **New `utilities/ai/story_bank.py`** — one module, all pure functions (no DB, no LLM):
  - `select_story` — relevance first (content-token overlap with the post subject + the user's `focus_topics`, floor `STORY_RELEVANCE_MIN_TOKENS`, default 1), then rotation (least-used, longest-unused, never-used first). Returns `None` for an empty/retired/irrelevant bank.
  - `story_directive` / `no_story_directive` — the entry's facts as **the only permitted personal specifics**, or an explicit "write the industry observation, do NOT invent an anecdote" fallback. The #416 no-fabrication rule stays absolute either way.
  - `unsourced_specifics` — deterministic detector: numbers (digits or spelled, normalized) and named months stated in the draft's **first-person** sentences that appear in no supplied source. Scoped to first-person on purpose, so a stat quoted from the research layer isn't mistaken for a fabricated personal claim.
- `create_text_post` selects one entry per post and threads its directive through all six post generators, the post-type fallbacks and the review-gate retry, so one post stays anchored to one entry and the bank is read once. The use is counted **only when a post actually came out** — a failed generation never burns the anecdote.
[...]

Issue #620, opened 2026-07-26
G5 — Story bank & fact intake: human-sourced specifics as the content anchor


gitchrisqueen:
## Context
Every credible AI-using creator grounds drafts in their OWN material (Hassid: voice-trained tool + manually mined ideas; Koe: Kortex on his corpus; Hills: 'If a stranger could have posted it, delete it'). Ghostwriting agencies enforce a story bank: no draft ships unless anchored to a real anecdote/number/opinion from the client. LEM's posts currently contain zero verifiable personal specifics — the defining feature of the generic content LinkedIn's May 2026 crackdown suppresses. Fully-AI posts measurably lose ~2.8x reach / ~5x engagement.

## Scope
- New table story_bank (migration): user_id, kind ENUM('anecdote','number','opinion','client_win','mistake','artifact'), title, body, happened_at, used_count, last_used_at, active.
- SPA quick-capture UI (Account or a new Story Bank page): add/edit entries, low-friction (a textarea, not a form wizard). API CRUD.
- Generation: text-post and build-receipt prompts must select ≥1 relevant unused-recently entry (rotate by used_count/last_used_at); the entry's facts are the ONLY permitted specifics — absolute no-fabrication rule stays (#416). When the bank is empty or has no relevant entry, fall back to non-story archetypes (industry observation grounded in research layer) instead of inventing experience.
- Onboarding nudge (existing onboarding_nudges pattern) prompting the user to seed 5–10 entries.

## Acceptance
- Unit tests: entry selection/rotation, empty-bank fallback, fabricated-specific detection path.
- profiles.synthesis voice brief unchanged; this feeds FACTS, synthesis feeds VOICE.

PR #671, opened 2026-07-26
fix(review): post-merge adversarial review of #661/#662/#665 (closes #670)

### LINKED ISSUE #670: Post-merge adversarial review of the unreviewed batch: #661, #662, #665
## Context
During the 2026-07-26 Copilot-quota outage, three CI-green PRs merged under the interim warn-merge policy WITHOUT any review (before the Claude adversarial-review gate went live): #661 (story bank & fact intake, migration V20260726071321), #662 (70/20/10 content mix governor, migration V20260726072311), #665 (network activation: name cleaning, connection flow, migration V20260726074650). They are deployed-bound in release 0.99.0.

## Scope
Perform the MODE=selfreview-style adversarial review these PRs skipped, against the merged code on main:
- Review each PR's diff (gh pr diff 661/662/665) plus their integration seams with each other (all three touch run_content_plan/ai_helper/db.py) — the pairwise interactions were never reviewed either.
- Check against the linked issues' acceptance criteria (#620, #618, #623) and CLAUDE.md conventions; hunt for real defects, silent failure modes, migration hazards, and test gaps. Style nits are not findings.
- FIX real findings in this issue's PR (tests included). If something needs a product decision, escalate per RUNBOOK.
- PR body must list per-PR verdicts (PASS or findings+fixes) so the review record exists.

## Acceptance
- A verdict for each of the three PRs; fixes with tests for any findings; full unit suite green.

gitchrisqueen:
Post-merge adversarial review of the three PRs that merged CI-green but unreviewed during the review outage. Every diff was read line-by-line against its issue's acceptance criteria, the merged state of the shared files (`run_content_plan.py`, `ai_helper.py`, `db.py`, `content_alignment.py`) was re-verified, and the integration seams between the three were specifically attacked.

## PR #661 — Story bank & fact intake (#620)

**Verdict: PASS with one finding (fixed).**

Checked: migration (timestamp-versioned, FK + index, no ENUM hazards), API CRUD (session-scoped, 401/422/500 paths, cross-user scoping in `upsert`/`delete`), selection/rotation semantics (never-used first, longest-unused next, active-only, relevance via focus topics with a sane no-signal fallback), empty-bank no-fabrication fallback, use accounting (outermost call only, only when content shipped, DB failure non-fatal), fabrication detector scoping (first-person sentences only, skipped with no allow-list), onboarding nudge ordering (behind VOICE_SET, once, ahead of trial warning, safe with `started_at=None`), and the recursion paths (type fallbacks and review-gate retry reuse the same `story_directive`; the bank is read at most once per outermost call; no double `record_story_bank_use`).

**Finding (fixed):** the review gate's fabrication check treated only the story entry + profile synthesis as legitimate sources. The lead-magnet CTA directive — text *we* hand the writer, which can carry a number in the user's configured resource name ("my 5-step checklist") — was not a source, so a correctly woven CTA triggered a spurious regeneration whose repair directive explicitly steers the model to drop that number, fighting the keyword-listener mechanic that `ensure_lead_magnet_cta` then has to repair. `_fabricated_specifics` now accepts extra sources and the gate passes `lead_magnet_cta` (both first-draft and retry checks). Regression test added.

## PR #662 — 70/20/10 content mix governor (#618)

**Verdict: PASS with one finding (fixed).**

Checked: deterministic assignment (promo exactly 1-in-N with N clamped 10–30 so promo can never exceed the 10% ceiling; authority phased off promo indices; promo wins collisions; offset continues the cadence across plan boundaries), plan wiring (`_take_planned_post_type` claims a text post with a safe fallback, class persisted via `insert_planned_post`, read back for regenerates so the mix can't drift), the meeting-ask detector/repair/gate triple (repair is deterministic, no-artifact users get the ask dropped rather than an invented asset, existing keyword ask never doubled, gate lints ALL post types and holds at PENDING per the acceptance criteria), nullable `posts.content_mix` (unclassified never treated as promo, excluded from ratios), and the analytics endpoint (days clamped before both queries; empty plan is not a violation).

**Finding (fixed):** the bare-noun pattern in `_MEETING_ASK_PATTERNS` (`discovery call`, `strategy session`, `15-minute call` with no ask context) violated the file's own stated precision-over-recall rule: first-person *narrative* — "I ran a discovery call with them last week", frequently the story-bank anecdote itself — was flagged, and `replace_meeting_ask_cta` **deletes** flagged sentences, silently destroying real story content and tripping the `meeting_cta` hold on clean posts. The pattern now requires an offer-verb context (`grab/claim/snag/book/schedule/join/reserve/get/offer(ing)` + article); every previously listed positive still matches (verified by the existing parametrized tests) and three narrative negatives were added.

## PR #665 — Network activation (#623)

**Verdict: PASS, no code changes.**

Checked against all five scope items: name cleaning handles the exact production string plus aria-label, multiline SDUI stacks, screen-reader duplication, and non-name badge text without touching real names; degree detection fails open (missing badge = unknown, never "connected") at all three enforcement points (rank filter, file-time belt-and-braces, profile-page abort before hunting for a Connect button); ICP floor enforced before the request row exists and unscored candidates store NULL instead of the misleading `ICP_UNKNOWN`; `invite_to_connect_now`'s tuple return updated at both call sites with `result` initialized on every path; `failure_reason` written on every status transition (stale reasons cleared) and truncated to the column; funnel sourcing files drafts only — the approval status defers to `connection_request_mode` exactly as the issue specifies, dedups against `connection_requests` and existing funnel rows, respects a per-scan budget and an open-backlog ceiling, and comment drafts route through `generate_ai_response` so the #617 quality contract applies; the reply-check that unblocks the nurture queue cannot loop (the no-template branch in `process_user_followups` stops the row without re-enqueueing, and the nurture event type is excluded from getting a second check); every early exit now logs. Beat entry, queue lane, and 429/pause gating verified.

## Integration verdict

**One real seam defect found and fixed (#618 × #620):** a promo-classed post whose story bank was empty or irrelevant was forced into the `case_snapshot` blueprint *and* given the promo directive demanding "one specific real number or observable result" — while the fabrication detector was skipped entirely (no story = no allow-list) and the no-story fallback simultaneously forbade personal specifics. The post most likely to invent a client figure was the only one shipping with zero deterministic check, and the prompt actively pushed it to invent. Fix: story selection now runs before blueprint selection (no information is lost — `blueprint["subject"]` is always None at that point in this flow), and an ungroundable promo slot is demoted to `value` with a `log_warning` before the case-study blueprint or promo directive can attach. The plan row keeps its `promo` class, which only makes the dashboard's compliance ratio conservative. Recursive calls and the review-gate retry inherit the demoted class.

Other seams verified clean: all ~11 generator call sites thread `story_directive` AND `content_mix` consistently through every fallback/retry path; the mix repair's appended artifact CTA cannot trip the fabrication gate (runs after the review gate; with the #661 fix its numbers are sourced anyway); #665's funnel comment drafts reuse the shared comment core rather than a parallel prompt helper; #665's writes respect `max_invites_per_day` / `max_dms_per_day` and the approval gates (nothing new auto-sends).

`poetry run pytest tests/unit -q`: **5963 passed** (283 in the directly touched areas).

Closes #670

### DIFF
--- a/src/cqc_lem/app/run_content_plan.py
+++ b/src/cqc_lem/app/run_content_plan.py
@@ -1072,15 +1072,18 @@ def _score_and_persist_dwell(user_id: int, post_id: int, content: str) -> Option
 
 
 def _fabricated_specifics(content: str, story: Optional[dict],
-                          profile_synthesis: Optional[str] = None) -> list:
+                          profile_synthesis: Optional[str] = None,
+                          *extra_sources: Optional[str]) -> list:
     """First-person specifics in the draft that appear in NONE of the material we supplied — i.e.
     invented facts about the author (issue #620, enforcing the #416 no-fabrication rule). Only
     meaningful when a story entry was selected: with no entry there is no allow-list, so every
-    number would look fabricated and the check is skipped entirely."""
+    number would look fabricated and the check is skipped entirely. `extra_sources` covers other
+    text we legitimately handed the writer (e.g. the lead-magnet CTA directive, whose keyword or
+    resource name can carry a number) so it is never flagged as invented."""
     if not story or not content:
         return []
     return _story_bank.unsourced_specifics(
-        content, _story_bank.fact_sources(story, profile_synthesis))
+        content, _story_bank.fact_sources(story, profile_synthesis, *extra_sources))
 
 
 def _review_generated_post(user_id: int, stage: str, post_type: str, user_profile: LinkedInProfile,
@@ -1106,7 +1109,7 @@ def _review_generated_post(user_id: int, stage: str, post_type: str, user_profil
     too_similar = score > threshold
     missing_proof = not has_first_person_proof(content)
     proof_regen = missing_proof and _proof_regen_enabled()
-    fabricated = _fabricated_specifics(content, story, profile_synthesis)
+    fabricated = _fabricated_specifics(content, story, profile_synthesis, lead_magnet_cta)
     fabrication_regen = bool(fabricated) and _fabrication_regen_enabled()
     # No-fabrication guard (issue #619 / G4) — only on the archetypes whose value IS the specifics.
     # Checked against the user's verified story bank (#620 / G5), so a real number out of their own
@@ -1171,7 +1174,7 @@ def _review_generated_post(user_id: int, stage: str, post_type: str, user_profil
         log_warning("Post still lacks a concrete first-person lived detail after retry "
                     "(A2 proof slot); keeping second attempt",
                     user_id=user_id, post_id=post_id, task_name="create_text_post")
-    still_fabricated = _fabricated_specifics(second, story, profile_synthesis)
+    still_fabricated = _fabricated_specifics(second, story, profile_synthesis, lead_magnet_cta)
     if still_fabricated:
         log_warning(f"Post still states first-person specifics not in the story bank after retry "
                     f"({', '.join(still_fabricated)}); keeping second attempt",
@@ -1259,6 +1262,36 @@ def create_text_post(user_id: int, stage: str, post_type: str = None, user_profi
     if history_directive is None:
         history_directive = history_avoidance_directive(recent_texts)
 
+    # Story bank (issue #620): anchor the post to ONE thing the user actually did, and make that
+    # entry the ONLY personal specific the writer may state. An empty bank — or a bank with nothing
+    # related to the user's focus topics — ships the explicit no-fabrication fallback instead, so a
+    # missing story degrades to an industry observation rather than an invented anecdote. Recursive
+    # calls (type fallbacks / the review-gate retry) receive the directive so the whole post stays
+    # anchored to the same entry and the bank is read at most once per outermost call. Selected
+    # BEFORE the blueprint so the promo demotion below can still steer shape selection.
+    story = None
+    story_selected_here = story_directive is None
+    if story_selected_here:
+        story = _select_story_for_post(user_id, prefs, blueprint)
+        story_directive = _story_bank.story_directive(story)
+        if story:
+            myprint(f"Story bank anchor for post_id={post_id}: "
+                    f"[{story.get('kind')}] {story.get('title')}")
+        else:
+            myprint("No story bank entry available — writing a non-story archetype")
+
+    # Integration seam (#618 x #620): the promo slot demands a case study built on ONE real outcome
+    # number, but without a story-bank anchor the fabrication detector has no allow-list and is
+    # skipped entirely — so the one post most likely to invent a client figure would ship unchecked.
+    # A promo post the bank cannot ground is demoted to audience-value content instead (and NOT
+    # forced into the case_snapshot blueprint below). The plan row keeps its 'promo' class, which
+    # only makes the dashboard's mix-compliance ratio conservative.
+    if story_selected_here and story is None and content_mix == ContentMix.PROMO.value:
+        log_warning("Promo slot has no story-bank anchor — demoting this post to 'value' so a "
+                    "case study cannot be invented", user_id=user_id, post_id=post_id,
+                    task_name="create_text_post")
+        content_mix = ContentMix.VALUE.value
+
     # ONE assigned SHAPE per post from the SHARED framework core (content_framework): rotate away
     # from this user's recently used post archetypes/hook styles (V51 history) exactly the way
     # newsletter editions rotate — chosen in code, no extra LLM call.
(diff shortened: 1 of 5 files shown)


gitchrisqueen:

Reviewed the full diff line-by-line against #670's scope, plus the surrounding code it changes (`create_text_post` ordering, `_review_generated_post`, `story_bank.fact_sources`/`specific_tokens`, `_MEETING_ASK_PATTERNS`, `replace_meeting_ask_cta`) and every `create_text_post` call site.

- **Meeting-ask detector lost recall the #618 gate depends on (fixed, `content_alignment.py`):** narrowing the bare-noun pattern to require an offer verb + article was the right call — but it overshot and stopped matching CTA shapes that WERE flagged before it. `an` is absent from every article set while the adjective list carries `intro(?:ductory)?`, which only ever takes `an`, so `Book an intro call` / `Schedule an introductory call` / `Grab an intro session` became unmatchable — a meeting ask that now ships APPROVED instead of holding at PENDING. The no-verb framings went the same way: `Want a free strategy session?`, `Interested in a discovery call?`, `Free discovery call for the first 5 people who reply.`, `Free strategy session — link in comments.` Added `an` to the offer-verb + `book`/`schedule` article sets and two patterns that restore the no-verb forms without reopening the narrative hole: the interest form requires an explicit interest verb (past-tense narrative `I wanted a quick call` cannot match), the headline form requires an unambiguous booking marker inside the same sentence. Every narrative negative stays clean — including two new adversarial ones the PR's own rule implies but didn't cover, `I book discovery calls with new clients every Friday` and `that discovery call opened my eyes` — and the `LEAD_MAGNET_CTA_REPAIR_MENU` / newsletter repair lines still don't self-trigger. 9 positives + 3 negatives added to the parametrized lists.

Verified clean, no change needed:
- **Story-selection reorder is behavior-neutral for relevance.** `_select_story_for_post` reads `blueprint["subject"]`, and the PR's claim that it is always `None` here holds: `select_blueprint("post", …)` is called without `subject=`, and no caller of `create_text_post` passes a `blueprint` at all except its own recursion (which also passes `story_directive`, so `story_selected_here` is False there). Relevance still falls back to `focus_topics`.
- **Promo demotion.** Gated on `story_selected_here`, so it fires once in the outermost call; the demoted class rides into the generators, the review gate, both type-fallback recursions and the retry, and `guidance="case_snapshot"` is correctly skipped. The DB row keeping `promo` overstates promo, which only makes `content_mix_compliance`'s `promo_ratio <= 10%` check stricter. `log_warning` per CLAUDE.md.
- **Fabrication allow-list widening is bounded.** `fact_sources` already took `*extra`; `specific_tokens` extracts only digit runs, spelled numbers and month names, so the CTA directive contributes the user's own configured resource text plus `one`→`1` from its prose — not a general escape hatch. `*extra_sources` after a defaulted param is valid and both call sites (first draft + retry) pass it.
- **`meeting_ask_excerpts`** still gets plain strings from `findall` — every added group is non-capturing.

`tests/unit`: **5973 passed**.

gitchrisqueen:

Second pass. The previous marker (15:27Z) predates head commit `60f0769` (15:41Z), so this review targets that commit — the meeting-ask recall restoration — plus a re-verification of the rest of the diff.

- **Restoring recall reopened the hole the commit before it closed (fixed in `32ee36c`):** both verb-less patterns added in `60f0769` match anywhere in a sentence, and `replace_meeting_ask_cta` **deletes** every sentence the detector matches. Confirmed against the merged head, six ordinary-prose sentences were flagged and would have been silently destroyed: `Most buyers want a quick call before they commit.` / `Clients want a free consultation before they'll even look at pricing.` / `They're ready for a strategy session long before we are.` / `My team was up for a quick call but the client ghosted us.` / `That quick call led to a sign up the same day.` / `After our discovery call I told them to DM me anytime.` The pattern comment asserted narrative "uses the past tense" and therefore cannot match — true only in the first person; third-person present uses the identical words. The last two are a noun phrase inside a clause that happens to land within the 40-character marker window. Fix: both patterns now require the ask to be ADDRESSED — it opens the sentence (start-of-string or the end of the previous sentence, so headlines/imperatives still fire mid-post) or names the reader as `you`. A meeting ask is addressed by definition; narrative about a call is not. Every positive `60f0769` added still matches; four mid-post placements and six narrative negatives added as regression tests. `meeting_ask_excerpts` strips the punctuation the anchor now pulls into the match, since that excerpt is shown to a human in the review-queue finding.

Re-verified clean (no changes): `_fabricated_specifics`' `*extra_sources` (`fact_sources` filters falsy, so a `None` CTA is safe; the directive's only variable text is the user's own 240-char resource description, so the allow-list widening is exactly the intended one); the story-before-blueprint reorder (`_select_post_blueprint` never passes `subject`, so `blueprint["subject"]` is provably `None` at the new call site — no relevance signal is lost); `story_selected_here` keyed on `is None` rather than truthiness, so the empty-bank fallback directive still suppresses re-selection in recursive calls; the promo demotion running strictly before both the `case_snapshot` guidance and `alignment_directive`, with the review-gate retry inheriting the demoted class; and all seven `create_text_post` call sites threading `blueprint` and `story_directive` together, so no caller can reach the reorder with a half-populated pair.

`poetry run pytest tests/unit -q`: **6077 passed**.

gitchrisqueen:

Issue #403, opened 2026-07-23
F2 — Validate unverified Selenium selectors before building on them


gitchrisqueen:
## Scope
- [ ] Live-validate `auto_post_to_group` group-composer selectors (`run_automation.py:1484`, flagged 'best-effort').
- [ ] E2E-validate `check_dm_replied` (`run_automation.py:1937`) before trusting it to suppress follow-ups.

## Files
`app/run_automation.py`. Use selenium-lem / playwright MCP.

## Acceptance
Both paths confirmed against live LinkedIn SDUI; selectors updated where drifted.

---
_Part of the LEM 2026 Growth Roadmap (gap analysis vs. 2026 LinkedIn algorithm authorities). Follows repo norms: feature branch `feature/claude-<task>`, ≥80% patch coverage, migrations advance from V56, CI gates green before merge._

🤖 Generated with [Claude Code](https://claude.com/claude-code)

gitchrisqueen:
## 🧑‍⚖️ Human decision needed — reply with option letters
Held (`needs-human`, risk: live-linkedin). No implementation attempted this tick.

This issue's acceptance criteria is "confirmed against live LinkedIn SDUI" for two paths:
- `auto_post_to_group` (`app/run_automation.py:2202`) — opens a group's share box (`Start a post` button via XPath text/aria-label match), types into a `div[role='textbox']`, clicks a `Post` button. Docstring already says "Best-effort — the group composer selectors are validated in the live pass."
- `check_dm_replied` (`app/run_automation.py:3704`) — navigates to a profile, clicks `Message`, reads the last message sender via `_LAST_SENDER_JS`, compares against `my_name`. Docstring already says "DOM structure confirmed live; the full profile->Message->detect flow (and self-name match) still needs E2E validation before this is trusted to suppress follow-ups."

Both are exactly what they say: unverified against a real, authenticated LinkedIn session. This pipeline tick runs headless in a dedicated worktree with no LinkedIn credentials, no Selenium/docker access (ground rules forbid running docker), and no `.mcp.json` in this worktree — the `selenium-lem`/`playwright` MCP servers referenced in the issue's "Files" note are only wired up for an interactive Claude Code session on the VPS (pre-approved via `.claude/settings.local.json`, but MCP tools load at session start, so a headless pipeline run never has them). I can't produce a real "confirmed against live LinkedIn SDUI" result from here — only claim it, which would be worse than leaving the docstrings as-is.

[...]

Issue #395, opened 2026-07-23
D1 — Engagement-rate analytics dashboard


gitchrisqueen:
## Why
Dashboard shows counts + a best-times list only — no engagement-rate trends, per-post drill-down, or format/hook leaderboards. No visibility into what works.

## Scope
- [ ] Extend `ui/src/pages/Dashboard.tsx`: per-post performance table, engagement-rate/impression trends, follower growth, format/hook leaderboards (from B1–B3).
- [ ] Backend endpoints for the new aggregates (`api/main.py`).
- [ ] Use the `dataviz` skill for chart design.

## Files
`ui/src/pages/Dashboard.tsx`, `api/main.py`, `utilities/db.py`, `utilities/post_stats.py`.

## Acceptance
Dashboard renders real engagement-rate trends + a hook/format leaderboard from `post_stats`. Depends on B1–B3.

---
_Part of the LEM 2026 Growth Roadmap (gap analysis vs. 2026 LinkedIn algorithm authorities). Follows repo norms: feature branch `feature/claude-<task>`, ≥80% patch coverage, migrations advance from V56, CI gates green before merge._

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #673, opened 2026-07-26
feat(analytics): follower & audience telemetry with daily capture (closes #627)

### LINKED ISSUE #627: D3 — Follower & audience telemetry: daily follower/profile-view tracking
## Context
LEM tracks post stats, newsletter subscribers, and engagers — but NOT the user's follower count or profile views. Audience growth is the primary outcome of the whole system and it is currently invisible. Lara Acosta's operating metric is profile visits, not likes. Also: perf-tracking snapshot sums ALL post_stats rows per post (repeated scrapes inflate cumulative impressions).

## Scope
- New table follower_stats (migration): user_id, captured_at, follower_count, connection_count, profile_views (nullable — own-profile analytics surface), search_appearances (nullable).
- Daily Selenium capture task (celery beat) via the user's session: own profile follower/connection counts; profile-view count from the analytics page if accessible. Best-effort fail-closed: a missing anchor logs and skips, never blocks other automation. Selector work follows the #403/#404 validation pattern.
- db.py accessors + API endpoint + a growth panel in the analytics dashboard (#395 surface): follower delta 7/30-day, profile views, overlaid with posting/commenting activity.
- PostHog event per capture for cross-referencing with content outcomes.
- Fix the audit's measurement bug: anywhere cumulative impressions are computed (incl. scripts/perf snapshot guidance), use latest-row-per-post (MAX(id) GROUP BY post_id pattern already in db.py).

## Acceptance
- Unit tests with mocked driver + DB; migration follows timestamp/version rules (check current max V-number first).

gitchrisqueen:
## What & why

Audience growth is the primary outcome of the whole system and it was **invisible**. LEM tracked post stats, newsletter subscribers and engagers — but never the user's own follower count or profile views. (Lara Acosta's operating metric is profile visits, not likes.)

### Follower telemetry
- **Migration `V20260726160255__add_follower_stats.sql`** — `follower_stats`: one row per capture with `follower_count`, `connection_count`, `profile_views`, `search_appearances`. Every count is **NULLABLE on purpose**: a missing anchor records "not measured", never a `0` that would read as a collapsed audience in a growth delta.
- **`utilities/audience_stats.py`** (pure, no DB/Selenium — the #403/#404 pattern: thin browser layer, unit-tested parsing):
  - LinkedIn label parsing (`1,234 followers`, `3.2K followers`, `500+ connections`, the stacked `48\nProfile views` analytics card). Precedence is number-**before**-label (how LinkedIn renders it everywhere), bounded to at most one line break so another card's number can't silently bind to this label.
  - `follower_growth()` — daily series + 7/30-day deltas. A delta stays `None` until a baseline snapshot old enough to measure against exists, so a two-day history can't report explosive growth. Same-day re-captures collapse to the latest row (a re-run is a correction, not another data point).
- **`capture_follower_stats`** Celery task on the `se_content` lane, dispatched nightly at **23:40** by `auto_capture_follower_stats` — after the 23:00 post-stats scrape, so a night's audience snapshot and its content outcomes land together. Throttle-breaker aware and session-gated. **Fail-closed per signal**: a LinkedIn DOM change leaves that count NULL, logs, and never raises or blocks other automation. An all-NULL snapshot is not written at all.
- **`db.py`**: `record_follower_stat`, `get_follower_stats`, `get_daily_action_counts` (the activity overlay).
- **`GET /user/audience-growth`** + an **Audience Growth panel** on the Dashboard (#395 surface): followers, 7/30-day change, profile views, connections, and two aligned single-series charts — followers over time and daily posting/commenting/DM activity over the same day window (never a dual axis, per the existing chart convention).
- **PostHog `audience_snapshot` event** per capture, so follower growth is queryable next to the content outcomes that drove it.

### Measurement bug fix
`scripts/perf_snapshot.sh` summed **every** `post_stats` row per post. `post_stats` is append-only and each nightly scrape writes a fresh *cumulative* row, so cumulative reactions/comments/impressions were multiplied by how many times each post had been scraped. Now restricted to `id IN (SELECT MAX(id) … GROUP BY post_id)` — the same latest-row-per-post pattern every analytics read in `db.py` already uses. (`db.py` itself was already correct everywhere; the script was the only offender.)

## Testing notes

- `tests/unit/utilities/test_audience_stats.py` — 28 tests: label shapes, the cross-wiring guard (a number two line breaks up must not bind), NULL-not-zero, same-day collapse, baseline/window delta semantics, negative deltas, activity bucketing.
- `tests/unit/app/test_follower_capture.py` — 25 tests with mocked driver + DB: full snapshot read, missing anchor → all-None, no profile URL, conditional second analytics page, navigation failure swallowed, `<body>` fallback, task record/track/skip paths, driver always quit, dispatcher session-gating + throttle, beat registration, and every db accessor incl. error paths.
- `tests/unit/api/test_api_coverage_misc.py::TestAudienceGrowth` — endpoint shape, the wider baseline read (`days + 30`), the `days` clamp, and 401.
- `tests/unit/app/test_selenium_queue_routing.py` — the new task's `se_content` lane.
- Full local suite: **6190 passed**. SPA typechecks clean (the only `tsc` errors in this sandbox come from a stale `node_modules` missing `vitest`, unrelated to this diff).

The LinkedIn selectors are best-effort and **unvalidated against a live session** — they are registered as such in `docs/LIVE_VALIDATION_FORMAT_AND_STATS.md`'s selector map, and the fail-closed design means an unread count is a NULL in the series rather than a broken run. The first real nightly capture is worth eyeballing.

Closes #627

### DIFF
--- a/compose/local/database/migrations/V20260726160255__add_follower_stats.sql
+++ b/compose/local/database/migrations/V20260726160255__add_follower_stats.sql
@@ -1,0 +1,19 @@
+-- Follower & audience telemetry (issue #627). LEM tracked post stats, newsletter subscribers and
+-- engagers, but never the AUDIENCE itself — follower growth and profile views are the primary
+-- outcome of the whole system and were invisible.
+--
+-- One row per daily capture run. Every count is NULLABLE on purpose: the capture is best-effort
+-- Selenium against LinkedIn's SDUI, and a missing anchor must record "not measured" rather than a
+-- zero that would read as "lost all followers" in the growth deltas. profile_views and
+-- search_appearances only exist on the author's own analytics surface, so they are frequently NULL.
+CREATE TABLE IF NOT EXISTS follower_stats (
+    id                 INT AUTO_INCREMENT PRIMARY KEY,
+    user_id            INT NOT NULL,
+    follower_count     INT NULL,   -- "N followers" on the user's own profile; NULL when unreadable
+    connection_count   INT NULL,   -- "N connections"; NULL when unreadable ("500+" reads as 500)
+    profile_views      INT NULL,   -- own-profile analytics; NULL when the surface isn't accessible
+    search_appearances INT NULL,   -- own-profile analytics; NULL when the surface isn't accessible
+    captured_at        DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    KEY idx_user_captured (user_id, captured_at),
+    FOREIGN KEY (user_id) REFERENCES users(id) ON DELETE CASCADE
+);
--- a/docs/LIVE_VALIDATION_FORMAT_AND_STATS.md
+++ b/docs/LIVE_VALIDATION_FORMAT_AND_STATS.md
@@ -99,6 +99,8 @@ Nothing below is a guess; the provenance column says where each came from.
 | Reactions/comments/reposts/saves rows | line pair `<Label>` → `<count>` (label-first) | `_stacked_counts` / `_STACKED_LABEL_FIRST` | live grab 2026-07-23 |
 | Social-bar counts (detail view) | regex `<count> <label>` on the card text | `_COMMENTS_RE` … `_SAVES_RE` | in-repo, live-used |
 | Post permalink on a feed card | `a[href*='/feed/update/']` | `_post_permalink_from_card` | in-repo, live-used |
+| Follower / connection counts | page text of `main` on the user's own profile → `parse_follower_count` / `parse_connection_count` | `capture_audience_snapshot` (#627) | in-repo, **best-effort / unvalidated** — first real run should be supervised |
+| Profile views / search appearances | `https://www.linkedin.com/analytics/profile-views/` (falls back to `/analytics/search-appearances/`), value-first line pair | `capture_audience_snapshot` (#627) | in-repo, **best-effort / unvalidated** |
 | Feed composer trigger | `//button[contains(normalize-space(),'Start a post') or contains(@aria-label,'Start a post') or contains(@aria-label,'Create a post')]` | `auto_post_to_group`, `probe_composer` | in-repo, **best-effort / unvalidated** (F2 #403) |
 | Document upload control | — | — | **unmapped by design** (§1); `--probe-composer` captures it |
 | Document media card | — | — | **unmapped**; `probe_document_render` captures it |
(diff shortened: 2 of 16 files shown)


gitchrisqueen:

- **Label parsing cross-wired metrics on a stacked label-first page** (`utilities/audience_stats.py`): the value-first pattern only refused to span >1 line break, which doesn't help when the neighbouring card is adjacent. `"Profile views\n288\nSearch appearances\n88"` returned **288 for both**, and `"Followers\n4,312\nConnections\n500"` — the label-first shape the module explicitly supports and tests — recorded the follower count as the **connection** count. Silently wrong numbers are exactly what the NULL-not-zero design exists to avoid, and they'd be indistinguishable in the growth series. Fixed with `_claimed_by_another_label`: a number that is already the value of a *different* audience label sitting in front of it is skipped, unless that label already carries its own number (so `"4,312 followers\n500+ connections"` still yields 500). When every value-first candidate is claimed it falls through to the label-first read, and returns None rather than a wrong count. Two tests added covering both directions.
- **`_read_page_text` never reached the `<body>` fallback on an EMPTY `<main>`** (`app/run_automation.py`): `return el.text or ""` returned `""` for a half-hydrated shell, so the documented fallback only ran when `main` was *missing*, and the whole snapshot came back all-NULL. Now falls through unless the text is non-blank; test added.
- **Delta tiles labelled a change "7-day" regardless of how old the baseline was** (`ui/.../Dashboard.tsx`): `_on_or_before` takes the newest capture on **or before** the window edge, so after a run of failed captures the "7-day change" can really span 25 days, and the 7- and 30-day tiles can show the same number. The payload already carried `from_date` but the UI dropped it — each tile now names the date it was measured from ("since Jul 1" / "not enough history").

Verified: 6193 unit tests pass, SPA `tsc --noEmit` clean. Also checked and found no issue with: the `follower_stats` migration (timestamp version, nullable-by-design counts, FK/index), `record_follower_stat`'s all-NULL short-circuit, the parameterised `get_daily_action_counts` IN-list, the `se_content` lane routing + beat registration, the throttle/session gating in the dispatcher, driver cleanup on every task path, and the `perf_snapshot.sh` MAX(id)-per-post fix (which matches `db.py`'s existing latest-row pattern).

gitchrisqueen:

Issue #547, opened 2026-07-25
fix(engagement): reliable pre-post feed-commenting window + per-post observability


gitchrisqueen:
## What
Make the **pre-post feed-commenting reliably fire in the intended ~15-min window before each scheduled post** (the engagement-hacking intent: be active on the feed right before your own post publishes), and add observability so we can confirm it happened per post.

## Findings from investigation (2026-07-25)
- The wiring is correct: `auto_check_scheduled_posts` dispatches `automate_commenting.apply_async(eta=scheduled_time − 15 min, loop 15 min)` and the profile-viewer task at −10 min. The 429 breaker/pause was CLEAR, so it wasn't being throttle-skipped.
- **The dominant cause it "isn't happening near posts" is that posts are sparse/gapped** (user 1 had nothing scheduled Jul 19→28), so there's nothing for pre-post commenting to precede — fixed by the content-buffer issue (regular posts ⇒ regular pre-post commenting).
- Residual reliability gaps worth fixing:
  1. **Late pickup:** `get_ready_to_post_posts` looks back to "yesterday" and ahead 20 min; a post picked up <15 min before its time yields an `eta` in the PAST, so commenting runs immediately (a few min before) rather than a clean 15-min warm-up — acceptable but imprecise, and a post already past its time still triggers a pointless "pre"-post run.
  2. **No per-post observability:** nothing records whether the pre-post commenting actually ran/commented for a given post, so we can't verify the engagement window fired.
  3. **Worker restarts** (frequent deploys) can interrupt the 15-min loop.
  4. Depends on the TZ audit: if `scheduled_time` drifts vs the displayed time, the window fires at the wrong wall-clock moment.

## Fix / acceptance
- Guard the eta so a post already within/after its window doesn't schedule a stale "pre"-post loop; clamp to fire ASAP but only if meaningfully before the post.
- Record a per-post engagement-window marker (log + a small stat) so a report can confirm "pre-post commenting ran for post N, M comments."
- Keep it gated by the throttle/pause (correct) but log clearly when skipped for that reason.
- Verify against the TZ contract (linked issue) so the window aligns with the user-intended post time.
- Unit tests for the eta-clamping + skip/observability logic ≥90% patch.

PR #674, opened 2026-07-26
feat(engagement): human-pacing engine — read-time delays, schedule jitter, variable daily volumes (closes #626)

### LINKED ISSUE #626: D2 — Human-pacing engine: read-time delays, schedule jitter, variable daily volumes
## Context
LinkedIn's comment-bot detection keys on cadence: engagement seconds after a post appears, machine-exact daily times, identical intervals, constant daily volumes. LEM currently comments immediately upon selection and runs on fixed beat schedules.

## Scope
- Read-time-realistic delay before commenting: scale with target post length (e.g., 45s–4min) + random browse time; never interact faster than a human could have read.
- Jitter all engagement task start times ±30–90 min (beat schedule offsets or in-task sleep-with-jitter honoring worker capacity).
- Variable daily volumes: derive each day's comment budget from the cap as a random draw (e.g., 40–100% of cap) with occasional rest days and weekend asymmetry; persist the day's budget so retries don't re-roll.
- Centralize in one pacing module consumed by commenting, replies, connections, DMs — a single account-level governor so combined tasks can't exceed the envelope.
- Interplay: respects rate_limit.py breaker; G7 golden-hour replies are exempt from long delays (replying fast to comments on YOUR OWN post is human) but still jittered.

## Acceptance
- Unit tests with seeded RNG: delay bounds, budget persistence, rest-day distribution, governor accounting.
- No Celery task blocks a worker >5 min sleeping — use countdown/eta scheduling, not in-process sleeps, for long delays.

gitchrisqueen:
## What & why

LinkedIn's comment-bot detection keys on **cadence**, not content: engaging seconds after a post appears, machine-exact daily start times, identical intervals, and a flat daily volume. LEM commented immediately on selection and ran on fixed beat schedules with a constant per-day cap — a textbook machine signature.

`src/cqc_lem/utilities/human_pacing.py` is now the ONE place cadence is decided, consumed by commenting, replies, DMs and connection invites.

### 1. Read-time-realistic delay
`read_delay_seconds` / `pace_read` — reading time at `PACING_READ_WPM` plus a browse allowance, clamped into `[PACING_READ_MIN_SECONDS=45, PACING_READ_MAX_SECONDS=240]`. The **floor** is the point: a one-line post still costs 45s. The ceiling sits below `MAX_INLINE_SLEEP_SECONDS=300`, so **no worker ever parks >5 min sleeping**. Replaces the old `simulate_reading_time/2 + simulate_thinking_time` pair (both deleted), which could clear a short post in ~3s.

Wired into `_engage_card` (feed + roster) and `generate_and_post_comment` (permalink path), so no lane is the fast one that gives the account away.

### 2. Schedule jitter
`dispatch_jitter_seconds` returns a random forward countdown — `apply_async(countdown=…)`, i.e. eta scheduling, never an in-process sleep. Applied to `auto_daily_engagement`, `auto_appreciate_dms`, `dispatch_comment_followups` and `auto_check_connection_requests` (30–90 min).

**G7 exemption honoured:** replying to comments on the user's OWN post is human, so `dispatch_scheduled_reply_sweeps` and the golden-hour sweeps use `PACE_RESPONSIVE` — jittered by *seconds*, not delayed by an hour. Golden-hour jitter is additively capped at half a step, so a misconfigured value can never reorder or collapse two sweeps.

### 3. Variable daily volumes
`daily_budget` turns each per-day cap into a random draw (40–100% by default) with weekend asymmetry (`PACING_WEEKEND_RATIO=0.55`) and occasional **account-wide rest days** (`PACING_REST_DAY_CHANCE=0.08`). The draw is **seeded on (user, action, date)** and persisted in Redis for the day — so a retry, a second worker, a mid-day cap edit, or a Redis outage all re-derive the same number instead of re-rolling a fresh (possibly larger) budget. A non-zero cap never silently becomes 0; only a rest day does.

### 4. Account-level governor
`remaining_actions` takes the smaller of the lane's remaining paced budget and the remaining **account envelope** (the sum of every capped lane's budget), fed by `record_action` at every site where an action actually lands: comments (`_engage_card`), DMs (`send_dm_now`), invites (`invite_to_connect_now`). Replaces the raw-cap checks in `comment_on_feed_inline`, `send_scheduled_dm`, `_fire_funnel_stage` and `auto_check_connection_requests`, so the funnel can't quietly push the day past the envelope.

**Replies are tracked but deliberately stay OUTSIDE the envelope** (`ENVELOPE_ACTIONS`): going quiet on people who commented on the user's own post in order to protect a commenting budget is worse for the account than the cadence risk it would avoid. They are still jittered.

### Interplay & safety
- **Fails open everywhere.** No Redis, or `HUMAN_PACING_ENABLED=false`, restores the exact pre-#626 behaviour (full cap, no jitter, no added delay).
- The 429 breaker / kill-switch in `rate_limit.py` is untouched and still the harder gate — pacing only ever makes us slower, never more aggressive.
- No DB migration; all state is Redis + env.

## Testing notes

- `tests/unit/utilities/test_human_pacing.py` (36 tests) — the engine, all seeded or env-pinned: delay bounds + the inline-sleep ceiling (including a misconfigured `PACING_READ_MAX_SECONDS`), jitter bands + an inverted band, budget range/stability/persistence/corrupt-value re-derivation, weekend asymmetry, rest-day **distribution** (~10% over 2000 users), governor accounting, envelope enforcement, Redis-down fail-open, and a Redis that raises on every call.
- `tests/unit/app/test_human_pacing_integration.py` (15 tests) — every wiring seam: each dispatcher's countdown band, the responsive profile on own-post replies, golden-hour sweeps jittered-but-never-reordered, paced budgets at the commenting/DM/invite lanes, and `record_action` firing on success only (not on a gate-skipped comment, a failed DM, or a failed invite).
- The suite defaults `HUMAN_PACING_ENABLED=false` via an autouse fixture (same pattern as `HUMANIZE_ENABLED`) — otherwise every commenting test would sleep 45s+ and the day's budget would depend on the calendar date the CI job happens to run on. The pacing tests opt back in explicitly.
- Full unit suite: **6185 passed**.

Closes #626

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -664,3 +664,33 @@ YOUTUBE_CLIENT_ID=
 YOUTUBE_CLIENT_SECRET=
 YOUTUBE_REFRESH_TOKEN=
 YOUTUBE_PRIVACY_STATUS=unlisted
+
+# --- Human-pacing engine (issue #626) -------------------------------------------------------
+# LinkedIn's comment-bot detection keys on CADENCE: engaging seconds after a post appears,
+# machine-exact daily start times, and a flat daily volume. Set false to restore the pre-#626
+# behaviour (full cap, no jitter, no read delay) — the whole engine is one switch.
+HUMAN_PACING_ENABLED=true
+# Read-time-realistic delay before commenting: reading time at PACING_READ_WPM plus a browse
+# allowance, clamped into [min, max]. The FLOOR is the point — a one-line post still costs 45s.
+# The ceiling is also what keeps the sleep inline-safe (a worker must never park >5 min).
+PACING_READ_MIN_SECONDS=45
+PACING_READ_MAX_SECONDS=240
+PACING_READ_WPM=240
+PACING_BROWSE_MIN_SECONDS=3
+PACING_BROWSE_MAX_SECONDS=25
+# Beat-dispatched engagement doesn't start on the crontab minute — it gets a random forward
+# countdown in this band (Celery eta scheduling, never an in-process sleep).
+PACING_JITTER_MIN_MINUTES=30
+PACING_JITTER_MAX_MINUTES=90
+# Replies to comments on the user's OWN post are exempt from the long delay (answering fast is
+# human) but still jittered, by seconds rather than by an hour.
+PACING_RESPONSIVE_JITTER_MAX_SECONDS=180
+# Each day's volume is a random draw from the user's cap rather than the cap every day, with
+# weekends quieter and the occasional full rest day. The draw is seeded on (user, action, date),
+# so a retry or a Redis outage re-derives the same budget instead of rolling a fresh one.
+PACING_BUDGET_MIN_RATIO=0.4
+PACING_BUDGET_MAX_RATIO=1.0
+PACING_REST_DAY_CHANCE=0.08
+PACING_WEEKEND_RATIO=0.55
+# Optional salt for the seeded draws — change it to reshuffle every user's budget/rest-day pattern.
+PACING_SEED=
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -52,6 +52,7 @@ src/cqc_lem/
 │   │   └── helper.py, profile.py, token_refresh.py
 │   ├── marketing/ Outbound production
 │   │   └── video_tutorials.py  automated SPA tutorial videos (capture→script→TTS→ffmpeg→YouTube)
+│   ├── human_pacing.py  ONE cadence engine (read delays, schedule jitter, variable daily volume, account governor)
 │   ├── db.py      All database access (no raw SQL outside this file)
 │   ├── proxy.py   Per-user static residential proxy resolution
 │   ├── geocoding.py  Login Location city/state geocoding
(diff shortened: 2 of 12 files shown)


gitchrisqueen:

- **Lowering a per-day cap mid-day was ignored** (`human_pacing.py:222`): `daily_budget` returned the Redis-persisted draw verbatim, so a user who drops `max_comments_per_day` from 20 to 3 because they're worried about the account kept spending the stored 20 for the rest of the day — the module's own "pacing only ever makes us slower, never more aggressive" invariant, broken. Clamped the stored value to the current cap; raising a cap still can't re-roll the day's draw. (The existing test only covered a cap *increase*.)
- **`HUMAN_PACING_ENABLED=false` did not restore pre-#626 behaviour** (`human_pacing.py:314`): only `daily_budget` short-circuited on the switch. `remaining_actions` still applied the account envelope over a Redis counter that `record_action` writes regardless of the flag, so with pacing OFF a busy DM/invite day could throttle commenting through a shared pool that didn't exist before this PR. The envelope is now skipped entirely when pacing is disabled.
- **The golden-hour jitter silently killed the pre-post warm-up run** (`run_scheduler.py:346`): the 30–90 min countdown rode `automate_commenting`, which is `QueueOnce(keys=['user_id'])`. celery-once takes that lock *inside* `apply_async`, and `unlock_before_run` only releases it when the task actually RUNS — so the countdown held the user's key for the whole delay, and every pre-post warm-up dispatch for that user in the window (same key, `graceful: True`) was dropped with no exception and no log. Since posts are scheduled into the same golden hours the beat fires in, that collision is likely, and #547's engagement-window marker would just go missing. The countdown now rides a new lock-free `dispatch_golden_hour_engagement` shim, so the lock is taken only when the run is really queued. Still the `se_engage` lane, still no in-process sleep.
- **Invite countdown could outlast the orphan-recovery gap** (`run_scheduler.py:224`): a request is marked `sending` at dispatch and re-queued once it has sat there `lookback_hours=2`; `send_connection_request` accepts a row in `sending`, so a jitter past that gap means inviting the same person twice. Default max (90 min) was inside 2h, but `PACING_JITTER_MAX_MINUTES` is env-tunable and the coupling was undocumented. Both values now come from one constant and the countdown is clamped 15 min inside it.

Also verified and found clean: the read-delay floor/ceiling math (including misconfigured min > max — still ≤ `MAX_INLINE_SLEEP_SECONDS`), the golden-hour sweep jitter's ordering guarantee, `record_action` firing on success only, Redis bytes handling in `int()`, no dangling references to the deleted `simulate_reading_time`/`simulate_thinking_time`, and no Celery task time limits that the new per-card delay could newly breach.

Tests added: cap-decrease clamp, envelope-off-when-disabled, the shim never handing a countdown to the QueueOnce task, and the invite clamp under a tuned-up jitter band. Existing `test_daily_engagement.py` assertions retargeted at the shim. Full unit suite: **6189 passed**.

gitchrisqueen:

Issue #478, opened 2026-07-24
feat(automation): auto-follow-up (react + reply-to-questions) on replies to OUR automated comments on others' posts


gitchrisqueen:
## What
When someone (usually the post author) **replies to a comment WE automated on their post**, we currently do nothing. We should:
1. **React** (like) to their reply — always.
2. **Auto-reply** when their reply contains a **question** — a voice-aligned answer.

**Scope guard (hard requirement):** only for comments **we** automated via `comment_on_feed_inline` — never a comment the user typed themselves on some post. This is naturally enforced by iterating the `commented_posts` ledger (only automated feed comments land there); manual comments are never in it.

Example: author replied with a question to our comment on `urn:li:activity:7486451907129958400`.

## Why it's not covered today
All existing reply handling — `sweep_reply_comments` → `_reply_to_comments_on_open_post` — operates on the user's **OWN** posts (`get_recent_posted_post_ids`). There is no path that revisits a **third-party** post where we commented and handles the reply to our comment.

Notably, **the trigger already arrives and is wasted**: LinkedIn "X replied to your comment" emails ARE classified as comment notifications (`_COMMENT_PHRASES` includes `"replied to"`) and fire `sweep_reply_comments` — but that sweep only looks at our own posts, so for a third-party reply it's a no-op.

## Depends on #474
Only after #474 does the `commented_posts` ledger store a **navigable** key: `feedurn://urn:li:activity:<id>` → `https://www.linkedin.com/feed/update/urn:li:activity:<id>/`. Pre-#474 rows are `feedpost://<hash>` and can't be revisited. So this feature works for comments made **after** #474 ships.

## Proposed design (v1)
New task `sweep_comment_followups(user_id)` (queue `se_engage`, `QueueOnce(user_id)` + the #475 single-flight lock, 429-safe):
- For each recent `commented_posts` row (last N days, status `commented`, key starts `feedurn://`): navigate to the post, locate OUR comment (our profile slug in the comment author link), read replies to it.
- For each reply **not authored by us** that we haven't handled yet:
  - **React** (like) to it.
  - **If it contains a question** (`?` heuristic + a cheap `lem-simple` classifier to avoid rhetorical/emoji false-positives): generate a voice-aligned reply via the existing `generate_ai_response` + `build`-style machinery and post it with `_reply_to_comment_inline`.
[...]

PR #675, opened 2026-07-26
feat(engagement): comment outcome tracking — replies, likes, Most-Relevant visibility (closes #628)

### LINKED ISSUE #628: D4 — Comment outcome tracking: author replies, likes, Most-Relevant visibility
## Context
LinkedIn's May 2026 enforcement demotes automated-looking comments out of 'Most Relevant' (the default view) — a silent kill that makes commenting effort worthless. LEM posts comments and never looks back: no record of whether the author replied, anyone liked it, or whether it's still visible. Comment→reply rate is the truth metric separating valuable comments from spam (valuable comments also drive 200–500 profile visits per breakout thread per practitioner data).

## Scope
- New table comment_outcomes (migration): user_id, log_id (FK to the comment's logs row), checked_at, author_replied, reply_count, like_count, visible_most_relevant (nullable bool), our_reply_sent.
- Follow-up Selenium task T+24–48h per posted comment (batched, capped, breaker-aware): revisit the post, locate our comment (SDUI selectors via find_first; #478 thread map is prior art), record outcomes; comparing default 'Most Relevant' sort vs 'Most recent' presence gives the demotion signal (best-effort, nullable when ambiguous).
- Weekly aggregate: author-reply rate, like rate, demotion rate → PostHog + analytics dashboard; feed the quality score back so G2 can tighten (e.g., demotion rate >X% pauses commenting and flags needs-human).
- Reuse commented_posts URLs/log post_url hashes to find targets.

## Acceptance
- Unit tests with mocked driver; graceful skip when post/comment not findable (deleted, private) with reason logged.
- Live selector validation via #403/#404 pattern before hard reliance.

gitchrisqueen:
## What & why

Feed commenting was **write-only**. LEM posted a comment, logged `success`, and never looked back — no record of whether the author replied, whether anyone liked it, or whether it was still visible at all. LinkedIn's May 2026 enforcement demotes automated-looking comments out of the default **'Most relevant'** comment view, which is a *silent* kill: the comment still exists, nobody sees it, and every log stays green.

This adds the read-back loop.

### The sweep (`sweep_comment_outcomes`, `se_engage`, read-only)
Runs T+24h per posted comment, batched (15/run), capped, breaker-aware, single-flight locked. Work list = `logs` comment rows with a navigable `feedurn://` key, aged 24h–7d, with no outcome row yet — the `LEFT JOIN` is what makes the check at-most-once. It revisits the post, finds our comment with the same #478 thread map (`_load_comment_thread` / `_comment_items`, **extracted and reused**, not duplicated), and writes one `comment_outcomes` row.

### `visible_most_relevant` is three-valued on purpose
* `1` — present under the default 'Most relevant' sort
* `0` — absent there but present under 'Most recent' → **the demotion signal**
* `NULL` — ambiguous (sort control unreadable, or the flip couldn't be verified)

NULL rows are excluded from the demotion denominator entirely, never counted as healthy — treating "we couldn't tell" as "fine" is exactly how a silent kill stays silent. A sort flip that doesn't *verify* never reads as a demotion.

### Graceful skips
A post that's deleted/private, or a comment findable in neither sort, writes a `status='skipped'` row with a reason and a log line. That's what stops an unfindable comment being re-walked every night, and it keeps un-observed comments out of the reply-rate average instead of fabricating a zero.

### Weekly aggregate + the G2 feedback loop
`auto_weekly_comment_quality` scores each user (author-reply rate, reply rate, like rate, demotion rate) → PostHog `comment_quality` + `comment_outcome` events, and the same numbers render on the Dashboard via `/user/engagement-analytics`. If the demotion rate clears `COMMENT_DEMOTION_HOLD_RATE` (0.5) on ≥ `COMMENT_QUALITY_MIN_SAMPLE` (10) *readable* readings, it **holds that user's feed commenting** and escalates as `CRITICAL` (→ PostHog) for a human. The hold is deliberately narrower than the global `pause_automation`: posting, replies and DMs are untouched, because the problem is the comments. The system does **not** self-resume.

A scary rate on a thin sample is reported as `watch` and stops nothing — three reads must never pause a user's engagement.

## Live selector validation (#403/#404 pattern)
The demotion signal leans on a DOM control we have not grounded live, and this is one place a wrong read is expensive. So: nothing is trusted until it verifies (unknown → NULL), the guard needs a real sample, and `scripts/linkedin_live_validation.py --comment-outcome-url '<post we commented on>'` reports exactly what the reader sees under each sort (`visible` / `demoted` / `ambiguous: …`) so the assumption can be checked on a supervised run before the hold is relied on.

One known open question the probe is meant to settle: LinkedIn may surface the *viewer's own* comment preferentially, which would suppress true demotions. That biases the signal toward false negatives (under-reporting demotion), never toward a spurious hold.

## Files
| Area | Change |
|---|---|
| migration | `V20260726170427__add_comment_outcomes.sql` — `UNIQUE (user_id, log_id)`, nullable `visible_most_relevant`, CASCADE on user + log |
| `utilities/comment_outcomes.py` | the pure scoring core shared by the PostHog event, the API and the dashboard, so the number the user reads and the number the guard acts on can't diverge |
| `utilities/db.py` | `get_comment_outcome_targets` / `record_comment_outcome` / `get_comment_outcomes` |
| `run_automation.py` | sort reading + verified flip, our-comment matching, like/reply/author-reply extraction, the sweep; the commenting hold gate |
| `run_scheduler.py` | daily dispatcher + weekly quality report |
| `rate_limit.py` | `hold_commenting` / `release_commenting_hold` (per-user, fails open) |
| `observability.py` | `track_comment_outcome`, `track_comment_quality` |
| `api/main.py` + `Dashboard.tsx` | `comment_quality` block on `/user/engagement-analytics` and its panel |

## Testing
* `tests/unit/utilities/test_comment_outcomes.py` (16) — rate arithmetic, three-valued visibility, verdict thresholds, env tunables.
* `tests/unit/app/test_comment_outcomes.py` (49) — mocked-driver DOM helpers, every skip path, "failed sort switch is not a demotion", sweep orchestration (lock contention, 429, no slug, one bad post), the commenting-hold gate, the weekly report + hold, the dispatcher, the live probe's verdict.
* `tests/integration/test_comment_outcomes_db.py` (11) — real MySQL: the age/navigability window, at-most-once (incl. after a *skipped* check), NULL surviving the round trip, upsert-not-duplicate, user scoping, CASCADE.
* `poetry run pytest tests/unit -q` → **6434 passed**. `tsc -p tsconfig.app.json` clean on `Dashboard.tsx`.

Closes #628


gitchrisqueen:

- **Sort flip could never match its menu option (kills the whole demotion signal).** `_switch_comment_sort` located the option with `normalize-space()='{target.title()}'` → `'Most Recent'`, but LinkedIn renders `Most recent` and XPath string comparison is case-sensitive. `find_first` would always return None, the flip always returned False, so `visible_most_relevant` could never be `0` — every demoted comment was recorded as `skipped/comment-not-found` and the hold could never fire. Fixed: added `_sort_option_locators()` and case-folded every sort locator (including `_COMMENT_SORT_LOCATORS`) through an XPath `translate()`; added tests that assert no exact-case literal survives in either locator set.
- **Profile identity matched by substring.** `f"/in/{our_slug}" in author_href` is also true for every slug ours is a prefix of — `/in/chris` matches `/in/chris-queen-9b1`. In `_find_our_comment` that can select a stranger's comment as ours (recording their replies/likes as our outcome); in `_read_comment_outcome` it discounts a real reply as `our_reply_sent` and can misattribute `author_replied`. Fixed with `_href_is_profile()` (exact `_profile_slug` equality, case-insensitive on the href) at all four call sites, plus `--our-slug` normalization in the live probe so a URL or a bare slug both work. Tests cover the prefix collision on both paths.

Verified, not changed (noting for the record):
- `_load_comment_thread` clicks "load more comments" before the visibility read, so a comment that is merely paginated below the fold under 'Most relevant' is counted as visible. That biases the signal toward **false negatives** (under-reporting demotion), never toward a spurious hold — the safe direction, and exactly what `--comment-outcome-url` is there to ground.
- `_MAX_OUTCOME_CHECKS_PER_RUN=15` sits under the default `max_comments_per_day=20`, so a user at cap leaves ~5 comments/day to age out at the 168h bound. The sample stays representative (oldest-first, quality-independent), and raising it would push a single sweep toward the 1800s run lock, so left as-is.
- Skip/NULL semantics, the at-most-once LEFT JOIN, the fail-open Redis hold, and the three-valued visibility round trip all hold up under the tests and the integration suite.

`poetry run pytest tests/unit -q` → 6465 passed.

gitchrisqueen:

Issue #650, opened 2026-07-26
PH5 — Business KPI funnels, consolidated dashboards, threshold alerts + weekly report subscription


gitchrisqueen:
## Context
41 insights / 6 dashboards exist but there is no FUNNEL view of LEM's actual business loops, no alerting on metric floors/ceilings, and the 2-week perf report is a hand-run cron. PostHog insight threshold alerts (5 free) + dashboard email subscriptions can replace metric-watchdog crons.

## Scope
- Funnels (over existing + PH1/PH2 events): content loop (content_plan_generated → post_approved → post posted (celery_task/post_outcome) → engagement recorded); engagement loop (comment posted → author reply (D4 data when it lands) → connection accepted → DM reply); onboarding funnel (onboarding_step exists).
- Consolidate dashboards: one 'LEM Health' dashboard (task failures, 429s, LLM cost/day from $ai_generation, comments/day, posts/day, follower delta once D3 lands) + one 'Growth' dashboard (funnels above, ER trends).
- Threshold alerts: comments-posted/day below floor, $ai_generation daily cost above cap, celery task failure-rate spike, 429 count spike → email.
- Weekly email subscription of the Growth dashboard to the owner — automates the perf-report cadence.
- Prefer creating insights via API/HogQL in a idempotent setup script (scripts/posthog_provision.py) committed to the repo so dashboards are reproducible, not hand-built.

## Acceptance
- Provision script runs idempotently against POSTHOG_PERSONAL_API_KEY (no-op when unchanged); dashboards/alerts exist and fire on synthetic threshold breach; script covered by unit tests with mocked API.

PR #678, opened 2026-07-26
feat(ui): instrument the SPA with PostHog — identify, autocapture, product events (closes #646)

### LINKED ISSUE #646: PH1 — SPA instrumentation: posthog-js, identify, autocapture, product events
## Context
The React SPA has ZERO PostHog: no posthog-js dependency, so no pageviews, autocapture, funnels, web vitals, or session context. FeedbackWidget.tsx already reads window.posthog?.get_session_id() — it has silently returned undefined forever. Backend events use distinct_id=str(user_id)/'system' with no $identify and no person properties, so persons are anonymous shells.

## Scope
- Add posthog-js (npm) + init in the SPA entry (env-gated: VITE_POSTHOG_KEY/VITE_POSTHOG_HOST from build; absent key = fully disabled no-op).
- $identify on login with distinct_id matching the backend convention (str(user_id)) so server and browser events unify into one person; person properties: plan/role, timezone, created_at (NO LinkedIn credentials or PII beyond email).
- Autocapture ON with privacy defaults: mask_all_inputs, and mask text on any element carrying a data-ph-mask attribute; add that attribute to DM/template/story editors.
- SPA-router pageview capture ($pageview on route change) + web vitals.
- Manual product events for the moments autocapture can't name: post_approved / post_rejected (with post_type+archetype), prefs_saved (section), dm_template_saved, story_bank_entry_added, content_plan_generated, feedback_opened.
- FeedbackWidget: posthog session id now resolves — verify it lands in feedback reports.

## Acceptance
- Unit/UI build passes; events verified against a dev PostHog key; docs snippet in CLAUDE.md observability section.
- With no key set, bundle adds no network calls (guard import cost).
Prerequisite for PH4 (replay) and PH8 (surveys).

gitchrisqueen:
## What & why

The React SPA had **zero** PostHog. No `posthog-js` dependency meant no pageviews, no autocapture, no web vitals and no session context — and `FeedbackWidget.tsx` has been calling `window.posthog?.get_session_id()` against nothing for as long as it has existed. Backend events already used `distinct_id=str(user_id)`, but with no `$identify` anywhere, every PostHog person was an anonymous shell.

This adds the browser half, as **one** surface: `ui/src/utils/analytics.ts`. Components never touch `posthog` directly, so the key, the privacy defaults and the distinct_id convention live in exactly one place.

### Off means off
`posthog-js` is loaded through a **dynamic import** gated on `VITE_POSTHOG_KEY`. With no key, Vite ships it as a lazy chunk the browser never fetches, `initAnalytics()` resolves `null`, and every `capture`/`identify`/`reset` is a no-op — no network calls, no cost.

### One person, not two
`$identify` fires from `AuthContext` with `String(user_id)` — byte-for-byte the convention `observability.py` uses server-side — so a user's Celery/API events and their browser events land on the same person. Person properties are `plan` / `plan_status` / `timezone`, plus `created_at` as `$set_once`; `GET /auth/session` now returns them (`users` has no `created_at`, so signup time is `trial_started_at` falling back to `updated_at`, the same convention the cohort query uses). **No LinkedIn credentials, cookies or profile data** — email is the only PII. `logout()` calls `posthog.reset()`.

Side benefit: `login()` now reads the session back, so `user.userId` is populated immediately instead of only after the next page load (it was previously `undefined` for the whole first session — which also meant feedback reports filed right after login carried `user_id: null`).

### Privacy defaults
Autocapture is on with `mask_all_element_attributes`. `maskProps()` stamps both `ph-no-capture` (autocapture skips the element) and `data-ph-mask` (replay's `maskTextSelector`, ready for PH4) onto every editor holding the user's own content: draft posts, carousel slides, DM templates, scheduled DM drafts and story bank entries.

### Pageviews + web vitals
`capture_pageview: false` and a `usePageviews()` hook off the router, because posthog's own listener only fires on a full page load and would miss every in-app navigation (`?tab=` included — that's a different screen). Web vitals ride on `capture_performance`. Replay stays off; that's PH4.

### Product events
The moments autocapture can't name, all in the `EVENTS` vocabulary:

| Event | Where | Properties |
|---|---|---|
| `post_approved` / `post_rejected` | Content Studio editor, bulk bar, soft delete | `post_id`, `post_type`, `archetype`, `source` |
| `prefs_saved` | the one Save All registry | `section`, `ok` |
| `dm_template_saved` | DM Templates card | `templates`, `followup_steps` |
| `story_bank_entry_added` | Story Bank save | `added`, `total`, `kinds` |
| `content_plan_generated` | Generate Weekly Content | `source` |
| `feedback_opened` | Feedback widget | `route` |

Two deliberate calls there: a post decision fires only on a real **status transition** (re-saving an already-approved draft is an edit, not a second approval), and `story_bank_entry_added` fires on the SAVE, not on "+ Add entry" — a blank row the user abandons never became an entry. `/posts/` now returns `posts.archetype` (V51) so the decision can carry the shape the draft was written to.

## Testing

- `src/utils/analytics.test.ts` — disabled path never loads posthog-js and swallows every call; init config (pageview ownership, mask flags, web vitals, replay masking selector); init-once; identify shape incl. `$set_once` and omitted props; pageview/product capture; reset; session-id fallback to a script-tag posthog; a throwing posthog never reaching the caller; `maskProps`.
- `SettingsSaveContext.test.tsx` — one `prefs_saved` per saved section, and a failed save recorded as `ok:false` rather than dropped.
- `tests/unit/utilities/test_db_analytics_profile.py` — including an assertion that the SELECT touches no credential column.
- `tests/unit/api/test_main_general.py` — the new `/auth/session` person facts, and the empty-profile fallback.

`poetry run pytest tests/unit -k "posts or db"` → 1505 passed. `tsc -b` clean, `eslint` clean on the new files. The local box is on node 18, so `vitest` and `vite build` (which need ≥20.19) run in CI, not here.

**Not verifiable headless:** the acceptance line "events verified against a dev PostHog key" needs a real browser against a dev project — worth a manual pass once `UI_POSTHOG_KEY` is set. CI builds with the secret unset, which exercises the fully-disabled path.

## Deploy note

`VITE_POSTHOG_KEY` / `VITE_POSTHOG_HOST` are **build-time** (Vite bakes them into the bundle), wired as Dockerfile build args and passed by both UI workflows. Browser analytics stays off until the repo secret `UI_POSTHOG_KEY` (the public `phc_…` project key, never the personal API key) and the `UI_POSTHOG_HOST` variable are set.

Prerequisite for PH4 (replay) and PH8 (surveys).

Closes #646

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -450,6 +450,13 @@ POSTHOG_PERSONAL_API_KEY=
 # PostHog project the dashboards live in, and the app host used to build their links.
 POSTHOG_PROJECT_ID=475262
 POSTHOG_APP_HOST=https://us.posthog.com
+# --- PostHog in the SPA (issue #646) ---
+# Read by Vite at BUILD time and baked into the bundle, so they must be set for `npm run build`
+# (docker build-arg / CI env), not in the running container. Leave VITE_POSTHOG_KEY empty and the
+# SPA never even fetches the posthog-js chunk — browser analytics is fully off.
+# This is the PROJECT key (phc_…), which is public by design — never the personal API key.
+VITE_POSTHOG_KEY=
+VITE_POSTHOG_HOST=https://us.i.posthog.com
 
 # --- Unit economics / margin report (docs/cost-performance-margin-plan.md §C.1) ---
 # Monthly price per subscription tier, used to compute MRR for contribution margin. Defaults match
--- a/.env.prod.example
+++ b/.env.prod.example
@@ -146,6 +146,10 @@ CAPSOLVER_API_KEY=
 POSTHOG_API_KEY=CHANGE_ME
 POSTHOG_HOST=https://us.i.posthog.com
 POSTHOG_LOG_LEVEL=ERROR
+# Browser-side PostHog (issue #646). BUILD-time only — the SPA bundle bakes these in, so setting
+# them here does nothing; they are passed as docker build-args (CI: UI_POSTHOG_KEY secret).
+# VITE_POSTHOG_KEY=phc_...
+# VITE_POSTHOG_HOST=https://us.i.posthog.com
 
 # =============================================================================
 # Tunnel & Networking (Ngrok) — DISABLED in prod (Cloudflare Tunnel is used)
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -54,6 +54,8 @@ jobs:
           build-args: |
             INSTALL_DEV_DEPS=false
             VITE_API_TOKEN=${{ secrets.UI_API_TOKEN }}
+            VITE_POSTHOG_KEY=${{ secrets.UI_POSTHOG_KEY }}
+            VITE_POSTHOG_HOST=${{ vars.UI_POSTHOG_HOST }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
--- a/.github/workflows/ui-build.yml
+++ b/.github/workflows/ui-build.yml
@@ -29,3 +29,5 @@ jobs:
         run: npm run build
         env:
           VITE_API_TOKEN: ${{ secrets.UI_API_TOKEN }}
+          VITE_POSTHOG_KEY: ${{ secrets.UI_POSTHOG_KEY }}
+          VITE_POSTHOG_HOST: ${{ vars.UI_POSTHOG_HOST }}
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -204,6 +204,37 @@ from cqc_lem.utilities.observability import track_llm_call, track_task, track_ap
 
 PostHog receives LLM usage (model, tokens, latency, cost) and Celery task metrics for fine-tuning decisions.
 
+### Browser-side analytics (SPA, issue #646)
+
+The SPA has ONE PostHog surface — `ui/src/utils/analytics.ts`. Never call `posthog` directly from a
+component; import from there so the key, the privacy defaults and the distinct_id convention stay
+in one place.
+
+```ts
+import { EVENTS, capture, maskProps } from '../utils/analytics'
+
+capture(EVENTS.postApproved, { post_id, post_type, archetype })
+
+// Any editor holding the user's own content (DM, story, draft post):
+<textarea {...maskProps('w-full border rounded-lg px-3 py-2 text-sm')} />
+```
+
+- **distinct_id is `String(user_id)`** — identical to `observability.py`'s server-side convention,
+  so a user's browser and Celery/API events land on ONE PostHog person. `$identify` fires from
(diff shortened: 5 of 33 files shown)


gitchrisqueen:

- **`prefs_saved` fired from only one of two doors** (`SettingsSaveContext.saveAll`). Groups, Lead Magnet, Engagement Roster, Story Bank and DM Templates each keep their **own** Save button that calls its mutation directly and never enters the Save-All registry, so the event would have counted only the users who happen to use "Save All" — a silently biased denominator for every rate built on it, and the same "another door" problem `post_approved` already solves for the editor/bulk/delete paths. Added `sectionSaveCallbacks(section)` as the one reporting point and spread it onto each card's own `mutate()` call, so neither door double-counts; covered in `SettingsSaveContext.test.tsx`.
- **Acceptance gap on the mask sweep** — the criteria call for `data-ph-mask` on the DM/template/story editors, but the outreach-funnel comment/connect/DM drafts, the leads-inbox reply, the catch-up message, the connection note, the newsletter body and the lead-magnet DM message were still unmasked. `maskTextSelector` is configured here and PH4 (#650) switches replay on against whatever carries the attribute at that point, so the sweep belongs with the convention that introduces it. `maskProps()` applied to all seven.

Verified and NOT changed:
- `posthog.identify()` on a distinct_id switch — checked the shipped posthog-js 1.407.2 build: it does `register({distinct_id})` unconditionally when the id changes, so a second user on the same browser can't inherit the first user's person. The once-per-page-load `identifiedUserId` ref is safe.
- Disabled path — `import type` is erased and the only value import of `posthog-js` is dynamic, so with no `VITE_POSTHOG_KEY` the chunk is never fetched and `capture`/`identify`/`reset` are no-ops.
- Autocapture never carries editor content: posthog treats `textarea`/`select`/non-button `input` as sensitive, so `getSafeText` returns empty for them regardless of the mask.
- `capturePostDecision`'s status-transition guard reads the pre-invalidation react-query row, and the UI's status literals are lowercase on every path (editor select, bulk bar, soft delete), so the `approved`/`rejected` comparison holds.
- `users` really does carry `subscription_tier` / `subscription_status` (V18/V22), `timezone` (V28) and `trial_started_at` (V22), and no `created_at` — the `COALESCE(trial_started_at, updated_at)` fallback is honest and `$set_once` freezes it.
- `posts.archetype` exists (V51) and `get_posts` uses a dictionary cursor, so the added column can't shift any positional read.
- `tsc -b` clean; `pytest tests/unit/utilities/test_db_analytics_profile.py tests/unit/api/test_main_general.py` → 59 passed. vitest needs node ≥20 (box is on 18), so the UI suite runs in CI.

gitchrisqueen:

PR #679, opened 2026-07-26
feat(observability): LLM analytics — LiteLLM→PostHog native callback ($ai_generation) (closes #647)

### LINKED ISSUE #647: PH2 — LLM analytics: LiteLLM→PostHog native callback ($ai_generation)
## Context
Every LLM call flows through the LiteLLM proxy, but LEM tracks them only as custom llm_call events — PostHog's LLM-analytics product (generations/traces/costs/latency dashboards, per-user per-model breakdowns, evals) sits unused. LiteLLM ≥1.77.3 ships a native posthog logger: proxy config litellm_settings.success_callback: ['posthog'] (+ failure_callback) with POSTHOG_API_KEY/POSTHOG_API_URL env — zero app-code changes emits $ai_generation for every lem-* call.

## Scope
- .litellm/config.yaml: add posthog to success_callback + failure_callback; wire env vars through docker-compose (litellm service).
- Verify against our pinned LiteLLM version, incl. the reported proxy serialization bug (BerriAI/litellm#18332, UserAPIKeyAuth not JSON serializable) — pin/upgrade if needed.
- User/feature attribution: pass metadata (user_id, feature from llm_attribution context) per-request from utilities/ai/client.py so $ai_generation carries who/what — LiteLLM forwards request metadata to the logger.
- De-dupe strategy: keep track_llm_call (feeds cost_ledger + rollups + margin reports) but document $ai_generation as the canonical analytics stream; ensure dashboards don't double-count by using one or the other per insight.
- Stretch (separate PR ok): trace structure — wrap multi-call pipelines (post generation: research→draft→review→humanize) in a $ai_trace/$ai_span hierarchy via posthog-python AI helpers so a post's full generation chain reads as one trace.

## Acceptance
- $ai_generation events visible with model, tokens, cost, latency, user_id, feature; failure events captured; no regression in existing cost ledger tests. LLM analytics dashboard shows per-model costs matching the LiteLLM proxy's own accounting within tolerance.

gitchrisqueen:
## What & why

PH2 of the PostHog milestone. Every LEM model call already flows through the LiteLLM proxy, but PostHog only ever saw the app's own `llm_call` **estimate** — the LLM-analytics product (generations, per-model cost/latency, per-user/feature breakdowns, evals) sat unused.

LiteLLM's native `posthog` callback makes the proxy the canonical analytics stream: one `$ai_generation` (`$ai_embedding` for embeddings) per call carrying `$ai_model`/`$ai_provider` (**the model that actually served it**, post-fallback and post-down-route), `$ai_input_tokens`/`$ai_output_tokens`, the provider's own `$ai_total_cost_usd`, `$ai_latency`, and — via `failure_callback` — `$ai_is_error`/`$ai_error`, so a broken model reads as errors instead of silence.

### Attribution lives in the ONE client

`utilities/ai/client.py` now stamps `metadata: {user_id, feature}` on every `lem-*` request, read from the ambient `llm_attribution()` scope. It goes there rather than at the ~10 direct call sites (embeddings for comment dedup and feedback clustering, `content_research`, `tools`, `carousel_creator`, image/TTS) because a call that skips it is invisible to cost routing **and** lands on an anonymous PostHog person — a silent failure nobody would notice. `_call_llm` still sets its own metadata first, which is what lets an explicit `_track_user_id`/`_track_feature` beat the ambient scope.

`metadata.user_id` is what LiteLLM uses **verbatim** as the event's distinct_id, so unattributed traffic now sends the same `"system"` sentinel `observability.py` uses server-side (and the SPA's `String(user_id)` resolves to) — one PostHog person per user across browser, Celery and proxy, instead of a throwaway anonymous person per housekeeping call.

That sentinel is not a user, so `routing_policy.assign_arm` reads it exactly like a missing id: unattributed traffic still lands in the cost experiment's **control** arm rather than all hashing into one arm together. Real ids are unchanged (the salt is `f"...|{user_id}"`, so int `7` and `"7"` bucket identically — there's a test).

### Privacy

`turn_off_message_logging: true` redacts `$ai_input`/`$ai_output_choices` before any callback sees them. LEM's prompts are the user's own LinkedIn material (profile synthesis, story-bank anecdotes, draft DMs) and the SPA already masks exactly that content with `maskProps()`. Metrics are unaffected. Flipping it to false gives PostHog's generation view full conversation text — a deliberate decision, not a default.

### De-dupe

Both streams fire per call and **must never be summed**. `docs/llm-analytics.md` + a new CLAUDE.md subsection spell out the split: money questions use `llm_call` (it is the `cost_ledger`/margin-report source and joins to Stripe); latency, error rate, model mix and volume use `$ai_generation` (it is the truth about what ran). Also documents the cache-hit caveat — a LiteLLM cache hit emits `$ai_generation` at cost 0 but with the served request's input tokens.

## Version verification (issue asked for this)

- Callback requires LiteLLM **≥1.77.3**.
- The reported proxy serialization bug (BerriAI/litellm#18332, `Object of type UserAPIKeyAuth is not JSON serializable`, reported on v1.80.8) was fixed by `posthog serilization fix (#20668)` on 2026-02-07 — the batch is now written with `safe_dumps`.
- The stack runs `ghcr.io/berriai/litellm:main-latest`, which is well past both. **No pin/upgrade needed for this issue.** (Worth noting separately: `main-latest` is unpinned in general — out of scope here, but a candidate for its own issue.)
- Known upstream wart: the logger copies raw request metadata into event properties, and LiteLLM's proxy puts a `user_api_key_auth` object there. `safe_dumps` stops it crashing, so it lands as a stringified property containing the master key's **hash** (not the key). Noise, not a leak.

## Failure mode

The callback is best-effort by construction — LiteLLM initializes it lazily and `_init_custom_logger_compatible_class` swallows the failure (`return None`). A missing `POSTHOG_API_KEY` in the litellm container costs **no** generations; it just silently emits no events. That's the first thing to check if `$ai_generation` doesn't appear.

## Testing

- `tests/unit/utilities/ai/test_client_attribution.py` (new, 9 tests) — drives **real** request builds through the OpenAI SDK against a recording transport, so an SDK upgrade that moves the injection point fails CI instead of quietly dropping attribution. Covers: scope attribution, the `"system"` sentinel, raw provider models untagged, a caller's own metadata winning, other `extra_body` fields surviving, attribution errors never losing the generation, and embeddings.
- `tests/unit/utilities/ai/test_litellm_posthog_config.py` (new, 4 tests) — the config half has no runtime guard, so these assert both callbacks are wired, the complexity-router `custom_callbacks` hook isn't shadowed, message logging is off, and the compose service gets both env names.
- `test_routing_metadata.py` + `test_routing_policy.py` extended for the sentinel and int/str bucket stability.
- Full unit suite: **6555 passed**. No cost-ledger regressions.

## Not in this PR

The issue's explicit stretch item — `$ai_trace`/`$ai_span` hierarchies wrapping a post's generation chain (research → draft → review → humanize) into one readable trace. Every event already carries a `$ai_trace_id`, but it is per-call; grouping is app-side `posthog-python` work and the issue says a separate PR is fine.

Closes #647

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -441,6 +441,9 @@ CONTENT_GENERATION_RESULT_TTL_SECONDS=3600
 # --- PostHog (analytics) ---
 POSTHOG_API_KEY=your_posthog_api_key
 POSTHOG_HOST=https://us.i.posthog.com
+# Both are also passed to the litellm container (as POSTHOG_API_KEY / POSTHOG_API_URL), where the
+# native posthog callback emits one $ai_generation per LLM call — issue #647, docs/llm-analytics.md.
+# Prompts/completions are redacted before they reach it (litellm_settings.turn_off_message_logging).
 # Minimum log level forwarded to PostHog (DEBUG/INFO/WARNING/ERROR/CRITICAL). Default: ERROR
 POSTHOG_LOG_LEVEL=ERROR
 # Personal API key (insight + dashboard write scope) used by scripts/posthog_dashboards.py to
--- a/.litellm/config.yaml
+++ b/.litellm/config.yaml
@@ -134,6 +134,22 @@ litellm_settings:
     port: 6379
   custom_callbacks:
     - /app/.litellm/complexity_router.py
+  # ── LLM analytics → PostHog (issue #647, docs/llm-analytics.md) ──────────────
+  # LiteLLM's native logger (>=1.77.3) emits ONE $ai_generation / $ai_embedding per call with the
+  # provider model, tokens, response_cost and latency — the proxy's own accounting, so it needs no
+  # app-side estimate. distinct_id comes from the request's `metadata.user_id`, which
+  # utilities/ai/client.py stamps on every lem-* call; every other metadata key (feature) rides
+  # along as an event property. Needs POSTHOG_API_KEY in this container — see docker-compose.yml.
+  # The app's own `llm_call` event stays: it is what the cost ledger and margin report are built on.
+  # Do NOT sum spend across both streams — see the doc for which insight uses which.
+  success_callback: ["posthog"]
+  failure_callback: ["posthog"]
+  # Prompts and completions are the USER'S LinkedIn material (profile synthesis, story-bank
+  # anecdotes, draft DMs). LEM's analytics contract is that none of it leaves the stack, so the
+  # logging pipeline redacts $ai_input/$ai_output_choices before any callback sees them — the same
+  # rule the SPA follows with maskProps(). Metrics (tokens/cost/latency/model) are unaffected.
+  # Flip to false only deliberately: it turns PostHog's generation view into full conversation text.
+  turn_off_message_logging: true
   drop_params: true
   request_timeout: 180
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -204,6 +204,30 @@ from cqc_lem.utilities.observability import track_llm_call, track_task, track_ap
 
 PostHog receives LLM usage (model, tokens, latency, cost) and Celery task metrics for fine-tuning decisions.
 
+### LLM analytics (LiteLLM → PostHog, issue #647)
+
+There are TWO LLM streams in PostHog and they must never be summed together — see
+`docs/llm-analytics.md` for the full split.
+
+- **`llm_call`** (app, `track_llm_call`) — LEM's cost ESTIMATE, keyed by the tier alias the caller
+  asked for. It is what `cost_ledger`, the margin report and the budget alerts are built on, so
+  **every money question uses this one**.
+- **`$ai_generation`** / **`$ai_embedding`** (proxy) — LiteLLM's native `posthog` callback emits one
+  per call with the model that ACTUALLY served it (post-fallback, post-down-route), the provider's
+  own `response_cost`, tokens and latency. This is PostHog's LLM-analytics product; **use it for
+  latency, error rate, model mix and volume**.
+
+`utilities/ai/client.py` stamps `metadata: {user_id, feature}` on every `lem-*` request — attribution
+lives in the ONE client, not at the ~10 call sites, because a call that skips it is invisible to both
+cost routing and analytics. `metadata.user_id` becomes the event's distinct_id and falls back to the
+`"system"` sentinel (same one `observability.py` uses, same person the SPA's `String(user_id)`
+resolves to), so a user's browser, Celery and proxy events land on ONE PostHog person. Don't remove
+`_attach_routing_metadata` from `_call_llm`: it is what lets an explicit `_track_user_id` beat the
+ambient `llm_attribution()` scope.
+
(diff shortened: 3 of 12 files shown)


gitchrisqueen:

- **Test gap: nothing pinned the shared `client` singleton to `AttributedOpenAI`.** Every test in `test_client_attribution.py` builds its *own* `AttributedOpenAI`, so all 9 still pass with `client = OpenAI(...)` — the single line the entire "attribution lives in the ONE client" design rests on. That regression silently strips `distinct_id` + `feature` off every proxy `$ai_generation` and leaves no other trace. Added `TestTheSharedClient`: an `isinstance` pin plus a request driven through the real singleton's own `_build_request` (no network, no substitute client). Verified adversarially — with the singleton reverted to plain `OpenAI(...)` the two new tests fail and the other 9 still pass, so the gap was real. Full unit suite: **6557 passed**.

Verified against upstream source rather than taken on trust (no changes needed):
- `litellm/integrations/posthog.py` — `_get_distinct_id` does read `litellm_params.metadata["user_id"]` first (then `end_user`, then `trace_id`, then a fresh UUID), so the sentinel genuinely prevents a throwaway person per housekeeping call. `feature` survives the `litellm_internal_fields` filter; `user_id` is filtered out of properties (it is the distinct_id), which is correct.
- Privacy claim holds: `$ai_input`/`$ai_output_choices` are read from `standard_logging_object`, which is what `turn_off_message_logging` redacts. Checked the one documented hole — redaction is unreliable for **streaming** — and LEM sends nothing through the proxy with `stream=True`, so it doesn't apply. Also confirmed `perform_redaction` deep-copies the result before mutating, so the global flag can't poison the Redis `cache: true` write path.
- Serialization fix is on `main` (`async_send_batch` uses `safe_dumps`), so #18332 doesn't drop the batch — worth noting the issue itself was closed *not planned*, it was fixed by the separate PR.
- **One correction to the PR body:** the `user_api_key_auth` property is *not* the master key's hash. `user_api_key_auth` (the `UserAPIKeyAuth` pydantic object) is put in request metadata by the proxy and isn't in the logger's exclusion list, so it does ship as a stringified property — but on the master-key auth path LiteLLM substitutes `LITELLM_PROXY_MASTER_KEY_ALIAS` for `api_key` specifically so "neither the master key nor its hash propagates ... to any downstream consumer of `UserAPIKeyAuth.api_key`". Safer than described, no leak either way.
- Seams checked and clean: `_build_request` signature matches openai 2.44.0 (`options, *, retries_taken`); `extra_json` merge is shallow so preserving the caller's other `extra_body` keys is required and done; re-entry on retry is idempotent; `assign_arm` treats the sentinel as no-user (both new tests confirm int/str bucket stability); `routing_policy.py` stays stdlib-only for the container mount; `docker-compose.prod.yml` doesn't override the litellm `environment`, so prod gets both vars; `lem-embedding`/`lem-image`/`lem-tts` newly carry metadata but are outside `TIERS`, so the down-route path is untouched.

gitchrisqueen:

PR #680, opened 2026-07-26
feat(observability): capture $exception into PostHog error tracking (closes #648)

### LINKED ISSUE #648: PH3 — Error tracking: $exception autocapture + issue-based alerting (replace log-grep cron)
## Context
Errors currently reach PostHog only as LOG entries (logger.py handler), and the daily error→GitHub-issues cron (~/error-to-issues) greps those logs and hand-rolls dedup. PostHog Error Tracking gives $exception capture with issue GROUPING (fingerprinting), statuses, spike detection, and new-issue/reopened alerts — 100K exceptions/mo free. Issues ARE the dedup.

## Scope
- posthog-python: enable_exception_autocapture=True in observability.py init; Celery task_failure/task_retry signal handlers capturing exceptions with task_name/user_id context (no dedicated Celery hook exists — wire signals in my_celery.py); FastAPI exception middleware capture in api/main.py.
- log_error/log_critical: also capture_exception when exc= is provided (keep the log stream as-is — logs stay for context, alerting moves to issues).
- SPA: posthog-js exception autocapture (depends on PH1) + source-map upload in the UI build workflow so stacks are readable.
- Migrate the error→issues cron to consume error-tracking ISSUES (created/spiking) via API instead of grepping logs: 1 PostHog issue ↔ 1 GitHub issue, id-based dedup, keep agent:ready labeling. Simplify/remove the homegrown dedup.
- Alerts: new-issue + spike alerts → email (channel configurable).

## Acceptance
- A thrown test exception in a Celery task and a FastAPI route each produce ONE grouped issue with context; repeated throws don't create duplicate GitHub issues; unit tests for the signal handlers with mocked posthog.

gitchrisqueen:
## What & why

Errors only ever reached PostHog as **log lines**, and the daily `~/error-to-issues` cron grepped those logs, grouped them by the exact message string and hashed it into a dedup marker. Any error whose message carries an id — `Post 41 failed to post` — therefore filed a brand-new GitHub issue every single day, and nothing grouped, counted or alerted.

PostHog Error Tracking already groups `$exception` events into **issues** by fingerprint. The issue IS the dedup key, so the homegrown layer collapses into "one PostHog issue id ↔ one GitHub issue".

### Capture

| Surface | How |
|---|---|
| Any Python process | `posthog.enable_exception_autocapture` (`utilities/observability.py`) — uncaught exceptions |
| Anything logged with `exc=` | `log_error`/`log_critical` also call `capture_exception`; the log stream is unchanged and still carries the message + structured context |
| Celery | `task_failure` + `task_retry` handlers in `my_celery.py` with `task_name`, `task_id`, dispatched `user_id` |
| FastAPI | the unhandled branch of `observability_middleware` with `route` + `method` |
| SPA | `posthog-js` `capture_exceptions` (unhandled errors + rejections) and a `captureException()` helper on the one analytics surface |

Two deliberate exclusions: a route's own `HTTPException` is turned into a response before the middleware sees it, so **a 4xx never files an issue**; and browser **`console.error` capture stays off** — it is noisy and routinely carries the user's own content.

No double counting: `posthog.capture_exception` is idempotent per exception *instance*, so a task that does `log_error(..., exc=e)` and re-raises produces ONE occurrence, not one per layer. Both paths have kill switches (`POSTHOG_EXCEPTION_AUTOCAPTURE`, `POSTHOG_EXCEPTION_CAPTURE`), and with no `POSTHOG_API_KEY` nothing is sent at all.

### Readable stacks

`vite.config.ts` emits source maps and the **ui-builder stage** of the image runs `posthog-cli sourcemap inject` + `upload`. It has to be there rather than in a separate CI job: `inject` writes chunk ids INTO the bundle, and the injected bundle is the one that must ship or an uploaded map can never be matched. Every `.map` is deleted from `dist/` afterwards, so source maps are never served — including on builds that skip the upload. Missing token or failed upload = minified stacks, never a failed release.

### Cron migration

`scripts/posthog_error_issues.py` (cron wrapper `scripts/error_to_issues.sh`) reads the error-tracking columns off `events`, groups by `issue_id`, and files ONE `agent:ready`+`bug` issue per **active** PostHog issue with no GitHub issue yet — body shaped for `MODE=start`, with a link to the PostHog issue for the stack trace.

- Dedup marker is `posthog-issue-<issue_id>`, searched across **open AND closed** issues (a fixed exception that trickles in for one more day must not reopen the ticket), and the returned bodies are re-checked for the literal marker because GitHub's search tokenizes on hyphens.
- **Fail closed**: if the GitHub search itself fails, the run aborts rather than reading "cannot read" as "nothing filed" and duplicating the whole window.
- Resolved/suppressed PostHog issues are never filed; `--max-new` (default 10) caps a bad deploy at 10 tickets per run.

### Scope notes

- **Alerts are PostHog UI configuration**, not code — new-issue/reopened notification + spike detection, documented in `docs/error-tracking.md` and listed in `Needs_From_Chris.md` along with the cron switch and the `POSTHOG_CLI_TOKEN` secret. Each is independently optional; skipping one costs only that one thing.
- The source-map upload landed in the **release** build (`build-and-push.yml` → Dockerfile) rather than `ui-build.yml`, which only runs on PRs — uploading PR builds to PostHog would be wrong, and the shipped bundle is the one that needs matching maps.

## Testing

- `tests/unit/utilities/test_error_tracking.py` — capture context/attribution/distinct_id, disabled path, never-raises, logger forwarding + the kill switch (14 tests)
- `tests/unit/app/test_celery_error_signals.py` — failure/retry handlers, user-id coercion, non-exception retry reasons, signals actually connected (13)
- `tests/unit/api/test_error_capture_middleware.py` — 500 captured with route context, `HTTPException` and healthy requests not captured (3)
- `tests/unit/test_posthog_error_issues.py` — query, parsing, planning, marker dedup, `gh` I/O and every `main()` exit code (41)
- SPA: `analytics.test.ts` covers the `capture_exceptions` config and `captureException`
- `poetry run pytest tests/unit -q` → **6628 passed**; `tsc -b` clean; the HogQL was executed against the live project to confirm the columns and aggregate syntax parse.

Closes #648

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -446,9 +446,16 @@ POSTHOG_HOST=https://us.i.posthog.com
 # Prompts/completions are redacted before they reach it (litellm_settings.turn_off_message_logging).
 # Minimum log level forwarded to PostHog (DEBUG/INFO/WARNING/ERROR/CRITICAL). Default: ERROR
 POSTHOG_LOG_LEVEL=ERROR
+# Error tracking (issue #648, docs/error-tracking.md). Both default ON and only need setting to
+# turn capture OFF: AUTOCAPTURE is the excepthook for uncaught exceptions, CAPTURE is the
+# log_error/log_critical(exc=...) forwarding. Logs are unaffected either way.
+POSTHOG_EXCEPTION_AUTOCAPTURE=true
+POSTHOG_EXCEPTION_CAPTURE=true
 # Personal API key (insight + dashboard write scope) used by scripts/posthog_dashboards.py to
 # provision the cost/margin dashboards (§E.1), and — read-only — by the daily budget alerts to read
 # LLM cache-hit rate (§E.2). Optional in the stack: unset just skips the cache-hit check.
+# Also read (scope query:read) by scripts/error_to_issues.sh, the daily error-tracking-issue ->
+# GitHub-issue cron.
 POSTHOG_PERSONAL_API_KEY=
 # PostHog project the dashboards live in, and the app host used to build their links.
 POSTHOG_PROJECT_ID=475262
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -56,6 +56,12 @@ jobs:
             VITE_API_TOKEN=${{ secrets.UI_API_TOKEN }}
             VITE_POSTHOG_KEY=${{ secrets.UI_POSTHOG_KEY }}
             VITE_POSTHOG_HOST=${{ vars.UI_POSTHOG_HOST }}
+            POSTHOG_CLI_ENV_ID=${{ vars.POSTHOG_PROJECT_ID }}
+            POSTHOG_CLI_HOST=${{ vars.POSTHOG_APP_HOST }}
+          # Personal API key with error_tracking:write — a build secret, never an ARG, so it can't
+          # be read back out of the image history. Absent = the upload step skips itself.
+          secrets: |
+            posthog_cli_token=${{ secrets.POSTHOG_CLI_TOKEN }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -228,6 +228,28 @@ ambient `llm_attribution()` scope.
 Prompts and completions are redacted (`litellm_settings.turn_off_message_logging`) — they are the
 user's own LinkedIn material, and the SPA masks the same content.
 
+### Error tracking (`$exception` → issues, issue #648)
+
+Errors reach PostHog TWICE on purpose, and the two are not redundant — see `docs/error-tracking.md`.
+
+- **Logs** (`logger.py` → PostHog Logs) keep the message and its structured context. Unchanged.
+- **`$exception`** is the grouped, fingerprinted ISSUE — what alerting and the error→GitHub-issue
+  cron are built on. Emitted by `posthog.enable_exception_autocapture` (uncaught), by
+  `log_error`/`log_critical` **when `exc=` is passed**, by the `task_failure`/`task_retry` handlers in
+  `my_celery.py`, by the unhandled branch of `api/main.py`'s `observability_middleware`, and by
+  `posthog-js` in the SPA.
+
+Use `observability.capture_exception(exc, user_id=..., **context)` for anything you catch and do NOT
+re-raise; everything else is already covered. It is safe to call twice with the same exception
+object — `posthog.capture_exception` is idempotent per instance, so the log call and the Celery
+signal do not double-count. A route's own `HTTPException` is NEVER captured: a 4xx is a response,
+not an issue.
+
+`scripts/posthog_error_issues.py` (cron: `scripts/error_to_issues.sh`) files ONE `agent:ready` GitHub
+issue per ACTIVE PostHog issue, deduped on `posthog-issue-<issue_id>` across open AND closed issues.
+It replaced the old log-grep scan, whose sha1-of-the-message dedup refiled any error with an id in
+its text every single day. Don't add a second dedup layer — the fingerprint IS the dedup.
+
 ### Browser-side analytics (SPA, issue #646)
 
 The SPA has ONE PostHog surface — `ui/src/utils/analytics.ts`. Never call `posthog` directly from a
--- a/Needs_From_Chris.md
+++ b/Needs_From_Chris.md
@@ -48,6 +48,30 @@ POSTHOG_HOST=https://us.i.posthog.com
 
 ---
 
+## PostHog Error Tracking (issue #648 — `docs/error-tracking.md`)
+
+Code is shipped and default-ON; these three are configuration only, and each is independently
+optional (skipping one costs that one thing, nothing else).
+
+- [ ] **Point the daily cron at the new scan.** The old `~/error-to-issues/scan.sh` grepped LOGS and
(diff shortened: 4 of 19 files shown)


gitchrisqueen:
🔎 Claude adversarial review — PASS (no findings, 0 fixes)

I did not write this code and did not take the PR body's claims on faith — the checks below were
executed, not read.

**Verified against the real dependency source / live services (not just the mocks):**
- **The HogQL actually parses against the live project.** Ran `build_query()`'s SQL verbatim through
  the PostHog query API on project 475262 (720h window): all eleven error-tracking columns
  (`issue_id`/`issue_name`/`issue_description`/`issue_status`/`issue_first_seen`) resolve and the
  `HAVING occurrences >= N` alias aggregate is accepted. 0 rows, as expected — nothing emits
  `$exception` until this ships. The unit tests only string-match the SQL, so this was the gap worth
  closing by hand.
- **The idempotency claim is real.** `posthog/client.py::capture_exception` short-circuits on
  `exception_is_already_captured()`, which checks a `__posthog_exception_captured` attribute on the
  instance. So `log_error(exc=e)` + re-raise + `task_failure` genuinely produces ONE occurrence.
- **`posthog.enable_exception_autocapture = …` at import is read at the right time.** `setup()` takes
  no arguments and reads module globals at *call* time (not def time), and only builds the client if
  `default_client` is falsy — so setting the flag before the first capture is correct, and it does
  **not** disturb `configure_posthog_for_worker`'s `sync_mode = True` (which depends on the client
  still being unbuilt after fork). An eager `posthog.setup()` here would have broken that; good that
  it isn't done.
- **`@posthog/cli` exists and ships the `posthog-cli` bin** (npm 0.9.1), and `POSTHOG_CLI_TOKEN` /
  `POSTHOG_CLI_ENV_ID` / `POSTHOG_CLI_HOST` are its real env names. `--mount=type=secret` is
  non-required by default, so a local/dev build with no secret takes the `else` branch instead of
  failing. `.map` deletion is in the same stage as `npm run build`, before anything copies `dist`,
  and `src/cqc_lem/ui/dist/` is gitignored — so `sourcemap: true` can't leak sources into the repo or
  the image.
- **The dedup search holds up against GitHub's real tokenizer.** I probed the live repo: `posthog`
  matches 39 issues, `posthog-issue` matches 9 (tokenized AND, none containing the literal string),
  and `posthog-issue-0198` matches 0. Because the marker carries a UUID, the AND-of-tokens search is
  effectively exact, so `gh issue list`'s default `--limit 30` cannot push the real match out of the
  window — and the literal-body re-check kills the false positives the 9-result case would otherwise
  produce. Dedup is sound in both directions.

**Seams and error paths I attacked and found correct:**
- `observability_middleware` is declared *first*, so it is the **innermost** user middleware
  (Starlette inserts at index 0) — it wraps only `ExceptionMiddleware` + router. A route's
  `HTTPException` is converted to a response before it, so 4xx never files. Confirmed by the probe
  routes in the new test, which I ran.
- No recursion loop: `logger._capture` → `observability.capture_exception` → on failure
  `log_warning`, which never captures. `observability.py` has zero `log_error(exc=)` call sites.
- `capture_exception(exc, **props)` can't collide on `exc`/`user_id` — both are consumed as named
  params of `log_error` before `**context` is built.
- `on_task_failure`'s `kwargs`/`sender` shapes match what celery actually sends from BOTH senders
  (`app/trace.py:267` and `worker/request.py:645`); `exc` is an exception instance on every path
  through `Request.on_failure` (`ExceptionWithTraceback` is unwrapped first), so the retry handler's
  extra `isinstance` guard is the only place it was needed.
- `scripts/error_to_issues.sh`'s `sudo -n grep /opt/lem/.env` idiom matches the three existing crons
  (`weekly_model_check.sh`, `weekly_linkedin_version_check.sh`, `perf_snapshot.sh`); `print()` in
  `scripts/` matches `model_health_check.py` / `posthog_dashboards.py` — neither is a CLAUDE.md
  violation.

**Tests:** ran the full unit suite against this branch — **6628 passed** (the 71 new tests included).
No regression from `log_error`/`log_critical` now reaching into `observability`.

Three non-blocking operational notes for @gitchrisqueen (no code change — each is either upstream
behaviour or a judgement call that isn't mine):

1. **Autocapture arms lazily.** `posthog-python` only installs its excepthook when the client is
   *constructed*, which is on the first PostHog call in that process — posthog's own source carries a
   `TODO` saying exactly this. So a process that crashes *before* emitting any telemetry is still
   missed. In practice every LEM service emits early (`track_api_call` / `track_task`), so this only
   affects import-time/boot crashes. `docs/error-tracking.md`'s "Any Python process" is a touch
   stronger than what the SDK delivers.
2. **Auto-filed issues sort last in the agent pipeline.** `tick.sh::select_next_issue` orders by
   milestone number first; an issue with no milestone falls to `999`, behind every roadmap issue,
   regardless of `priority:`. So a production error won't be picked up until the milestone backlog
   drains. Fixing that means assigning a milestone (or reordering the pipeline) — a scheduling call,
   not a code defect.
3. **Enabling autocapture replaces `threading.excepthook` without chaining it**, so an uncaught
   exception in a non-main thread stops printing its traceback to stderr. LEM only spawns threads in
   `selenium_load_test.py` (a manual CLI, and via `ThreadPoolExecutor`, which captures into futures),
   so the practical impact here is nil — noting it in case that changes.

Nothing here rises to a blocking finding, so no commits were pushed. CI is green on `5f7ada3`.

gitchrisqueen:

PR #681, opened 2026-07-26
feat(observability): error-triggered + sampled session replay (closes #649)

### LINKED ISSUE #649: PH4 — Session replay: error-triggered + sampled SPA recording
## Context (depends on PH1 #646)
Debugging SPA issues (Content Studio 422s, approval-flow confusion, Save-All-hidden bugs like #596) currently needs repro steps from the user. PostHog session replay (5K recordings/mo free) supports privacy masking, console/network capture, and RECORDING RULES: record only sessions that hit an error, plus a small random sample.

## Scope
- Enable replay in the posthog-js init: mask_all_inputs + data-ph-mask respected (PH1's attributes), canvas capture OFF, console log capture ON, network performance ON (no request bodies).
- Recording rules: trigger on $exception + 10% sample of normal sessions; minimum duration filter to skip bounces.
- Feedback link-up: FeedbackWidget already sends posthog_session_id — include a direct replay URL in the feedback→GitHub issue body so every feedback report links its replay.
- Document the privacy posture in docs (who is recorded: the app's own users; what is masked).

## Acceptance
- A forced SPA error produces a watchable replay reachable from the error-tracking issue; feedback issues include replay links; masked fields verifiably masked in a test recording.

gitchrisqueen:
## What & why

PH4 of the PostHog milestone. Debugging an SPA report (Content Studio 422s, approval-flow confusion, Save-All-hidden bugs) started with *"can you tell me the steps?"*. A replay **is** those steps — and the recordings kept here are the ones worth watching.

### Recording rules (in the SDK, not project settings)

One testable place, `ui/src/utils/analytics.ts`:

| Rule | How | Default |
|---|---|---|
| Sample of ordinary sessions | `session_recording.sampleRate` | `0.1` (`VITE_POSTHOG_REPLAY_SAMPLE`) |
| Every session that throws | `posthog.on('eventCaptured')` → `startSessionRecording({ sampling: true })` | always |
| Skip bounces | PostHog **minimum duration** + `strictMinimumDuration: true` | 5s (project setting) |
| Kill switch | `VITE_POSTHOG_REPLAY=false` | replay on |

The error trigger fires on the `$exception` event itself, so it covers **both** posthog's own unhandled-error autocapture and anything the app catches and reports via `captureException()`, and it **overrides sampling** — that override is the point. The project's own sampling must stay at 100%: the local `sampleRate` takes precedence, so configuring both would multiply them (called out in the doc and in CLAUDE.md).

Minimum duration is the one knob that is remote config, so it is documented rather than coded; `strictMinimumDuration: true` makes PostHog measure it against the recorded buffer instead of wall-clock session age, so a bounce split over two page loads is still a bounce.

### Privacy posture

- `maskAllInputs: true` — every input/textarea/select value is masked **before it leaves the browser**.
- `maskTextSelector: [data-ph-mask]` — `maskProps()` already marks every content editor; its text is masked too.
- Network: `network_timing` only. `recordHeaders: false` (no session token), `recordBody: false` (no post/DM content).
- `captureCanvas.recordCanvas: false` — carousel previews are expensive and never the bug.
- `enable_recording_console_log: true` (into the masked, access-controlled recording) — deliberately still NOT `capture_console_errors`, which would turn a console line into a fingerprinted `$exception`.
- No `VITE_POSTHOG_KEY` → posthog-js is still a lazy chunk the browser never fetches.

### Every report links its replay

`FeedbackWidget` already sent `posthog_session_id`; it now becomes a link:

- **auto-filed feedback issue** — "Watch the session replay" above `## Scope`
- **the `+1` comment** a repeat report leaves — that reporter's OWN session, often the one that explains what the first report couldn't
- **error-tracking issues** (`scripts/posthog_error_issues.py`) — the query now reads `$session_id`, so a browser exception's GitHub issue links the session it was thrown in; backend exceptions have no session and omit the line

`observability.session_replay_url()` builds it from `POSTHOG_PROJECT_ID` + `POSTHOG_APP_HOST`. A session id that isn't the SDK's uuid-ish shape, or no project configured, **omits the line** — the value is pasted into GitHub markdown, and a guessed link is worse than none. The widget now also says so to the user in plain language.

## Testing notes

- **vitest** (`analytics.test.ts`): the init config, the env-driven sample rate incl. clamping (`7` → `1`, `lots` → default), the error trigger, the already-recording and non-`$exception` no-ops, a throwing `startSessionRecording` not reaching the page, and the `VITE_POSTHOG_REPLAY=false` build.
- **pytest**: `session_replay_url` shaping + rejection of unshaped ids, the issue body, the dedup comment, both orchestrator paths (filed + deduped), and the error-filer's new column/link. `6659 passed` locally; `tsc -b` and `eslint` clean.
- vitest could not be run in this worktree (local node is 18, the toolchain needs 22) — CI's `UI Build` job is the check.

Manual verification steps (forced error → watchable replay, masked fields, feedback link) are in `docs/session-replay.md`.

## Config

New BUILD-time vars, wired the same way as `VITE_POSTHOG_KEY` (Dockerfile ARG + `ui-build.yml` / `build-and-push.yml` from repo vars `UI_POSTHOG_REPLAY`, `UI_POSTHOG_REPLAY_SAMPLE`). Both unset = replay on at 10% + every errored session, so no repo config is required to ship this.

**One action outside this PR:** set Replay → **minimum duration = 5000 ms** and leave sampling at 100% in the PostHog project settings.

Closes #649

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -467,6 +467,12 @@ POSTHOG_APP_HOST=https://us.posthog.com
 # This is the PROJECT key (phc_…), which is public by design — never the personal API key.
 VITE_POSTHOG_KEY=
 VITE_POSTHOG_HOST=https://us.i.posthog.com
+# --- Session replay (issue #649, docs/session-replay.md) ---
+# Also BUILD-time. Replay is ON by default: VITE_POSTHOG_REPLAY_SAMPLE of ordinary sessions plus
+# EVERY session that produces an $exception (the sample is overridden for those). Set
+# VITE_POSTHOG_REPLAY=false to ship a bundle that never records at all.
+VITE_POSTHOG_REPLAY=true
+VITE_POSTHOG_REPLAY_SAMPLE=0.1
 
 # --- Unit economics / margin report (docs/cost-performance-margin-plan.md §C.1) ---
 # Monthly price per subscription tier, used to compute MRR for contribution margin. Defaults match
--- a/.env.prod.example
+++ b/.env.prod.example
@@ -150,6 +150,10 @@ POSTHOG_LOG_LEVEL=ERROR
 # them here does nothing; they are passed as docker build-args (CI: UI_POSTHOG_KEY secret).
 # VITE_POSTHOG_KEY=phc_...
 # VITE_POSTHOG_HOST=https://us.i.posthog.com
+# Session replay (issue #649) — same build-arg story (CI vars UI_POSTHOG_REPLAY /
+# UI_POSTHOG_REPLAY_SAMPLE). Unset = replay on at a 10% sample + every errored session.
+# VITE_POSTHOG_REPLAY=true
+# VITE_POSTHOG_REPLAY_SAMPLE=0.1
 
 # =============================================================================
 # Tunnel & Networking (Ngrok) — DISABLED in prod (Cloudflare Tunnel is used)
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -56,6 +56,8 @@ jobs:
             VITE_API_TOKEN=${{ secrets.UI_API_TOKEN }}
             VITE_POSTHOG_KEY=${{ secrets.UI_POSTHOG_KEY }}
             VITE_POSTHOG_HOST=${{ vars.UI_POSTHOG_HOST }}
+            VITE_POSTHOG_REPLAY=${{ vars.UI_POSTHOG_REPLAY }}
+            VITE_POSTHOG_REPLAY_SAMPLE=${{ vars.UI_POSTHOG_REPLAY_SAMPLE }}
             POSTHOG_CLI_ENV_ID=${{ vars.POSTHOG_PROJECT_ID }}
             POSTHOG_CLI_HOST=${{ vars.POSTHOG_APP_HOST }}
           # Personal API key with error_tracking:write — a build secret, never an ARG, so it can't
--- a/.github/workflows/ui-build.yml
+++ b/.github/workflows/ui-build.yml
@@ -31,3 +31,5 @@ jobs:
           VITE_API_TOKEN: ${{ secrets.UI_API_TOKEN }}
           VITE_POSTHOG_KEY: ${{ secrets.UI_POSTHOG_KEY }}
           VITE_POSTHOG_HOST: ${{ vars.UI_POSTHOG_HOST }}
+          VITE_POSTHOG_REPLAY: ${{ vars.UI_POSTHOG_REPLAY }}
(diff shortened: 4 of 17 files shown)


gitchrisqueen:

Verified against the actual `posthog-js@1.407.2` sources (unpacked the package and read `lazy-loaded-session-recorder.ts` out of its sourcemap), not against the docs.

- **The `$exception` trigger never fired for the sessions it exists for** (blocks the issue's first acceptance criterion). The hook was gated on `posthog.sessionRecordingStarted()` → `SessionRecording.started` → `!!this._stopRrweb`, i.e. *is rrweb attached*. `start()` calls `_startRecorder()` unconditionally after `makeSamplingDecisions()`, so rrweb attaches for **every** session and sampling only governs whether the buffer is ever sent (a sampled-out session just `_clearBuffer()`s on each emit, `lazy-loaded-session-recorder.ts:1364`). The guard therefore read `true` in exactly the sampled-OUT case, and ~90% of errored sessions were silently never recorded. Replaced with an explicit once-per-session-id flag — same "don't snapshot on every exception in a loop" intent, re-arms on a new session id. The new test fails against the old guard (checked).
- **Feedback replay links resolved for only ~10% of reports**, so "every feedback report links its replay" (issue Scope) was false for most of them and the issue body promised a link that 404s. Opening the widget now calls the same `ensureSessionRecorded()`; the widget copy and the issue-body line were reworded to stop claiming lead-up a forced recording cannot contain (the buffer is discarded while sampled out — that part of the doc is now accurate too).
- **Docs shipped a feature that could never record**: `disable_session_recording: false` is a veto, not a switch — posthog-js only starts the recorder when the project's remote config comes back `enabled`. Added "Record user sessions must be ON" to the manual-steps section (it previously listed only minimum duration + sampling), plus the `sessionRecordingStarted()` gotcha in `docs/session-replay.md` and `CLAUDE.md`.

Checked and found sound, no change needed: the `session_recording` keys are all real in this version (`sampleRate`, `strictMinimumDuration`, `captureCanvas.recordCanvas`); local `sampleRate` does take precedence over remote (`o = local ?? remote`); `startSessionRecording({sampling:true})` persists the override when the recorder chunk hasn't loaded yet; `recordHeaders: false` is what keeps `X-Session-Token` out (the token is a header, never a query param) and the SPA has no `console.*` calls that could carry content into `enable_recording_console_log`; `SESSION_ID_RE` can't break out of the URL path or inject markdown; `parse_rows` tolerates the new trailing column; `replay_url_from_context` runs after `context_json` is parsed and is dict-guarded.

Verification: `vitest` 85/85 (had to fetch a Node 22 toolchain — Node 18 can't load the config), `tsc -b` and `eslint` clean on the touched files, and 275 pytest cases across `test_feedback_issue_service` / `test_observability` / `test_posthog_error_issues`.

gitchrisqueen:

PR #682, opened 2026-07-26
feat(observability): KPI funnels, consolidated dashboards, threshold alerts + weekly report (closes #650)

### LINKED ISSUE #650: PH5 — Business KPI funnels, consolidated dashboards, threshold alerts + weekly report subscription

gitchrisqueen:
## What & why

PostHog had 41 insights across 6 dashboards and still couldn't answer two questions: **do LEM's business loops close?** and **did something break overnight?** Every metric was a point reading, every watchdog was a hand-run cron, and the 2-week perf report was a human remembering to run it.

`scripts/posthog_provision.py` defines that surface as code — idempotent, `--dry-run` by default:

- **LEM Health** — Celery failures (count, rate, by task name), LinkedIn 429 breaker trips, proxy-reported LLM spend/day, posts published/day, follower delta, API error rate + p95.
- **LEM Growth** — the content-loop funnel (`content_plan_generated` → `post_approved` → `post_outcome`), the `signup_started` → … → `subscription_started` funnel, onboarding step drop-off, the comment → author-reply engagement loop, comment visibility and ER/audience trends.
- **Four threshold alerts** (of PostHog's 5 free): comments/week below floor, LLM spend/day over cap, Celery failure spike, 429 spike — emailed to the key owner.
- **One weekly email subscription** of LEM Growth, Monday 09:00 UTC. That is what replaces the hand-run perf-report cadence.

It sits **on top of** `scripts/posthog_dashboards.py` (cost/margin), which it does not replace — Health/Growth is the top level, the cost set is the drill-down.

## The one instrumentation change

The 429 alert had no signal to read. A breaker trip only ever produced a `log_warning`, and PostHog forwards `ERROR` and up by default — so the single event that says *LinkedIn is throttling us* was invisible to dashboards and alerts alike. `mark_rate_limited()` now also emits `rate_limit_trip` (`cooldown_seconds`, `consecutive_trips`, `reason`) via `observability.track_rate_limit_trip`. System-scoped (LinkedIn throttles by egress IP, so a trip is account-wide), and it never raises — the breaker must still open when analytics is down.

## Decisions worth reviewing

- **Alertable tiles are native single-series `TrendsQuery` insights.** PostHog can alert on trends/funnel/SQL insights, but the threshold is evaluated against one series by index — a breakdown or second series makes `series_index: 0` ambiguous.
- **They filter on the `state` STRING, not the `success` boolean.** A boolean property filter that silently matches nothing produces an alert that never fires, which is worse than no alert.
- **The comment floor is weekly, not daily.** Human pacing (#626) takes account-wide rest days; a daily floor would page on a perfectly healthy zero-comment day.
- **Comments are counted from `comment_outcome`** (the T+24h sweep) — the only per-comment event that exists. It lags a day; that's documented, not hidden.
- **Money tiles read `$ai_generation`, never `llm_call`** (docs/llm-analytics.md) — a test asserts no tile mixes the two streams.
- **Insight names must stay unique across both provisioner scripts.** Each plans by name against the same project, so a shared name would make them fight over one insight forever — a unit test fails the build if they ever collide.
- **Subscriptions diff on (dashboard, recipient, frequency), never `start_date`**, which PostHog advances on every send — diffing it would recreate the subscription and mail a duplicate every run.
- **Engagement loop is a HogQL conversion tile, not a funnel.** Only comment → author-reply is instrumented today; connection-accepted and DM-reply join it as steps when that data lands, rather than shipping a funnel with two permanently-empty steps.

## Notes for the owner

- Several tiles read events that exist in code but haven't flowed into the project yet (`comment_outcome`, `audience_snapshot`, the funnel events, the SPA's `post_approved`/`content_plan_generated`). They render empty until those land — same deliberate behaviour as the cost dashboards. **The alerts are the exception**: run `--apply` once those events are actually flowing, or the comments-floor alert reads 0 and fires on its first evaluation.
- PostHog renders at most 6 tiles into a subscription email, so Growth's tile order is the email's priority order; audience growth is the one that falls off the bottom.
- New env: `POSTHOG_REPORT_EMAIL` (falls back to the API key owner's email). The personal key needs `insight`/`dashboard`/`alert`/`subscription` read+write plus `user:read`.

## Testing

- `tests/unit/test_posthog_provision.py` — 66 tests over the specs, funnel shapes, alert/threshold evaluation, the planner (create/update/unchanged/blocked), `apply_actions` dry-run vs apply, and `main()` end to end with a mocked client.
- `--simulate 'NAME=VALUE'` runs the same comparison PostHog runs server-side, which is how the "fires on synthetic threshold breach" criterion is provable without live data.
- New coverage for `track_rate_limit_trip` and for the breaker emitting it (including that a tracking failure never stops the cooldown being written).
- Full unit suite green locally: **6730 passed**.

Docs: `docs/kpi-dashboards.md`, plus a CLAUDE.md subsection.

Closes #650

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -460,6 +460,10 @@ POSTHOG_PERSONAL_API_KEY=
 # PostHog project the dashboards live in, and the app host used to build their links.
 POSTHOG_PROJECT_ID=475262
 POSTHOG_APP_HOST=https://us.posthog.com
+# Recipient of the weekly LEM Growth dashboard email provisioned by scripts/posthog_provision.py
+# (issue #650, docs/kpi-dashboards.md) — the automated replacement for the hand-run perf report.
+# Unset falls back to the personal API key owner's own email.
+POSTHOG_REPORT_EMAIL=
 # --- PostHog in the SPA (issue #646) ---
 # Read by Vite at BUILD time and baked into the bundle, so they must be set for `npm run build`
 # (docker build-arg / CI env), not in the running container. Leave VITE_POSTHOG_KEY empty and the
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -305,6 +305,25 @@ Every report links its recording: `session_replay_url()` in `observability.py` t
 uuid-ish, or a missing `POSTHOG_PROJECT_ID`, omits the line rather than guessing a URL. Full posture
 (who is recorded, what is masked, how to verify): `docs/session-replay.md`.
 
+### KPI funnels, dashboards, alerts + weekly report (issue #650)
+
+`scripts/posthog_provision.py` is the ONE place the business-KPI surface is defined — two
+consolidated dashboards (**LEM Health**: task failures, 429 trips, LLM spend/day, posts/day,
+follower delta, API errors · **LEM Growth**: the content-loop and signup→subscription funnels,
+onboarding drop-off, the comment→reply loop, ER/audience trends), four threshold alerts, and one
+weekly email subscription of Growth that replaces the hand-run perf-report cron. `--dry-run`
+(default) diffs, `--apply` converges, `--simulate 'NAME=VALUE'` proves an alert's threshold without
+waiting for a real breach. Details: `docs/kpi-dashboards.md`.
+
+It sits ON TOP of the cost/margin set (`scripts/posthog_dashboards.py`), which it does not replace.
+Both plan by insight NAME against the same project, so **insight names must stay unique across the
+two scripts** — a unit test fails the build if they collide. Alert-bearing tiles must stay native
+single-series `TrendsQuery` insights (a threshold is evaluated against one `series_index`), and they filter on STRING
+properties (`celery_task.state = 'FAILURE'`) rather than booleans: a boolean filter that matches
+nothing yields an alert that never fires. Money tiles read `$ai_generation`, never `llm_call`.
+`mark_rate_limited()` emits `rate_limit_trip` (issue #650) because the breaker's WARNING log never
+reaches PostHog at the default `POSTHOG_LOG_LEVEL`.
+
 ## CI Gates
 
 Before merging any PR, all of the following must pass:
--- a/docs/kpi-dashboards.md
+++ b/docs/kpi-dashboards.md
@@ -1,0 +1,125 @@
+# KPI funnels, dashboards, alerts and the weekly report (issue #650)
+
+LEM's PostHog project had 41 insights across 6 dashboards and still could not answer two questions:
+*do the business loops close?* and *did something break overnight?* Everything was a point metric,
+every watchdog was a hand-run cron, and the 2-week perf report was a human remembering to run it.
+
+`scripts/posthog_provision.py` defines the answer as code: two consolidated dashboards, the funnels
+over LEM's actual loops, four threshold alerts and one weekly email. Re-running it re-creates
+anything edited away in the UI, and does nothing at all when the project already matches.
+
+```bash
+# what would change (default; no writes)
+poetry run python scripts/posthog_provision.py --dry-run
+# create/update dashboards, insights, alerts and the weekly subscription
+poetry run python scripts/posthog_provision.py --apply
+# every tile's query, no network
+poetry run python scripts/posthog_provision.py --print-sql
+# would this reading page someone?
+poetry run python scripts/posthog_provision.py --simulate 'LEM — LinkedIn 429 spike=9'
+```
+
+Exit codes: `0` in sync / applied, `2` changes pending (dry run), `1` error.
+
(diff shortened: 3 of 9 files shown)


gitchrisqueen:

Reviewed against the live PostHog project (`CQC LEM`, 475262): I executed **all 10 HogQL tiles**, both funnel nodes and the alert-bearing `TrendsQuery` series through the read-only SQL/query API rather than trusting the unit tests, which never touch a HogQL parser. Nine tiles, both funnels and the Celery-failure alert series are correct (the alert filter returns the same 7 failures on 2026-06-30 that the HogQL failure-rate tile does, so the `state` STRING filter genuinely matches). One tile did not survive.

- **`Health — Follower delta per day` was invalid HogQL — the tile would have rendered an error, permanently.** `sum(followers) AS followers` re-binds the name its sibling `sum(followers - prev)` reads; ClickHouse rejects the whole query with *"Aggregate function sum(followers) AS followers is found inside another aggregate function"*. Not a wrong number — no number at all, and nothing in the suite would have noticed. Aliased to `total_followers` and re-ran it against the project: parses and executes. Added a regression test that fails any tile aliasing an aggregate to a column it reads, so the class is closed, not just the instance.
- **An alert created without a subscriber emailed nobody and was never repaired.** The script deliberately tolerates a key without `user:read` ("alerts will have no subscriber"), but `plan_alerts` diffed only bounds/insight/enabled — so once created empty, every later run with a *working* key reported `unchanged_alert` and the alert stayed silent forever. That is the exact silent failure the alerts exist to prevent. Subscribers now join the diff, `list_alerts` reports them (normalizing PostHog's nested-object-on-read / bare-id-on-write shapes, same as `insight`), and an update **unions** rather than replaces, so a recipient the owner added in the UI isn't dropped. A run that can't name the owner still reports no drift, so `--dry-run` doesn't exit 2 forever on something it couldn't fix.
- **A telemetry failure in `mark_rate_limited()` logged "Failed to set LinkedIn 429 circuit breaker".** The new `track_rate_limit_trip` call sits inside the breaker's outer `try`, so an import or capture failure fell into that handler and claimed the breaker failed — while it was in fact open. Whoever is debugging a doom loop would go looking for the wrong thing. Isolated into its own `try` with an accurate message; the existing "never breaks the breaker" test now also asserts the message.

Checked and found sound, for the record: every event and property the tiles read exists in `observability.py` / `analytics.ts` (`celery_task.state` does arrive, via `**extra` from the `task_postrun` handler); `$ai_total_cost_usd` matches `docs/llm-analytics.md`; no tile reads `llm_call`; insight names don't collide with `posthog_dashboards.py`; `next_report_start` handles the Monday-boundary cases correctly.

One caveat I could not settle without writing to your PostHog project, so I didn't: `plan_actions` compares the stored `query` for exact equality, and this PR is the first to provision **native** `TrendsQuery`/`FunnelsQuery` insights. Existing trends insights in the project carry server/UI-added keys (`version: 4`, `properties: []`, `filterTestAccounts: false`) that the spec doesn't send. If the API adds those on create, those 6 tiles would report `update_insight` on every run and `--dry-run` would never return 0 — idempotent in effect, but noisy, and short of the acceptance criterion's letter. Worth watching on the first `--apply`; the HogQL tiles (the shipped `posthog_dashboards.py` pattern) are unaffected either way.

Unit tests: 217 passed (`test_posthog_provision.py`, `test_rate_limit.py`, `test_observability.py`).

gitchrisqueen:

PR #683, opened 2026-07-26
fix(db): stop partial engagement-prefs updates from wiping the row (closes #639)

### LINKED ISSUE #639: set_default_video_quality wipes the user's entire engagement_preferences row to defaults
## Bug
`db.py set_default_video_quality()` calls `update_engagement_preferences(user_id, {'default_video_quality': quality})` with a single-key dict. `update_engagement_preferences` merges `{**_ENGAGEMENT_DEFAULTS, **prefs}` and upserts EVERY column in `_ENGAGEMENT_COLS` — so any caller passing a partial dict silently resets all other preferences (tone, targeting lists, caps, goals, reply modes…) to code defaults. The single-row upsert design means one sparse caller destroys the whole row.

## Fix
- In `update_engagement_preferences`, merge over the user's EXISTING row (get first), not over `_ENGAGEMENT_DEFAULTS`, when a row exists — defaults only for a genuinely new row. That makes partial updates safe for every current and future caller.
- Audit existing callers: `api/main.py` (full object from SPA — fine), `utilities/brand_account.py:124` (already merges existing — fine), `db.py set_default_video_quality` (broken partial).
- Add a regression unit test: seed a full prefs row, call `set_default_video_quality`, assert every other field survives.

## Context
Found 2026-07-26 while tuning user 1's targeting prefs. Related single-row-upsert prior incident: V52 tone-overflow rollback (see CLAUDE.md gotchas).

gitchrisqueen:
## What & why

`db.update_engagement_preferences()` built its write set as `{**_ENGAGEMENT_DEFAULTS, **prefs}` and then upserted **every** column in `_ENGAGEMENT_COLS`. Because `engagement_preferences` is a single row per user, any caller passing a partial dict silently reset every field it didn't mention — tone, include/exclude topics & keywords & authors, focus topics, business/personal goals, per-day caps, reply modes, quality-gate thresholds, cadence — back to code defaults.

`db.set_default_video_quality()` is exactly that caller: `update_engagement_preferences(user_id, {'default_video_quality': quality})`.

## The fix

- Extracted the row read into `_select_engagement_row(user_id)`, which returns the decoded row or `None` when the user has never saved one, and lets `mysql.connector.Error` escape. `get_engagement_preferences()` is now a thin wrapper over it, so the JSON/bool/NULL-column decode logic stays in exactly one place (no behaviour change to that function).
- `update_engagement_preferences()` now merges over the user's **saved row**; `_ENGAGEMENT_DEFAULTS` is the base only for a genuinely new row. Keys explicitly present in `prefs` still win, so `None`/`[]`/`False` can still clear a field.
- A **read failure aborts the write** (returns `False`) instead of falling through to "write all 39 columns as defaults" — a failed `SELECT` causing the same data loss would defeat the fix. The documented `mysql.connector.Error → False` contract for this helper is unchanged.

## Caller audit

| Caller | Before | After |
|---|---|---|
| `api/main.py` PUT `/user/engagement-preferences` | full object from the SPA — fine | unchanged |
| `utilities/brand_account.py:124` | already passed `{**existing, **overrides}` — fine | unchanged (the belt-and-braces merge is now redundant, left in place) |
| `db.set_default_video_quality` | **wiped the row** | preserves everything else |

All validation/clamping (reply modes, ICP score, sweeps, cadence, catch-up cap + per-plan allowance, gate thresholds, ENUM event types) runs after the merge exactly as before, so a stale/invalid stored value is still normalised on write.

## Testing

New `TestPartialUpdateKeepsTheRest` in `tests/unit/utilities/test_db_engagement_prefs.py`:
- a stored-row fixture asserted to cover **every** `_ENGAGEMENT_COLS` entry (so a future column can't quietly escape the regression test);
- `set_default_video_quality` round-trip — the quality changes, all 38 other columns persist byte-identical;
- same for a bare `{"tone": ...}` update;
- explicit `[]` / `None` / `False` still override the saved row while neighbours survive;
- a brand-new user (no row) still gets the code defaults;
- an unreadable row returns `False` and issues no `INSERT`.

`poetry run pytest tests/unit -q` → **6741 passed**.

Closes #639

### DIFF
--- a/src/cqc_lem/utilities/brand_account.py
+++ b/src/cqc_lem/utilities/brand_account.py
@@ -111,8 +111,11 @@ def sync_brand_preferences(phase: Optional[str] = None) -> Optional[dict]:
     """Push the current phase's outbound policy onto the brand account's engagement preferences.
 
     Returns the applied overrides, or None when there is no brand account to sync. The whole upsert
-    is one row (the V52 incident), so the existing prefs are merged in rather than replaced — voice,
-    tone and targeting the owner set stay exactly as they are.
+    is one row (the V52 incident), so ONLY the policy fields are sent — voice, tone and targeting the
+    owner set are preserved by `update_engagement_preferences`, which merges over the saved row and
+    aborts when it can't read it (issue #639). Re-sending `existing` here would defeat that abort:
+    a transient read error makes `get_engagement_preferences` return code defaults, and this nightly
+    task would then write all 39 of them over the brand account's real settings.
     """
     user_id = get_brand_user_id()
     if user_id is None:
@@ -121,7 +124,7 @@ def sync_brand_preferences(phase: Optional[str] = None) -> Optional[dict]:
     resolved_phase = (phase or current_launch_phase()).strip().upper()
     existing = get_engagement_preferences(user_id) or {}
     overrides = brand_preference_overrides(existing, resolved_phase)
-    if not update_engagement_preferences(user_id, {**existing, **overrides}):
+    if not update_engagement_preferences(user_id, overrides):
         log_warning("Could not sync brand account preferences", user_id=user_id)
         return None
     log_debug(f"Brand account synced to phase {resolved_phase}", user_id=user_id)
--- a/src/cqc_lem/utilities/db.py
+++ b/src/cqc_lem/utilities/db.py
@@ -3187,9 +3187,12 @@ def _coerce_json_list(value) -> list:
         return []
 
 
-def get_engagement_preferences(user_id: int) -> dict:
-    """Return the user's engagement preferences (voice/targeting/caps) with code-level
-    defaults when no row exists — so behaviour is unchanged until the user customizes."""
+def _select_engagement_row(user_id: int) -> Optional[dict]:
+    """The user's SAVED engagement row, decoded — or None when they have never saved one.
+
+    Deliberately lets `mysql.connector.Error` escape: a read failure is not the same as a missing
+    row, and `update_engagement_preferences` must be able to tell them apart before it rewrites
+    every column (issue #639)."""
     connection = get_db_connection()
     cursor = connection.cursor(dictionary=True)
     try:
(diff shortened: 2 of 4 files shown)


gitchrisqueen:

- **The caller audit was wrong about `brand_account.sync_brand_preferences` (real data loss, still live after the fix).** The PR table marks it "already passed `{**existing, **overrides}` — fine". It isn't: `existing` comes from `get_engagement_preferences()`, which **swallows** `mysql.connector.Error` and returns a full dict of `_ENGAGEMENT_DEFAULTS`. Spreading that into `prefs` puts all 39 columns into the explicit-keys layer, so the new saved-row merge is bypassed *and* the new unreadable-row abort can never trigger — one transient MySQL hiccup during the nightly `auto_sync_brand_account` beat resets the brand account's tone, targeting, goals, reply modes and caps to defaults. Fixed by sending only the policy fields it owns (`update_engagement_preferences`, `brand_account.py:124`); the db layer preserves everything else. `test_merges_phase_policy_onto_existing_prefs` was asserting the now-removed pass-through, so it is rewritten as `test_sends_only_the_policy_fields`, plus a new `test_unreadable_prefs_cannot_reset_the_whole_row`.
- **New abort path was invisible (CLAUDE.md convention).** The added failure branch used `myprint`, which CLAUDE.md marks as a legacy shim to avoid in new code and which logs at INFO — below the default `POSTHOG_LOG_LEVEL=ERROR`. A read failure that silently cancels a user's settings save would never surface as an issue. Now `log_error(..., exc=err, user_id=user_id)`.

Verified and found sound (no change needed): `get_engagement_preferences` is behaviourally identical after the extraction (missing row and `mysql.connector.Error` both still yield `dict(_ENGAGEMENT_DEFAULTS)`, and the `finally` still closes cursor+connection on the escaping error); `set(_ENGAGEMENT_COLS) == set(_ENGAGEMENT_DEFAULTS)` (39/39), so the `k in _ENGAGEMENT_DEFAULTS` merge filter can't drop a selected column and `_val`'s `merged[col]` can't `KeyError`; the JSON/bool round-trip through `_coerce_json_list` → `json.dumps` is byte-identical for every stored shape in the new fixture; `catchup_event_types`/`catchup_message_source`/`catchup_touch_mode`/`posts_per_week`/ICP/sweep clamps all still run *after* the merge, so a stale stored value is still normalised; and `EngagementPreferencesRequest.model_dump()` carries all 39 keys, so the SPA PUT is still a full overwrite and can still clear a field to `None`/`[]`/`False`.

`poetry run pytest tests/unit -q` → **6742 passed**.

gitchrisqueen:

Issue #629, opened 2026-07-26
D5 — Suppression tripwire: impression step-collapse auto-pauses engagement automation


gitchrisqueen:
## Context
2026 LinkedIn penalties are SILENT: flagged accounts see step-function reach collapse (documented pattern: 8,500→340 impressions overnight) with no notification; recovery takes 60–90 days of clean behavior. LEM already survived one 429 doom loop that paused commenting for days. A day of automation is worth far less than a month of suppressed reach — the system must detect collapse and stop itself.

## Scope
- Daily check (beat task): compare each new post's early impressions and the account's trailing engagement velocity vs the trailing 14-day median (post_stats history). Trip condition: >70% sustained drop across ≥3 consecutive days (tunable env thresholds), or D4's comment-demotion rate exceeding threshold.
- On trip: pause_automation() (existing Redis breaker in rate_limit.py — engagement tasks only; scheduled posts and analytics capture continue), log_critical → PostHog, and notify the user via the existing notification/nudge surface with a plain-language explanation and a manual re-enable path in the UI.
- Guard against false trips: require minimum sample (≥3 posts in window), ignore days with no posts, account for the platform-wide decline baseline (compare against own trailing median, not absolute numbers).
- Expose tripwire state via /api/app-info-adjacent status endpoint + Account UI banner.

## Acceptance
- Unit tests: trip/no-trip scenarios incl. sparse posting, first-week cold start, recovery clearing.
- Never auto-resumes; human re-enable only.

Issue #401, opened 2026-07-23
E4 — Reply-to-every-comment golden-hour amplifier


gitchrisqueen:
## Why
Replying to every comment doubles comment count and is a high-weight, low-risk 2026 lever.

## Scope
- [ ] Ensure own-post comments get a timely, substantive reply within the golden hour (extend `sweep_reply_comments`, `run_automation.py:1699`).
- [ ] Prioritize the golden-hour window; respect throttle.

## Files
`app/run_automation.py` (~L1699), `app/run_scheduler.py` (cadence).

## Acceptance
New comments on own posts receive a substantive reply within the golden-hour window in an E2E/live pass.

---
_Part of the LEM 2026 Growth Roadmap (gap analysis vs. 2026 LinkedIn algorithm authorities). Follows repo norms: feature branch `feature/claude-<task>`, ≥80% patch coverage, migrations advance from V56, CI gates green before merge._

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #684, opened 2026-07-26
feat(engagement): golden-hour report + 6-8h second-wave self-comment (closes #622)

### LINKED ISSUE #622: G7 — Golden-hour presence hardening: reply-all + second-wave self-comment
## Context
Author replies in the first 60–90 min multiply distribution (threaded replies ≈2.4x reach; top-1% creators reply 741% more than average). #401 shipped a reply-to-every-comment amplifier, but in 14 days of logs only 2 replies landed — audit why (429 history, sweep cadence, or no comments to reply to). Alić's playbook adds a second wave: repost/self-comment 6–8h later plus pinned self-comments with added insight.

## Scope
- Audit + harden the #401 amplifier: verify event-mode reply sweeps actually fire within 60–90 min of a post going live; log a per-post golden-hour report (comments found, replies sent, latency).
- Second-wave: one scheduled self-comment 6–8h after posting that ADDS a substantive insight (tip #8 for our own post — goes through the G2 quality contract; grounded in the post + story bank, never 'thanks for reading').
- Seed comment (#344 grounding) and second-wave must not collide — cap self-comments per post at 2 total.
- Respect rate_limit.py breaker and human-pacing (D2) once available.

## Acceptance
- Unit tests for sweep timing decision, second-wave scheduling, self-comment cap.
- PostHog event per golden-hour report for later analysis.

gitchrisqueen:
## What & why

**Audit half.** #401 spread several reply sweeps across the hour after each post publishes — and then nothing recorded whether they fired. Fourteen days of logs held two replies, and a late sweep, a rate-limited one and a post nobody commented on all looked identical. Each has a different fix, so this makes the amplifier answerable:

- New `utilities/golden_hour.py` — the ONE place the timing is decided (the sweep countdowns move out of the 6k-line task module so the decision is unit-testable), plus the latency verdict, the report builder, the second wave's schedule and the self-comment cap. Pure functions, no DB/LLM/Celery — the `story_bank.py` / `suppression.py` pattern.
- **One `golden_hour_report` per swept post**: comments found, replies sent, minutes since the REAL publish time (the successful POST log row — not `posts.scheduled_time`, or a late publish would read as a late sweep). Logged at INFO in-window, WARNING out of it, and shipped to PostHog by `track_golden_hour_report`.
  - The age is computed **in SQL** (`TIMESTAMPDIFF(SECOND, created_at, NOW())`): `logs.created_at` is written in the DB session's zone (`TZ=America/New_York` by default), so subtracting it from a Python UTC clock would have skewed every latency by the offset.
  - Posts older than a day emit nothing — the sweep walks them by design and they'd only add permanent out-of-window noise. An unknown publish time reports `latency_minutes=None` + `within_window=False`; unmeasured is never counted as on-time.
- **Retry**: a sweep that could not run at all (429, session failure) used to lose the whole window silently. It now gets one more attempt, bounded twice — by attempt count AND by the window (a retry that would land past minute 90 is never scheduled), so a sustained 429 decays instead of hammering LinkedIn.

**Second-wave half.** `auto_second_wave_comment` — ONE self-comment 6–8h after publish (fresh random draw per post, no machine-exact offset) that must ADD substance: the caveat, the number, what failed first.

- Runs the **same** #617 quality contract + similarity gate + slop lint as a feed comment (the loop is extracted as `_gated_comment` and shared with `generate_ai_response`, so a new comment surface can't ship with a weaker contract), and **ships nothing** when no draft passes — a filler self-comment costs more reach than silence.
- Its specifics come from the **story bank** (#620), so the added insight is the user's own material; an empty bank gets the explicit no-fabrication fallback.
- Posts through the socialActions API like the #344 seed comment — no browser, so it holds no Selenium lane and is 429-immune — but unlike the seed it is discretionary amplification, so it **stands down under `is_automation_paused()`** (#629 suppression pause / manual pause).
- **No collision with the seed**: the cap is enforced on the COUNT of our own comments on that post URL (`count_user_comments_on_post_url`, `SELF_COMMENT_MAX_PER_POST=2`), so neither task has to know the other ran, however either is re-dispatched.

New env (all defaulted, documented in `.env.example`): `GOLDEN_HOUR_REPLY_SWEEPS`, `GOLDEN_HOUR_RETRY_MINUTES`, `SECOND_WAVE_COMMENT_ENABLED` (kill switch, default on), `SECOND_WAVE_MIN_HOURS`/`SECOND_WAVE_MAX_HOURS`, `SELF_COMMENT_MAX_PER_POST`.

## Acceptance criteria

| From the issue | Where |
|---|---|
| Verify event-mode sweeps fire within 60–90 min | `golden_hour.within_golden_window` / `phase_window_minutes`, WARNING on every in-scope out-of-window sweep |
| Per-post golden-hour report (comments found, replies sent, latency) | `_record_golden_hour_report` + `golden_hour.golden_hour_report` |
| Second wave 6–8h out, substantive, G2 contract, post + story bank grounded, never "thanks for reading" | `auto_second_wave_comment` + `generate_second_wave_comment` (prompt bans it explicitly; the contract rejects filler openers) |
| Seed + second wave must not collide — cap self-comments at 2 | `count_user_comments_on_post_url` vs `self_comment_cap()` |
| Respect the rate_limit breaker and human pacing | Sweeps keep the 429-safe skip + the bounded in-window retry; the second wave checks `is_automation_paused()`; sweep slots keep their `PACE_RESPONSIVE` jitter and the second wave's 6–8h draw IS its pacing; the self-comment is recorded as an action but stays outside the outbound envelope, like a reply on our own post (#626) |
| Unit tests: sweep timing decision, second-wave scheduling, self-comment cap | `tests/unit/utilities/test_golden_hour.py`, `tests/unit/app/test_second_wave_comment.py`, `tests/unit/app/test_reply_sweep.py` |
| PostHog event per golden-hour report | `track_golden_hour_report` → `golden_hour_report` |

## Testing notes

- `poetry run pytest tests/unit -q` → **6816 passed** locally.
- New: `test_golden_hour.py` (35), `test_second_wave_comment.py` (19), `test_db_golden_hour.py` (7), plus new report/retry cases in `test_reply_sweep.py` and a `track_golden_hour_report` case in `test_observability.py`. `golden_hour.py` sits at 97% line coverage.
- `_reply_to_comments_on_open_post` now returns a `{status, summary, comments_found, replies_sent}` dict instead of a sentence (the report needs the counts the summary only ever rendered); its callers and existing tests are updated.
- The three `post_to_linkedin` test files now patch the new second-wave dispatch — without it those tests reached for a real Celery broker.
- No migration. No new API endpoint, so no integration test.

Closes #622


gitchrisqueen:

- **Second wave's 6–8h `countdown` exceeded the broker's visibility timeout (would have posted several self-comments).** `celeryconfig.py` sets `task_acks_late = True` and `visibility_timeout = CELERY_LONGEST_TASK_SECONDS + 15m` (**4500s ≈ 75 min** by default) — and its own comment names this trap. A countdown task is unacked for its entire wait, so one 8-hour message would be redelivered to another worker roughly every 75 minutes and ~6 copies would fire at the same ETA. `count_user_comments_on_post_url` can't hold that line: the copies race the same cap read, so the guard the PR relies on to keep seed + second wave at 2 fails exactly when it's needed. Every other countdown in the repo is ≤ 60 min; this was the first to cross it. **Fixed:** the wait is served in hops sized off `CELERY_VISIBILITY_TIMEOUT` (`second_wave_hop_seconds`, 60% of it), and the due offset comes from `second_wave_due_minutes(user_id, post_id)` — seeded on (user, post) so every re-arm and every redelivery recomputes the SAME target instead of walking forward. `post_to_linkedin` now dispatches only the first hop.
- **A pause during the wait silently cancelled the second wave.** With the hop in place the pause check would have run on every hop; it is now checked only at the DUE time, so a suppression/manual pause that lifts in the meantime no longer costs the post its amplification.
- **A sweep that could NOT run emitted no `golden_hour_report` at all.** The rate-limited / session-failed branches returned before any report, so in PostHog "429" and "nobody commented" still looked identical — the exact ambiguity the issue asks to remove, and one the `status` field was clearly meant to carry. **Fixed:** `_retry_golden_hour_sweep` now records a report with `status=rate_limited` / `session_failed` against the freshest post, and reuses that report's latency for the retry decision (so the post age is still read once).
- **Routine revisits polluted the on-time rate and logged a WARNING forever.** Each sweep walks the last 2 days of posts, so with a 24h report horizon yesterday's post produced an out-of-window event + a WARNING on *every* slot of *every* publish — permanently, and in the denominator of the one question this module exists to answer. **Fixed:** `report_horizon_minutes(phase)` = 2× the phase's own window (180 min for a reply sweep, ~18h for the second wave, hard-capped at 24h), so a genuinely late sweep is still graded and a revisit isn't.
- **Measurement could abort the thing it measures.** The PostHog call was guarded but `get_post_age_minutes` wasn't — a connection failure there would have killed the sweep mid-loop after replying to some posts. Now it degrades to unknown latency (which is already "never on-time"). Also: second-wave reports log `action_type="comment"`, not `"reply"`.

Verified but NOT changed: `count_user_comments_on_post_url` correctly excludes REPLY rows; the `_gated_comment` extraction preserves `generate_ai_response`'s behaviour exactly; the second wave is correctly kept off the Selenium lanes; latency in SQL (not a Python UTC clock) is right given `TZ`-local `logs.created_at`.

One interaction worth knowing about, left alone deliberately: `consolidate_duplicate_comments_for_user` treats *any* post with >1 of our own comments as a duplicate and deletes down to one — after this PR that is every post with a seed + a second wave. It's admin-triggered and `dry_run=True` by default, and re-teaching it the cap belongs with that task, not here.

`poetry run pytest tests/unit -q` → **6832 passed** (+16 new cases: hop bounds/stability, pause-during-wait, unknown-age fallthrough, unrunnable-sweep report, horizon scoping).

gitchrisqueen:

Issue #618, opened 2026-07-26
G3 — Content mix governor (70/20/10) + artifact CTAs, never 'book a call'


gitchrisqueen:
## Context
The July 2026 audit found nearly every post sells the diagnostic conversation; salesy content is penalized up to −70% reach in 2026 and business-goal-first posting is the owner's own top complaint. Winning creators run 70–95% audience-value content (Hills 70/20/10: viral awareness / authority education with NO selling / no-pressure case studies), and promo CTAs are always an ARTIFACT (guide, template, checklist, newsletter) — never a meeting ask.

## Scope
- Content-plan governor in run_content_plan.py + content_alignment.py promo policy: classify each planned post (value / authority / promo), enforce ≤1 promo post per 7–10 posts, and require promo posts be case-study shaped ('no pressure' framing).
- CTA policy: replace 'book a call/DM me to discuss' CTAs with artifact CTAs routed to existing lead-magnet/newsletter assets; ban meeting-ask CTAs in generation prompts and lint for them at review.
- Track mix compliance per plan in DB (reuse posts columns/plan metadata) and expose the ratio in the analytics dashboard (#395 surface).

## Acceptance
- Unit tests: a 30-day plan contains ≤10% promo posts; meeting-ask CTA text fails review; artifact CTA passes.
- Governor respects existing approval workflow; no schema change unless a plan-metadata column is genuinely needed (migration if so).

Issue #392, opened 2026-07-23
C3 — Link-in-first-comment mechanic


gitchrisqueen:
## Why
External links in the body carry a ~60–68% reach penalty. Best practice: put the link in the first comment.

## Scope
- [ ] When a post carries an external/blog/newsletter link, strip it from the body and auto-seed it into the first comment (reuse `auto_seed_comment_on_post`, `run_automation.py:1342`).
- [ ] Preference to opt in/out.

## Files
`app/run_automation.py` (~L1342), `app/run_content_plan.py`, `utilities/ai/content_alignment.py` (CTA/link handling).

## Acceptance
Integration test: a post with a link publishes link-free body + a seeded first comment containing the link.

---
_Part of the LEM 2026 Growth Roadmap (gap analysis vs. 2026 LinkedIn algorithm authorities). Follows repo norms: feature branch `feature/claude-<task>`, ≥80% patch coverage, migrations advance from V56, CI gates green before merge._

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #685, opened 2026-07-26
feat(engagement): close the owned-asset CTA loop (closes #624)

### LINKED ISSUE #624: G9 — Owned-asset CTA loop: route artifact CTAs to newsletter/lead-magnet delivery
## Context
Every studied creator converts LinkedIn attention into an owned email asset (newsletters bypass the feed algorithm; 40–60% open rates). LEM already has newsletter_editions/settings + lead_magnet_sent/settings; G3 mandates artifact CTAs. This issue closes the loop so a CTA actually delivers something.

## Scope
- Map each artifact CTA type to a deliverable: newsletter subscribe link, existing lead magnet, or a generated resource (e.g., the build-receipt's checklist as PDF via existing carousel/PDF tooling).
- Comment-gated delivery (‘comment X and I'll send it') is a common mechanic but DM-delivery at volume is a spam surface — implement ONLY the approval-gated path through existing scheduled_dms/nurture flow, capped by max_dms_per_day, with per-thread one-open-draft rule intact.
- link_in_first_comment already exists — verify the edit-in-later pattern (post without body link, add link in first comment) is used for artifact links, given the −19% to −60% in-body link penalty.
- Track delivery → newsletter-subscriber attribution where possible (newsletter_subscriber_stats).

## Acceptance
- Unit tests for CTA→asset mapping and gated delivery path; no new unapproved outbound DM path.

gitchrisqueen:
## What & why

G3 (#618) made every promo CTA an **artifact** ask — never a meeting ask. Nothing, though, guaranteed the ask actually delivered anything: the newsletter CTA named a newsletter with no link in it, and the lead-magnet CTA's payload went out as a **direct, unapproved DM** to every commenter who typed the keyword. This closes that loop.

### 1. CTA → asset map (`content_alignment.resolve_artifact_delivery`)
The ONE place a CTA resolves to the asset behind it. It names the **channel**, because the two assets arrive completely differently:

| kind | channel | delivers |
|---|---|---|
| `lead_magnet` | `dm` | the comment-keyword mechanic → an approval-gated DM |
| `newsletter` | `link` | the subscribe URL the reader clicks |
| `none` | – | nothing; we drop the ask rather than invent an asset |

`deliverable` is the honest half: an enabled newsletter with **no `newsletter_url` saved** can still be *named* in a CTA, but nothing arrives from it, so the delivery path never pretends otherwise.

**Deliberately not implemented:** a *generated* resource (the build-receipt checklist as a PDF). The scope names it as an example, but LEM has no public asset host to link a generated file from — the assets volume is container-local and the only static mount is the SPA bundle. Wiring one is a hosting decision, not a code change, so `resolve_artifact_delivery` has no `generated` kind rather than a kind that can't deliver.

### 2. The newsletter link rides the first comment (scope: *verify the edit-in-later pattern*)
`artifact_cta_line()` now includes the subscribe URL — the URL **is** the deliverable. That's safe in the body because `split_link_for_first_comment` (#392) runs at the single choke point every post passes through (`post_to_linkedin`) and moves carried links into the first comment, off the body where a link costs −19% to −60% reach. Verified by test, end to end from the CTA line through the split.

### 3. Comment-gated delivery is approval-gated (`_queue_artifact_delivery`)
The keyword listener no longer calls `send_private_dm.apply_async`. It drafts a `pending` `scheduled_dms` row (`source='artifact'`) into the operator's existing queue, beside the #485 nurture drafts:

- **one open draft per thread across BOTH mechanics** — a commenter mid-nurture doesn't also get an artifact draft stacked on the same thread;
- **capped on `max_dms_per_day`** at drafting time, and re-checked by `send_scheduled_dm` at send time;
- `record_lead_magnet_sent` fires on **queue**, not on send — its job is to stop the next sweep re-drafting the same resource for the same person;
- best-effort: a delivery that can't be drafted never breaks the reply sweep it rides on.

No auto-approve flag: the issue asks for **only** the approval-gated path, so there is no env switch back to a direct send.

### 4. Attribution (`count_artifact_cta_deliveries`)
`GET /user/newsletter-subscribers` now returns `attribution` alongside the subscriber history — lead-magnet DMs queued, and posts that carried the subscribe link into their first comment, over the same window. `newsletter_links` is `null` (not `0`) when no URL is configured: nothing to carry is a different fact from carried nothing. Rendered on the Account page's Newsletter card.

## Testing

`poetry run pytest tests/unit -q` → **6865 passed**. SPA `tsc --noEmit` clean.

New/updated:
- `tests/unit/utilities/ai/test_artifact_cta_delivery.py` — the CTA→asset map (priority, half-configured lead magnet, URL-less newsletter, no assets) and the newsletter link riding the first-comment split.
- `tests/unit/app/test_artifact_cta_gated_delivery.py` — the gated path: pending draft + source, keyword matching, already-delivered skip, one-open-draft (own **and** nurture), the daily cap incl. a zero cap, empty render, failed insert not marking the person delivered, DB error non-fatal.
- `tests/unit/app/test_reply_sweep.py` — the sweep now asserts **no** outbound send and a queued draft instead.
- `tests/unit/utilities/test_db_artifact_deliveries.py` — the delivery counts, the blank-URL guard, the DB-error shape.
- `tests/unit/api/test_newsletter_api.py` — the attribution passthrough.

No migration: `scheduled_dms.source` is already `VARCHAR(32) NULL`.

Closes #624


gitchrisqueen:

- **Attribution read 0 forever for the mainline case** (`count_artifact_cta_deliveries`, `db.py`): the link side matched only `posts.first_comment_link`, but `newsletter_url` is a **linkedin.com** article URL — `mark_newsletter_published` writes it from the publish flow's `driver.current_url` — and `is_external_link()` treats `linkedin.com`/`lnkd.in` as internal, so #392's split deliberately leaves it in the BODY (the −19/−60% penalty is off-platform only) and `first_comment_link` is never populated with it. `newsletter_links` would have reported 0 for every LinkedIn newsletter, silently failing the issue's "track delivery → subscriber attribution" criterion. The existing test dodged it by using the host `li` rather than `linkedin.com`. Now matches `content` **OR** `first_comment_link` (one row per post either way), with a regression test that runs a real `linkedin.com/newsletters/...` URL through `artifact_cta_line` → `split_link_for_first_comment` → the count.
- **Unescaped LIKE wildcards in the attribution query** (`db.py`): the pattern interpolated the subscribe URL raw, so percent-encoding (`%20`) inside a newsletter URL became a `%` wildcard matching any text — over-counting the number it feeds. Added `_like_literal()` + `ESCAPE '!'`, with a test.
- **One-open-draft-per-thread was one-directional** (`_nurture_after_reply`, `run_automation.py`): `_queue_artifact_delivery` checks for an open `nurture` draft, but nurture only checked its own `source`. A commenter with a queued lead-magnet draft who then replied to a DM got a second pending message stacked on the same thread — exactly the spam shape the PR says is prevented and the issue asks to keep "intact". Nurture now checks the `artifact` source too, with a test.

Also corrected the `artifact_cta_line` comment, the `SCHEDULED_DM_SOURCE_ARTIFACT` comment and the CLAUDE.md paragraph, all of which asserted the split always moves the subscribe URL into the first comment.

Verified as correct, no change needed: no unapproved outbound DM path remains (`send_private_dm.apply_async` is gone from the keyword listener; queued rows are `pending` and `get_due_scheduled_dms` only picks `approved`); artifact drafts are visible in the review queue (`get_scheduled_dms` does not filter on `source`); `send_scheduled_dm` re-checks `max_dms_per_day` at send; the cap falls back to `_ENGAGEMENT_DEFAULTS` (20), not 0, when the user never saved prefs; `record_lead_magnet_sent` is correctly skipped on a failed insert.

`poetry run pytest tests/unit -q` → **6868 passed**.

gitchrisqueen:

Issue #625, opened 2026-07-26
D1 — Deterministic slop-linter v2 across posts, comments, DMs, newsletter


gitchrisqueen:
## Context
LinkedIn's May 2026 crackdown names specific patterns it suppresses: the 'it's not X, it's Y' contrastive frame, engagement bait, generic thought-leadership. Reader-side tells are well catalogued: blacklist lexicon (delve, game-changer, tapestry, unleash, elevate, seamless, robust, transformative…), 'Here's the kicker'-class transitions, em-dash density, rule-of-three repetition, emoji-bullet listicles, rhetorical-question hooks + 'Agree?' closers, uniform sentence lengths ('perfect rectangle' paragraphs). #416 shipped an LLM humanization pass; this adds a cheap, DETERMINISTIC, explainable lint layer that catches what sampling misses — and runs on ALL surfaces (posts, comments, DMs, newsletter), not just posts.

## Scope
- New module in utilities/ai/ (slop_lint.py): pure functions, regex/string/statistics only — no LLM. Checks: banned-lexicon list (config-extendable), contrastive-negation frame, ta-da transitions, em-dash per-sentence density, rule-of-three repetition, emoji-bullet runs, bait closers, burstiness (sentence-length variance floor), all returning structured violation reasons.
- Wire into the existing review path in content_framework.py (posts/newsletter) AND the comment/DM generation paths; violations → regenerate with the violations injected as constraints (bounded retries) → then block with reasons (surface in review UI like #421).
- Threshold config via env; per-check severities (hard-fail vs warn).
- Explicit policy note in code: this exists to make content genuinely better, not to evade AI detection (#416 policy).

## Acceptance
- Table-driven unit tests per check with real-world positive/negative samples (including the production comment samples from the audit); ≥80% patch coverage.
- Lint runs in <50ms per draft (no network).

Issue #416, opened 2026-07-23
A5 — Humanization / anti-AI-tell rewrite pass for all AI-written text


gitchrisqueen:
## Why
Apply a **humanization / anti-AI-tell rewrite pass** to **every** piece of AI-written text LEM produces — posts, newsletters, DMs, feed comments, seed comments — as a final step **before** it enters human review and **before** it can ever be published. Goal: strip the generic-AI markers that LinkedIn's 2026 360Brew model demotes, and make copy read like the user actually wrote it.

Source technique: the "anti-ai" Claude skill from https://ruben.substack.com/p/how-to-bypass-ai-detectors — style rules that remove AI tells (≤1 em-dash, kill clichés like "delve/leverage/tapestry/paradigm/ecosystem", vary sentence length, add contractions, parenthetical asides, named entities, occasional one-sentence paragraphs). **@gitchrisqueen will attach the password-protected skill `.zip` to this issue** (drag-drop into a comment in the GitHub web UI; share the password out-of-band, NOT in the issue).

## Scope
- [ ] Add a `humanize_text()` pass in the shared content core (`utilities/ai/content_alignment.py`), driven by the skill's rule set, exposed to all generators.
- [ ] Wire it as the **final transform before status is set / draft persisted** in every path:
  - Posts — `app/run_content_plan.py::create_text_post` (before `_review_generated_post` / authenticity gate)
  - Newsletters — `utilities/ai/ai_helper.py::generate_newsletter_edition`
  - DMs — `build_dm_from_template` / `get_ai_message_refinement`
  - Comments / seed comments — the comment generators in `run_automation.py`
- [ ] Config toggle `HUMANIZE_ENABLED` (default ON) + per-type override, mirroring existing `*_ENABLED` flags.
- [ ] Runs **before** human review (APPROVED/PENDING) so nothing public skips it.

## Ordering vs existing M7 work
Pipeline: generate → A2 personal-expertise injection → **this humanization pass** → A1 authenticity gate ([[#382]]) → human review. Humanize first so the A1 gate scores the final text.

## Decisions to confirm (please weigh in)
- [ ] **Disclosure policy:** this conflicts with **A4 (AI-assistance disclosure, #385)** — evading detection vs. disclosing AI are opposite stances. Pick one: humanize-and-don't-disclose, or disclose-and-skip-humanize. (Both can't be true.)
[...]

Issue #382, opened 2026-07-23
A1 — Authenticity scoring gate (360Brew defense)


gitchrisqueen:
## Why
LinkedIn 2026 (360Brew + Authenticity Update) demotes generic AI content — early tests flagged generic posts ~94% of the time; organic reach down ~50% YoY. LEM is AI-content-heavy with **no authenticity safeguard**. This is the single biggest threat to the product.

## Scope
- [ ] Add an authenticity-scoring step in `utilities/ai/content_alignment.py`, run **before** the existing similarity gate `_review_generated_post` (`app/run_content_plan.py:616`).
- [ ] LLM-judge (reuse `lem-medium`) scores generic-AI risk + profile-topic consistency; returns 0–100 + reasons.
- [ ] Gate below threshold: APPROVED → PENDING (mirror the deterministic similarity gate pattern).
- [ ] Persist the score — give the **dead `posts.post_quality` column** (V30, currently no reader) a reader, or add `authenticity_score` (migration from V56).

## Files
`utilities/ai/content_alignment.py`, `app/run_content_plan.py` (~L616), `utilities/db.py`, migration.

## Acceptance
- Golden-set of known-generic vs known-authentic drafts: gate flags generic, passes authentic, **no false-demotion** of good AI-assisted-but-personal posts. Depends on research rubric (see 360Brew rubric spike).

---
_Part of the LEM 2026 Growth Roadmap (gap analysis vs. 2026 LinkedIn algorithm authorities). Follows repo norms: feature branch `feature/claude-<task>`, ≥80% patch coverage, migrations advance from V56, CI gates green before merge._

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #687, opened 2026-07-26
feat(observability): slop-score & engagement-rate telemetry with regression alerts (closes #630)

### LINKED ISSUE #630: D6 — Slop-score & engagement-rate telemetry in PostHog with regression alerts
## Context
Quality needs a trend line, not a one-time fix. Benchmarks: B2B ER-by-impressions ≈3.6% (overall ≈5.2%; documents ≈6.6%, text ≈2%); 1K–5K-follower accounts average ≈4.2%. User 1 is at ≈0.5%. Prompt/model changes (weekly model-retirement cron swaps models!) can silently regress content quality.

## Scope
- Nightly job: for content shipped that day (posts, comments, newsletter), compute and emit to PostHog: D1 slop-lint score (violation count/severity), self-similarity vs recent history (lem-embedding), authenticity-gate score (#382), hook length vs 140-char budget, and per-post ER-per-impression once stats exist.
- Weekly rollup event + alert conditions: slop-score regression after a model/prompt change, ER trending below configurable floor, similarity creeping up. Alerts via log_error→PostHog (existing pipeline) and a dashboard panel (#395 surface).
- Optional external AI-detector API scoring behind env flag (default OFF, cost-capped, sampled) as a REGRESSION SIGNAL only — never an evasion target, per the #416 policy; absence of the key must be a silent no-op.

## Acceptance
- Unit tests for score computation and alert conditions with mocked PostHog; no new hard dependency on external detector services.

gitchrisqueen:
## What & why

Every quality gate LEM already has is a **one-time verdict** — the slop lint (#625) blocks a draft, the comment contract (#617) throws one away, `score_authenticity` (#382) grades a post once. None of them can answer *is the writing getting worse?*, and it silently can: the weekly model-retirement cron swaps models underneath unchanged prompts. B2B engagement-by-impression benchmarks run ~3.6%; user 1 sits near 0.5%. A regression that costs half a percent of reach is invisible in any single draft and obvious in a week of them.

This adds the trend line.

### Nightly — `auto_nightly_content_quality` (02:40 UTC)
Scores everything **shipped** across all three writing surfaces (posts, feed comments, newsletter editions) into the new `content_quality_scores` table plus one `content_quality` PostHog event each:

| Metric | Source |
|---|---|
| Weighted slop score (HARD violations count triple) + which checks fired | `slop_lint.lint_report` (#625) |
| Self-similarity vs that surface's own recent history | one batched `lem-embedding` call, token-overlap fallback |
| Authenticity score | #382's **stored** `posts.authenticity_score` — no fresh judge call |
| Hook length vs the 140-char mobile budget | `content_framework.hook_report` |
| Engagement per impression | latest `post_stats`, once captured |

02:40 is after the 23:00 stats scrape, so yesterday's posts already have impressions. The window is **2 days, not 1**: a missed night self-heals, and a post scored the night it shipped (no stats yet) is re-scored once its impressions land. The write is an upsert, so the overlap costs one extra reading, never a double count.

### Weekly — `auto_weekly_content_quality` (Mon 09:45 UTC)
Reads two periods back out, emits `content_quality_rollup` with both summaries and the deltas, and raises three alert conditions through `log_error` (the existing PostHog pipeline, so a regression becomes a grouped `$exception` issue with no second alerting path):

- **`slop_regression`** — mean slop score rose ≥ `CONTENT_QUALITY_SLOP_REGRESSION_DELTA` (1.0) week-over-week
- **`engagement_floor`** — ER/impression under `CONTENT_QUALITY_ENGAGEMENT_FLOOR` (2%)
- **`similarity_creep`** — mean self-similarity rose ≥ `CONTENT_QUALITY_SIMILARITY_DELTA` (0.05)

**Nothing is ever paused.** Quality drift is a "go look at the prompts" signal; account safety stays #629's job.

### Surfaces
`GET /user/engagement-analytics` gains a `content_quality` block (reading the rollup's own two periods, not the 90-day analytics window — a 90-day "current" period has nothing to compare against), rendered as a Dashboard panel next to comment outcomes (the #395 surface).

### Optional external AI-detector (`AI_DETECTOR_*`)
OFF by default and a **regression signal only** per the #416 policy — it never rewrites, holds or steers content, and is not an evasion target. Gated on BOTH a flag and an API key (no key ⇒ silent no-op), sampled on a stable per-item draw so a retried run never re-bills, capped per user per run, and any failure (timeout, unparseable body, no `requests`) is skipped rather than allowed to break the pass. **No new dependency** — `requests` was already one.

## Decisions worth reviewing

- **Unscored is never zero.** No impressions yet, lint disabled for a surface, or no history to compare against are each recorded as NULL/None and excluded from their **own** denominator — so every dimension carries its own sample size. A 0 would invent a collapse (or, for slop, invent a clean week).
- **ER is impression-weighted** (total engagement / total impressions), not a mean of per-post rates: a post seen by 50 people must not outvote one seen by 5,000.
- **A regression is measured against the account's own prior period**, never an absolute target. The ER floor is the single exception and sits deliberately *below* the ~3.6% benchmark, so it catches a collapse instead of firing forever on a below-average account. Every regression alert needs `CONTENT_QUALITY_MIN_SAMPLE` (5) pieces on **both** sides.
- **A similarity-creep alert never compares an embedding period against a lexical one.** Cosine and token-overlap ceilings differ widely (0.82 vs 0.55), so a week the embedding endpoint was down would otherwise read as a large move.
- **Self-similarity is graded within a surface.** A post compared against the user's comments looks unique no matter how templated it is; an item's own text is excluded from its history, since a shipped comment is already in the log it's being compared against.
- **No content bodies leave the app.** The events and the table carry the *names* of the slop checks that fired, never the lint's reason strings (which quote the draft) — same posture as `turn_off_message_logging`.

## Testing notes

- `tests/unit/utilities/test_content_quality.py` — 93 tests, **100% coverage of the new module**: per-item scoring, batched similarity (self-match exclusion, lexical fallback, one call per batch, history cap), aggregation denominators, impression weighting, the period split, all three alert conditions incl. the sample minimums and the cross-measure guard, and the detector's sampling + silent no-op paths. The embedding call is mocked; nothing touches the network.
- `tests/unit/app/test_content_quality_telemetry.py` — 26 tests for the two beat tasks (what is scored/persisted/emitted, per-surface history, the reported item cap, the detector budget), the two PostHog events, and the beat entries. PostHog is mocked.
- `tests/unit/utilities/test_db_content_quality.py` — 17 tests for the three DB helpers, incl. the LEFT JOIN that keeps an unscraped post and NULL-vs-0 on write.
- `tests/unit/api/test_api_coverage_misc.py` — 2 new tests for the endpoint block.
- Full unit suite green: **6980 passed**. `npx tsc -b` clean for the SPA change (local Node is 18, so the full Vite build runs in CI).
- Migration `V20260726230423__add_content_quality_scores.sql` — additive `CREATE TABLE IF NOT EXISTS`, timestamp version, unique against every existing migration.

Closes #630


gitchrisqueen:

- **Nightly embedding spend was billed to `system`, not the user** (`run_scheduler.py`): `similarity_reports` is the only LLM spend in `auto_nightly_content_quality`, and the task loops over users instead of taking a `user_id` kwarg — so `current_llm_attribution()` had nobody to bill and every `lem-embedding` call landed on the `system` sentinel, invisible to the per-user cost ledger. Wrapped the similarity pass in `llm_attribution(user_id=..., feature=FEATURE_CONTENT)`, matching every other per-user LLM loop in this module; added a test asserting the scope is set and does not leak past the user.
- **The engagement-floor alert could never fire at the default posting cadence** (`content_quality.py`): it gated on `CONTENT_QUALITY_MIN_SAMPLE` (5), which counts every scored PIECE — but only posts carry impressions and `DEFAULT_POSTS_PER_WEEK` is 3, so the issue's headline alert condition was unreachable for every account on the default plan, and a dead alert reads as "engagement is fine". Added `min_engagement_sample()` / `CONTENT_QUALITY_MIN_ER_SAMPLE`, defaulting to `min(3, CONTENT_QUALITY_MIN_SAMPLE)` so it tracks the posting cadence and can never be the stricter of the two. Reported in the rollup `config` block, documented in `.env.example` + CLAUDE.md.
- **`similarity_avg` mixed cosine and token-overlap scores inside one period** (`content_quality.py`): the cross-period guard was there, but each surface embeds in its OWN batch, so one failed `lem-embedding` call leaves a lexical minority inside an otherwise-cosine period — moving the mean by the gap between two scales whose ceilings differ 0.82 vs 0.55 while the dominant *label* stays `embedding`, which the cross-period guard then waves through. `similarity_avg`/`_sample`/`_max` are now taken over the dominant measure only, so the module's own "never average two scales" rule holds within a period as well as across two.

Verified and found sound: the three-surface DB reader (LEFT JOIN keeps an unscraped post; `MAX(id)` picks the latest `post_stats` row), NULL-vs-0 on every dimension, impression-weighted ER, the upsert + 2-day window self-heal, the batch-embedding index arithmetic and self-match exclusion by normalized text, `_day()`'s datetime-before-date ordering, the period split boundaries, the detector's stable sampling / budget / silent no-op, the `content_quality` event carrying `slop_checks` but never `slop_reasons` (which quote the draft), all three MENUS surfaces resolving in `hook_report`, and the timestamp migration version being unique.

8 new unit tests; `content_quality.py` stays at 100% coverage. Full unit suite: **6988 passed** locally.

gitchrisqueen:

PR #688, opened 2026-07-26
feat(flags): runtime feature flags with local evaluation + SPA bootstrap (closes #651)

### LINKED ISSUE #651: PH6 — Feature flags: runtime toggles with local evaluation + SPA bootstrapping
## Context
LEM's operational toggles are env vars requiring a deploy/restart to change (COMMENT_RESEARCH_ENABLED, TUTORIAL_VIDEOS_ENABLED, feed fallback semantics, cost-routing enables...). PostHog flags give runtime control, % rollouts, and per-user targeting; Python LOCAL EVALUATION (personal API key, in-process) is the right pattern for Celery workers — no network hop per check; posthog-js bootstrapping kills SPA flag-flicker.

## Scope
- utilities/flags.py: thin wrapper — flag lookup with local evaluation, cached poller, and FAIL-OPEN TO THE ENV VAR (a PostHog outage or missing key must behave exactly like today). Every migrated toggle keeps its env var as default/fallback.
- Migrate a first cohort of NON-SAFETY toggles: COMMENT_RESEARCH_ENABLED, TUTORIAL_VIDEOS_ENABLED, feed_fallback default, one or two experiment enables. Explicitly DO NOT migrate the 429 breaker, automation pause, or caps — safety controls stay in Redis/env.
- SPA: react hooks (posthog-js/react) + server-side bootstrap of flags into index/config payload.
- Document the flag registry (name, default, owner) in docs.

## Acceptance
- Unit tests: env fallback when client unavailable; flag flip visible in worker within poll interval; no per-call network requests (local eval).

gitchrisqueen:
## What & why

LEM's operational toggles were env vars — changing one meant editing `/opt/lem/.env` and restarting the stack. This adds **`src/cqc_lem/utilities/flags.py`**, the ONE place a runtime toggle is read, backed by PostHog feature flags with in-process **local evaluation**.

The contract is **fail open to the env var**. Every path that isn't a clear PostHog answer — no personal API key, `POSTHOG_FLAGS_ENABLED=false`, definitions unloaded, flag undefined in PostHog, evaluation inconclusive, SDK raises — returns the flag's own env var. A deployment that never creates a single PostHog flag behaves *exactly* as it did before this landed. Each registered flag keeps its env var as both default and fallback.

### Local evaluation (and its constraint)

Every lookup passes `only_evaluate_locally=True` + `send_feature_flag_events=False`: definitions are polled into the process (`POSTHOG_FLAG_POLL_SECONDS`, default 30) and evaluated in memory, so **a feed loop checking a flag per post makes zero network requests**, and a flip reaches a long-running Celery worker within one poll interval without a restart. The single fetch happens at a process's first flag check, and a failed fetch retries no more than once per poll interval — an outage can't turn a feed loop into a fetch loop.

The price is a real registry constraint, documented and called out in the module docstring: **registered flags must use rollout-percentage / distinct-ID conditions only.** A person-property release condition can't be decided locally and would silently fall back to env everywhere.

### Migrated cohort

Each is read at the **call site** — an import-time constant can't be flipped, which is the classic way to ship a flag that does nothing.

| Flag | Env fallback | Scope |
|---|---|---|
| `comment-research-enabled` | `COMMENT_RESEARCH_ENABLED` | per user (`user_id` plumbed through `research_topic`) |
| `tutorial-videos-enabled` | `TUTORIAL_VIDEOS_ENABLED` | system — gates the producer AND the SPA section |
| `feed-fallback-when-empty-default` | `FEED_FALLBACK_WHEN_EMPTY_DEFAULT` | fleet default only; a user's saved row always wins |
| `cost-routing-enabled` | `COST_ROUTING_ENABLED` | system |

`COST_ROUTING_ENABLED` / `TUTORIAL_VIDEOS_ENABLED` are no longer frozen into `env_constants` — two readers of one var, one of them unflippable, is worse than none.

### Explicitly NOT migrated

Safety controls stay in Redis/env, where they're one authority a PostHog outage or an errant %-ramp can't touch: the 429 breaker, `hold_commenting`, `pause_automation`/the suppression tripwire, and every per-day cap. `COST_AWARE_ROUTING_ENABLED` (the proxy half of cost routing) stays env-only because `routing_policy.py` is mounted into the LiteLLM container and must remain stdlib-only. A unit test asserts no safety-shaped key enters the registry.

### SPA

`GET /api/flags` returns `{distinct_id, flags, local_evaluation}` — every registered flag, resolved **server-side**. `hooks/useFeatureFlags.ts` consumes it; `TutorialVideos.tsx` is the first consumer.

One deviation from the issue text worth flagging: the issue suggested `posthog-js/react` hooks. I did **not** use them, for three reasons — (1) it *is* the bootstrap: one server payload means a gated section renders correctly on the first paint instead of flickering, and posthog-js's `bootstrap` option needs the flags *before* `init()`, which the SPA can't have; (2) it's the same evaluation the API and workers just did, so browser/API/worker can never disagree; (3) it works with browser analytics fully off (no `VITE_POSTHOG_KEY`) — flags are a product control, not an analytics feature. `utils/analytics.ts` stays the SPA's ONE PostHog surface per CLAUDE.md; a second client-side posthog reader would reintroduce exactly the split-brain the endpoint removes. Happy to swap if you'd rather have the SDK hooks.

An absent/invalid session resolves flags for the `"system"` identity rather than 401ing — the landing page is logged out and still needs to know what to render.

## Testing

- `tests/unit/utilities/test_flags.py` (42) — registry integrity, env coercion, every fail-open path, variant coercion, local-eval kwargs, definitions-load-once + poll interval, retry cooldown, distinct_id convention, bootstrap payload.
- `tests/unit/utilities/test_flag_migrations.py` (13) — each migrated toggle provably follows its flag, the env var wins when PostHog doesn't answer, `CONTENT_RESEARCH_ENABLED` still overrides the flag, and cost routing resolves at call time rather than import.
- `tests/integration/test_flags_endpoint.py` (6) — the real endpoint through the real resolution: no-PostHog payload, anonymous vs. session-scoped identity, PostHog beating the env var.
- `src/cqc_lem/ui/src/hooks/useFeatureFlags.test.tsx` (6) — fallback while in flight, safe fallback on an unreachable API, session token passthrough.
- `tests/unit/conftest.py` pins the unit lane to env-only flags: `tests/conftest.py` calls `load_dotenv()`, so a dev box with a real `POSTHOG_PERSONAL_API_KEY` would otherwise make unit tests reach the network and depend on whichever rollout is live.

`poetry run pytest tests/unit -q` → **6923 passed**. `tests/integration` → 112 passed, 17 skipped (2 pre-existing failures in `test_link_first_comment.py` need the Redis service container; they fail identically on a clean tree). Vitest/tsc run in CI (`UI Build`) — no vitest in this box's node_modules.

## Deployment note

Local evaluation needs `POSTHOG_PERSONAL_API_KEY` (scope `feature_flag:read`) **in the app containers**. It's already in `.env.example` and the app services use `env_file: .env`, so if it's set on the box this works on deploy; if it isn't, every flag simply reads its env var and nothing changes. No migration, no schema change.

Docs: `docs/feature-flags.md` (registry, fail-open table, how to add a flag, verification commands) + a CLAUDE.md section.

Closes #651

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -109,6 +109,8 @@ POST_RESEARCH_ENABLED=true
 # Feed comments: default FALSE. Comments run at high volume (many/day/user) and are already
 # grounded in the target post — a Perplexity call per comment would multiply API spend for
 # marginal value. Opt in only if you want research-backed comments.
+# ALSO a runtime feature flag (`comment-research-enabled`, docs/feature-flags.md) — this env var is
+# what it falls back to when PostHog has no answer, so it stays the source of truth for the fleet.
 COMMENT_RESEARCH_ENABLED=false
 # Lead-magnet soft-ask cadence: when a user has the "comment KEYWORD → DM" lead magnet enabled,
 # weave the compliant "comment KEYWORD and I'll DM it to you" CTA into roughly 1 in N generated
@@ -521,6 +523,20 @@ POSTHOG_APP_HOST=https://us.posthog.com
 # (issue #650, docs/kpi-dashboards.md) — the automated replacement for the hand-run perf report.
 # Unset falls back to the personal API key owner's own email.
 POSTHOG_REPORT_EMAIL=
+# --- Feature flags (issue #651, docs/feature-flags.md) ---
+# Runtime toggles evaluated LOCALLY in-process (no network call per check) from definitions polled
+# by the SDK. They need the POSTHOG_PERSONAL_API_KEY above (scope: feature_flag:read) in the app
+# containers — without it, and on every other failure path, every flag falls back to its own env
+# var and behaviour is exactly what it was before flags existed.
+# Master kill switch for the whole mechanism: false = env vars only, no polling, no lookups.
+POSTHOG_FLAGS_ENABLED=true
+# How often a worker re-reads flag definitions — i.e. the worst-case delay before a flip is live.
+# Also the cooldown between retries when a definition fetch fails, so an outage can't make every
+# flag check in a feed loop attempt its own fetch.
+POSTHOG_FLAG_POLL_SECONDS=30
+# Fleet default for `feed_fallback_when_empty` for users who have never SAVED engagement
+# preferences. A user's own saved setting always wins over this and over the flag.
+FEED_FALLBACK_WHEN_EMPTY_DEFAULT=true
 # --- PostHog in the SPA (issue #646) ---
 # Read by Vite at BUILD time and baked into the bundle, so they must be set for `npm run build`
 # (docker build-arg / CI env), not in the running container. Leave VITE_POSTHOG_KEY empty and the
@@ -611,6 +627,8 @@ INFRA_FIXED_MONTHLY=0
 # proxy on later would start at whatever cohort the sham ramp had reached.
 # Shipped OFF by owner decision (PR #529): the loop is dormant — while COST_ROUTING_ENABLED is false
 # the weekly task observes nothing and just republishes the parked policy. Flip BOTH to true to start.
+# The app side is ALSO a runtime feature flag (`cost-routing-enabled`, docs/feature-flags.md) that
+# falls back to this var; the proxy side stays env-only (routing_policy.py must remain stdlib-only).
 COST_ROUTING_ENABLED=false
 COST_AWARE_ROUTING_ENABLED=false
 # Where the policy document lives. MUST be the same Redis the app and the LiteLLM container share.
@@ -768,7 +786,9 @@ FEEDBACK_MAX_ISSUES_PER_USER_PER_DAY=3
 
 # --- Marketing video tutorials (issue #505) ---
 # Automated SPA how-to videos: headless capture -> grounded script -> TTS -> ffmpeg MP4 -> YouTube.
-# OFF by default (it renders and uploads public marketing assets).
+# OFF by default (it renders and uploads public marketing assets). ALSO a runtime feature flag
+# (`tutorial-videos-enabled`, docs/feature-flags.md) that gates BOTH the producer and the SPA
+# section; this env var is its fallback.
 TUTORIAL_VIDEOS_ENABLED=False
 # Where the capture browser points. Empty = this deployment's own public URL.
 TUTORIAL_SPA_BASE_URL=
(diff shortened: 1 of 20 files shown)


gitchrisqueen:

- **`GET /api/flags` was inside the `API_ACCESS_TOKENS` bearer gate (silent logout).** The SPA bootstraps flags from the **landing page** (`TutorialVideos` → `useFeatureFlags`), which carries no bearer token, but `/api/flags` was not added to `_PUBLIC_API_PREFIXES`. With the gate armed the query 401s, and `api/client.ts`'s response interceptor treats **any** 401 as a dead session — it clears `lem_session`/`lem_email` and redirects — so a signed-in visitor landing on `/` would be logged out by a marketing section (twice over, given `retry: 1`). `/api/faq` is in the allowlist for exactly this reason and `/api/flags` is the same shape: GET-only, no user data, and its optional `session_token` is self-authenticating (an invalid one resolves the `"system"` identity rather than erroring). Fixed in `2f0dc0b` — added `/api/flags` to `_PUBLIC_API_PREFIXES` with the rationale comment, an integration test that drives the real handler through the armed gate (and asserts the leaf boundary still keeps a hypothetical `/api/flags-admin` gated), and a docs note.

Everything else checked out. Verified against the installed SDK (posthog 7.21.3) rather than the docs: `only_evaluate_locally` / `send_feature_flag_events` are real kwargs; `load_feature_flags()` swallows 401/402/network failures and starts the poller off `client.poll_interval`, so pushing the settings onto an already-built client is both necessary and sufficient, and `_loaded=True` after a silently-failed fetch is correct (the poller self-heals; every lookup fails open to env meanwhile). Also confirmed: no stale importers of the two constants dropped from `env_constants` (`COST_ROUTING_ENABLED`, `TUTORIAL_VIDEOS_ENABLED`); `get_session_user_id` returns `Optional[int]` and never raises, so the endpoint's "invalid session → system identity" docstring holds; `user_id` was already plumbed to `generate_ai_response` at all three call sites, so the per-user cohort actually reaches `research_topic`; `CONTENT_RESEARCH_ENABLED` still beats the flag; `prefs.get("feed_fallback_when_empty", True)` in `run_automation` is never the value that decides, since `get_engagement_preferences` always supplies the key; `QueryClientProvider` wraps `Landing`, so the new hook can't throw there; app services use `env_file: .env`, so `POSTHOG_PERSONAL_API_KEY` does reach the workers.

`tests/unit` 7043 passed; `tests/integration/test_flags_endpoint.py` 7 passed.

gitchrisqueen:

Issue #396, opened 2026-07-23
D2 — Lightweight outcome-tracked A/B harness


gitchrisqueen:
## Why
`generate_variants.py` produces variants for **human preview only** — no outcome tracking, no winner selection.

## Scope
- [ ] Persist which hook/format variant actually shipped for a post.
- [ ] Read back its `post_stats` to pick winners over time; feed into B3/B4 rankings.

## Files
`app/generate_variants.py`, `utilities/db.py`, `utilities/post_stats.py`.

## Acceptance
Unit/integration: a shipped variant is recorded and its performance attributed; winner selection reflects seeded results.

---
_Part of the LEM 2026 Growth Roadmap (gap analysis vs. 2026 LinkedIn algorithm authorities). Follows repo norms: feature branch `feature/claude-<task>`, ≥80% patch coverage, migrations advance from V56, CI gates green before merge._

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #689, opened 2026-07-27
feat(observability): port cost-routing + content A/B experiments to PostHog Experiments (closes #652)

### LINKED ISSUE #652: PH7 — Experiments: port cost-routing + content A/B harness to PostHog Experiments
## Context (depends on PH6 #651)
LEM hand-rolled experimentation twice: the cost/quality down-routing cohort (routing_policy.py / cost_routing.py, docs/cost-performance-margin-plan.md §D.1.1) and the outcome-tracked A/B harness (#396). PostHog Experiments provides multivariate flags + Bayesian AND frequentist stats engines, CUPED variance reduction, holdouts, and 'LLM A/B tests' for prompt variants — billed with flags, free at our volume.

## Scope
- Re-platform the cost-routing experiment: treatment assignment from a PostHog multivariate flag (respecting routing_policy.py's stdlib-only constraint — the flag DECISION gets passed in, e.g. resolved app-side and mounted/handed to the router, not imported); success metrics = engagement-rate + $ai_generation cost from PH2.
- Port the #396 A/B harness adapter so experiment exposure + outcome events flow to PostHog experiment analysis instead of (or alongside) the homegrown stats.
- One pilot LLM prompt experiment: comment quality contract prompt variant (G2) with author-reply-rate (D4) as the metric — honestly document the small-sample caveat at current volume; the harness is for the multi-user future.

## Acceptance
- Cost-routing cohorting reproducible from the flag; exposure events land; experiment readout renders in PostHog; stdlib-only constraint on routing_policy.py verified by existing tests.

gitchrisqueen:
## What & why

LEM hand-rolled experimentation twice: the cost/quality down-routing cohort (`routing_policy.py` / `cost_routing.py`, plan §D.1.1) and the #396 outcome-tracked media A/B harness. Both work. Neither can answer *"is this difference real"* with anything a statistician would sign, and neither renders anywhere a human looks. PostHog Experiments already has the multivariate flag, the exposure model, both stats engines (Bayesian + frequentist), CUPED and the readout — and it's billed with feature flags, so free at our volume.

`utilities/experiments.py` is therefore the **adapter, not a third implementation**. The homegrown loops keep running exactly as they did; PostHog gets the arms, the exposures and the outcome labels.

Full posture: **`docs/experiments.md`**.

### The contract

**An unresolvable experiment is the CONTROL arm** — no PostHog key, definitions not loaded, flag not defined, evaluation inconclusive, `EXPERIMENTS_ENABLED=false`, SDK raises. There is deliberately **no per-experiment env fallback** (unlike #651's flags: a toggle has a configured default worth honouring, an experiment arm does not), and `_raw_variant()` keeps *"PostHog said control"* apart from *"PostHog said nothing"* so `experiment_properties()` never stamps `$feature/x=control` on a metric event from someone who was never enrolled — a fabricated control arm makes a readout look populated.

Assignment reuses **flags.py's ONE local-evaluation bootstrap** rather than starting a second poller (zero network calls per lookup — these run in feed loops). Consequence, documented: every experiment flag must use rollout-percentage / distinct-ID conditions only.

Exposure is `$feature_flag_called` — PostHog's own event name and properties, not ours — emitted explicitly (flags.py suppresses it for hot toggles) and **deduped per (experiment, person, arm) per process**.

### 1. Cost-routing cohorting moved to the flag, without breaking stdlib-only

`routing_policy.py` is mounted into the LiteLLM container, so it can never import `experiments.py`. The DECISION is handed to it instead: the weekly run resolves each active user's arm app-side and writes it into each routing bucket as `arms: {"7": "treatment"}`, inside the policy document Redis already carries to the router.

- `flag_arm()` reads the map (string keys — the document round-trips as JSON); `assign_arm()` prefers it and **falls back to the original SHA-256 hash** for anyone PostHog has no answer for, so an outage moves no traffic.
- `resolve_tier()` now reports `assignment` (`flag`/`hash`) — the difference between *"the experiment is running"* and *"PostHog was unreachable"*.
- **The flag decides WHO, `cohort_pct` decides WHETHER**: a parked/rolled-back bucket can never be started by a flag, and inside a running one the flag wins both ways — including keeping a holdout the ramp would have swept up, which is what keeps the §D.3 quality gate measurable after adoption.
- Arms are applied **after** the weekly evaluation: the window being judged ran under the previous document's cohort.
- Map capped at 500 users, and a truncation is **logged** (the users left out are silently hash-assigned otherwise).

### 2. Pilot LLM prompt experiment (G2)

`comment-contract-prompt`: control vs. `author-question`, which adds one rule — end on a question **only this author could answer**. Metric is author-reply rate (D4) from #628's T+24h sweep, carried on `comment_outcome`.

Scoped deliberately: the six deterministically-graded contract rules are **identical in both arms** (an arm that loosened one would change what "passes the gate" means), and only **fresh feed comments** are enrolled — the seed / second-wave / reply surfaces are never measured by that sweep.

### 3. #396 harness adapter

The combo that shipped IS the arm (`assignment: shipped`, no flag to read). `record_shipped_variant()` is the exposure moment; a single `get_shipped_variant_keys()` query per stats sweep puts each post's variant on its `post_outcome` event. `select_variant_winners` is untouched — PostHog gets the stats engine, the ranking keeps its recency weighting.

### 4. Provisioning

`scripts/posthog_experiments.py` — `--print-specs` / dry-run (default) / `--apply` / `--rollout KEY=PCT`. It **never resets an existing flag's rollout**: PostHog owns the ramp once an experiment runs, and an `--apply` that reverted it would re-cohort a live experiment. Media arms are derived from `DEFAULT_COMBOS` so they can't drift from what the code can ship; flag conditions are property-free at 100% so local evaluation can always decide them.

### Honesty about the readout

At current volume (one active user, ~3 posts/week) none of this reaches significance on anything subtler than a collapse — `MINIMUM_DETECTABLE_EFFECT` is a deliberately coarse 30% and the small-sample caveat is spelled out in the doc. The harness exists so the multi-user future has a measurement plane already emitting. The attribution caveat (don't re-roll a running experiment's variants) is documented too.

### Bug found by the new tests

`build_routing_policy` shadowed its own incoming cohort with the per-bucket arm *split* (`arms = split_observations(...)`), so the resolved cohort never reached the document. Fixed (`arm_rows`).

## Dependency

Builds on **#651 / #688** (`utilities/flags.py`), which merged while this was in flight — rebased onto it, resolving `cost_routing.py`'s docstring and the `enabled: Optional[bool] = None` runtime-flag signature so the new `arms` parameter sits alongside #688's `routing_enabled()` rather than reverting it.

The `flags` import is still **lazy and latched** (documented in `_flag_surface`): experiments hold no import-time dependency on the flag surface, and any failure to load it reads as not-ready, which is the same fail-to-control path as a missing key.

## Acceptance criteria

| Criterion | Where |
|---|---|
| Cost-routing cohorting reproducible from the flag | `routing_policy.flag_arm` + `cost_routing.resolve_cohort`; `test_the_flag_cohort_survives_the_container_boundary` drives a PostHog answer → real policy document → JSON round-trip → the real LiteLLM hook's down-route |
| Exposure events land | `observability.track_experiment_exposure` (`$feature_flag_called` + `$feature_flag_response`), deduped in `experiments.track_exposure` |
| Experiment readout renders in PostHog | `scripts/posthog_experiments.py --apply` creates the multivariate flags + experiment records; metrics point at `post_outcome` / `comment_outcome` / `$ai_generation` |
| stdlib-only constraint verified by tests | `test_complexity_router.py` (unchanged, still green) + a new explicit AST guard, `test_the_decision_core_stays_stdlib_only` |

## Testing notes

- `poetry run pytest tests/unit -q` → **7089 passed** (81 new).
- New/extended: `tests/unit/utilities/test_experiments.py` (30), `tests/unit/test_posthog_experiments.py` (25), routing arms + stdlib guard, cohorting in `test_cost_routing.py`, exposure/metric labelling in `test_observability.py`, prompt-variant wiring in `test_comment_contract.py`, adapter in `test_generate_variants.py`, per-sweep variant lookup in `test_sweep_fixes.py`, `test_variant_ab_harness.py` (integration).
- The two `tests/integration/test_link_first_comment.py` failures locally are pre-existing on `main` (verified by stashing) — they need the service containers.
- Nothing turns on by itself: `EXPERIMENTS_ENABLED` defaults true but with no `POSTHOG_PERSONAL_API_KEY` (and cost routing still shipped OFF) every arm is the control and no flag exists until the provisioner is run.

Closes #652

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -648,8 +648,18 @@ COST_ROUTING_CONFIDENCE_Z=1.96
 # Days a rolled-back bucket is left alone before it may be re-proposed.
 COST_ROUTING_COOLDOWN_DAYS=28
 # Blast radius: share of users a new experiment starts on, and how many buckets may run at once.
+# NOTE (issue #652): once the `cost-routing-arm` PostHog experiment exists, the FLAG decides who is in
+# the treatment and this percentage is only the hash fallback for users PostHog has no answer for.
 COST_ROUTING_INITIAL_COHORT_PCT=0.1
 COST_ROUTING_MAX_EXPERIMENTS=1
+
+# --- PostHog Experiments (issue #652, docs/experiments.md) ---
+# ONE kill switch for every experiment in utilities/experiments.py. Off means every arm reads as its
+# CONTROL — the behaviour that shipped before the experiment existed — and no exposure is emitted.
+# There is no per-experiment env var on purpose: a toggle has a default worth honouring, an
+# experiment arm does not. Assignment needs POSTHOG_PERSONAL_API_KEY (local evaluation, see
+# docs/feature-flags.md); without it every arm is the control too.
+EXPERIMENTS_ENABLED=true
 
 # --- Codecov (coverage reporting) ---
 CODECOV_TOKEN=your_codecov_token
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -129,8 +129,10 @@ response = client.chat.completions.create(model="lem-simple", messages=[...])
 can additionally route a tier ONE step down for the treatment cohort of an active cost/quality
 experiment. `routing_policy.py` is the shared decision core — the app imports it, and docker-compose
 mounts that same file into the LiteLLM container — so it must stay **stdlib-only** (no `cqc_lem.*`
-imports). Off unless BOTH `COST_ROUTING_ENABLED` and `COST_AWARE_ROUTING_ENABLED` are set. See
-`docs/cost-performance-margin-plan.md` §D.1.1.
+imports). Off unless BOTH `COST_ROUTING_ENABLED` and `COST_AWARE_ROUTING_ENABLED` are set. Since
+issue #652 the treatment cohort comes from a PostHog experiment flag resolved app-side and handed to
+the router in the policy document's `arms` map (the hash stays as the fallback) — see
+`docs/experiments.md`. Also `docs/cost-performance-margin-plan.md` §D.1.1.
 
 See `ai_helper.py` for the per-function model assignment.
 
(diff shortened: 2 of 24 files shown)


gitchrisqueen:

- **A PostHog outage re-cohorted the live experiment** (`cost_routing.py`): `resolve_cohort()` returned `{}` for both *"PostHog enrolled nobody"* and *"we could not ask"* (no personal key, definitions unloaded, `assignments()` raised, routing parked), and `apply_arms` cleared the bucket's `arms` map on either. One unreachable weekly run therefore dropped every enrolled user back onto the 10% hash for a week — a **different** split, mid-experiment — which is the arm flip `assign_arm`'s own docstring exists to prevent, and the following run then split that window's observations under the hash and graded posts in the arm they weren't written in. The per-user fallback claim ("a PostHog outage moves no traffic") only held per user, not at the document level. Fixed: `resolve_cohort()` now answers `None` for *couldn't ask* and `{}` only for a real empty answer; `build_routing_policy` carries the previous document's cohort forward (`carried_arms()`) on `None` and still clears on `{}`. Routing being off carries it forward too, so `collect_routing_report`'s "resumes exactly where the loop left off" now includes who was in which arm. Tests: `test_a_posthog_outage_moves_nobody_between_arms`, `test_an_ended_experiment_does_clear_the_cohort`, `test_apply_arms_carries_the_cohort_forward_when_posthog_could_not_be_asked`, `test_carried_arms_unions_only_the_buckets_that_route`, `test_collect_routing_report_does_not_wipe_the_cohort_while_routing_is_off`, `test_resolve_cohort_reports_none_when_it_could_not_ask`.
- **The kill switch leaked the media arm** (`experiments.experiment_properties`): the `extra`/shipped path has no flag to read, so with `EXPERIMENTS_ENABLED=false` `post_outcome` still carried `$feature/post-media-variant` — a switched-off experiment rendering a populated property breakdown, the exact failure `_raw_variant`'s control/None split is there to avoid. Now returns `{}` when disabled. Test: `test_the_kill_switch_also_strips_a_shipped_arm_label`.
- **Orphan exposure on a failed variant write** (`generate_variants.record_shipped_variant`): the exposure fired before `_db_record_shipped_variant`, which returns False rather than raising — so a DB failure enrolled a post whose arm label is read back out of `post_variants` by the stats sweep and therefore can never exist. Exposure now follows a successful write. Test: `test_a_failed_db_write_reports_no_exposure`.

Verified but NOT changed:
- `routing_policy.py` stays stdlib-only (the new AST guard is the right check); the JSON string-key round trip is real and covered end-to-end through the actual LiteLLM hook.
- No import cycle: `observability` → `experiments` is top-level, `experiments` → `flags` and `flags` → `observability` are both lazy.
- The comment-contract treatment arm ("end on a question only this author could answer") does not collide with #625's `bait_closer` HARD check — that only matches `REFLEX_CLOSERS`, which the arm's prompt explicitly rules out — and the six graded rules are byte-identical in both arms.
- All three `generate_ai_response` call sites pass `user_id`, so no exposure lands on the `system` person; the seed and second-wave comments go through `_gated_comment` directly and are correctly unenrolled.
- `EXPERIMENTS_ENABLED` reaches every container (compose uses `env_file`, not an enumerated `environment:` list).

`poetry run pytest tests/unit -q` → **7152 passed** (8 new).

gitchrisqueen:

Issue #502, opened 2026-07-24
feat(feedback): auto-changelog + notify reporters on merged/deployed fixes


gitchrisqueen:
## Why
Closes the loop (docs/launch-and-marketing-plan.md (PR #495) Section B.4): when a fix ships, tell the users who asked and re-measure satisfaction.

## Scope
- `auto_changelog_notify` task: on a merged PR that 'Closes #<issue>', append a human-readable changelog line and notify every user attached to that feedback cluster (email + in-app: 'You asked, we shipped: …').
- Schedule a micro-CSAT ('did this fix it?') to those users; response re-enters the loop and updates GA-gate metrics.

## Files
`app/run_scheduler.py` (+beat entry), `utilities/db.py`, `utilities/linkedin/notification_email.py`, unit tests.

## Acceptance
- Reporters mapped from feedback→issue→cluster and notified once; changelog line generated; ≥80% patch coverage.

Issue #499, opened 2026-07-24
feat(billing): early-adopter extended-trial (60d) auto-grant, gated on review submission


gitchrisqueen:
## Why
Extended-trial program in docs/launch-and-marketing-plan.md (PR #495) Section A.2. Standard trial is 14d (`FREE_TRIAL_DAYS`); early adopters get 60d in exchange for a review.

## Scope
- New `POST /trial/extend` (feature-flagged, cohort-gated) → `extend_trial_for_user` in `utilities/db.py` sets trial_ends_at = trial_started_at + EARLY_ADOPTER_TRIAL_DAYS (default 60).
- **Gate: only grant if the user has a `feedback` row with source='review'**; otherwise return a 'submit a quick review to unlock 60 days' response.
- Atomic cohort slot counter (P0:25 / P1 first 100); fall back to 14d when exhausted.
- Extend `utilities/stripe_util.py:create_checkout_session` to accept `trial_period_days`/`discounts` (coupon) so the extension mirrors into Stripe on conversion.

## risk:product-decision
Auto-modifies billing terms — owner signs off on length (60d), cohort caps, and coupon via Decision Comment.

## Files
`api/main.py`, `utilities/db.py`, `utilities/stripe_util.py`, `utilities/env_constants.py`, migration if needed.

## Acceptance
- Extension only when a review exists; slot counter atomic; integration tests; ≥80% patch coverage.

PR #690, opened 2026-07-27
feat(surveys): NPS/CSAT via PostHog Surveys, wired into the feedback loop (closes #653)

### LINKED ISSUE #653: PH8 — Surveys: NPS/CSAT via PostHog surveys, wired into the feedback→GitHub loop
## Context (depends on PH1 #646)
LEM built homegrown survey prompts (survey_prompts table, track_survey_prompt/response) and an in-app FeedbackWidget. PostHog Surveys (1,500 responses/mo free) gives NPS/CSAT/PMF templates, targeting by events/person properties/flags, wait-between-surveys throttling, and response summarization — surveys are just events, renderable headless if the popover clashes with the SPA design.

## Scope
- NPS survey targeted at users ≥30 days after onboarding completion; CSAT micro-survey after N post-approvals (event-targeted); respect wait-between-surveys.
- Render: start with PostHog's popover; if it clashes, headless render in the existing widget capturing survey sent events.
- Route responses into the existing feedback→GitHub loop (low scores open a feedback report), replacing the homegrown survey_prompts scheduling for these two surveys (keep the table for anything bespoke).
- Keep track_survey_response for continuity or map it to the PostHog events — no double counting.

## Acceptance
- Surveys fire under targeting rules in a test account; responses visible in PostHog; low-score responses produce a feedback issue; homegrown scheduler no longer double-prompts.

gitchrisqueen:
## What & why

PostHog Surveys now own the two general-purpose asks, while the bespoke ones stay in `utilities/surveys.py`. The line between the two owners is **what an answer can DO**.

| Ask | Owner | Why there |
|---|---|---|
| **NPS** — 30 days after activation | PostHog Surveys | Targeting/cadence tunable without a deploy |
| **Post-quality CSAT** — after 5+ approvals | PostHog Surveys | Event-triggered, so it lands in the moment the user just judged a draft |
| **Review** — trial T-3d | `utilities/surveys.py` | It is the #499 extended-trial gate; PostHog cannot unlock anything |
| **Fix CSAT** — "did this fix it?" | `utilities/surveys.py` | Per-issue (`fix_csat_<n>`), scheduled by the shipped-notice queue (#502) |

### Headless, not the popover

Both surveys are created as PostHog **`api`** type and rendered in LEM's own modal (`PostHogSurveyModal.tsx`). A popover answer is a PostHog event and **nothing else** — it would never reach the feedback→auto-work loop, which is the reason LEM asks for scores at all. (The popover also renders bottom-right, exactly where the feedback widget already sits.)

So one answer takes two paths and is **counted once**:

1. The browser emits PostHog's native **`survey sent`** (`$survey_response` / `$survey_questions`, by index *and* by question id), so it shows up in the Surveys product.
2. The same answer POSTs to **`/api/survey/posthog`**, becoming a `feedback` row that the classifier files as a GitHub issue.

`track_survey_response` — the homegrown `survey_response` event — is deliberately **not** emitted on that path; two events per answer would double every response-rate number. The row carries `context_json.origin = "posthog_survey"` so anything counting rows out of MySQL can still tell the streams apart.

| Answer | Feedback status | Effect |
|---|---|---|
| NPS ≤ 6, or CSAT ≤ 2 | `new` | Auto-filing pass opens / `+1`s a feedback issue |
| Any score **with** free text | `new` | The text is the actionable part |
| Happy score, no comment | `resolved` | Nothing to classify; must not burn an LLM call per promoter |

Because we render, posthog-js never sees the ask — `markSurveySeen()` is what advances its seen/wait-period bookkeeping. Drop it and the 30-day `seenSurveyWaitPeriodInDays` throttle silently stops working.

### Targeting

Person properties set at `$identify` off `GET /auth/session`:

- **NPS** — `onboarding_completed_at is_date_before -30d`. That is `onboarding_state.activated_at`, the activation "aha", **not** signup: a user who signed up and stalled has no opinion worth a score, and their detractor answer would be about onboarding.
- **CSAT** — activation event `post_approved` **and** person `posts_approved >= 5`. The count spans `approved`/`scheduled`/`posted`, because an approved post moves on as automation runs and a current-status count would reset the tally. The SPA advances the property itself on each approval (`recordPostApproval`), so the ask fires on the approval that crosses the threshold rather than one session later.

### No double-prompting

`posthog-surveys-enabled` (registry flag, `POSTHOG_SURVEYS_ENABLED`, default off) makes `next_survey_for_user` pass `nps_enabled=False` into `select_survey`, retiring `nps_day3` + `nps_trial_end` from **both** the in-app modal and the daily email beat. The review offer is untouched. It fails open to the env var, and a flag-lookup exception keeps the homegrown asks — the failure mode is "asked the old way", never "asked twice". Answering in PostHog also closes the homegrown ask on its own with no flag involved, since the `feedback` row is what `get_latest_feedback_at(user_id, NPS)` reads.

### Provisioning

`scripts/posthog_surveys.py` is the ONE place the surveys are defined — same pure-spec/plan + thin-client split as `posthog_provision.py`.

```bash
python scripts/posthog_surveys.py --print-spec       # both specs as JSON, no network
python scripts/posthog_surveys.py --dry-run          # diff (exit 2 = pending)
python scripts/posthog_surveys.py --apply --launch   # converge, then start
```

`--apply` creates **drafts**; `--launch` is a separate opt-in so an apply can never start collecting from a definition nobody has read. Drift is measured on a narrow managed-field set (appearance tweaks in the UI are not drift) and property values compare by value, because PostHog echoes numbers back as strings.

## Testing

- `tests/unit/utilities/test_surveys_posthog.py` (30) — response → feedback row, resolve-happy-answers, the no-double-count guard, and the homegrown stand-down. Includes a contract test holding the script's survey names/thresholds to `utilities/surveys.py`.
- `tests/unit/test_posthog_surveys.py` (35) — specs, targeting rules, drift detection, planner, CLI exit codes.
- `tests/unit/api/test_survey_api.py` (+6) — the endpoint, including a CSAT `score=6` that is a legal NPS score and an impossible CSAT one.
- `tests/integration/test_survey_capture.py` (+4) — through the real db.py SQL: a detractor lands as `new`, a happy CSAT as `resolved`, and the NPS ask provably disappears with the flag on.
- `src/cqc_lem/ui/src/hooks/usePostHogSurvey{,.flow}.test.tsx` (21) — parse + `survey sent` payload shape (a 0 score is not a skipped answer), and the flow: one `survey sent` across a failed-POST retry, never a `dismissed` after an answer.

Run locally: `pytest tests/unit -q` → 7114 passed (2 pre-existing failures in `test_error_capture_middleware.py`, present on the branch base — verified with `git stash`). `npm test` → 112 passed; `npm run build` clean.

## Rollout

1. `python scripts/posthog_surveys.py --apply --launch`
2. Flip `posthog-surveys-enabled` (or set `POSTHOG_SURVEYS_ENABLED=true`)

Nothing changes until both happen. Full verification steps in `docs/surveys.md`.

Closes #653

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -537,6 +537,12 @@ POSTHOG_FLAG_POLL_SECONDS=30
 # Fleet default for `feed_fallback_when_empty` for users who have never SAVED engagement
 # preferences. A user's own saved setting always wins over this and over the flag.
 FEED_FALLBACK_WHEN_EMPTY_DEFAULT=true
+# --- Surveys (issue #653, docs/surveys.md) ---
+# ON hands NPS + the post-quality CSAT to PostHog Surveys (targeting, cadence, the 30-day
+# wait-between-surveys throttle) and stands the homegrown NPS asks down so nobody is prompted twice.
+# The bespoke asks — the review that unlocks the extended trial, the per-issue fix CSAT — are
+# unaffected either way. Provision the surveys first: `python scripts/posthog_surveys.py --apply`.
+POSTHOG_SURVEYS_ENABLED=false
 # --- PostHog in the SPA (issue #646) ---
 # Read by Vite at BUILD time and baked into the bundle, so they must be set for `npm run build`
 # (docker build-arg / CI env), not in the running container. Leave VITE_POSTHOG_KEY empty and the
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -394,6 +394,36 @@ stdlib-only. The SPA bootstraps from `GET /api/flags` (server-resolved, so no fl
 disagreement between browser, API and workers) via `hooks/useFeatureFlags.ts` — NOT through
 posthog-js, which stays the analytics surface. Full posture: `docs/feature-flags.md`.
 
+### Surveys — NPS/CSAT via PostHog (issue #653)
+
+LEM's four asks now have TWO owners, and the line between them is what an answer can DO. PostHog
+Surveys owns the two general-purpose ones — the **NPS** at 30 days past ACTIVATION (not signup; a
+stalled user's score is about onboarding) and the **post-quality CSAT** triggered by `post_approved`
+once `posts_approved >= 5` — because their targeting and cadence should be tunable without a deploy.
+`utilities/surveys.py` keeps the two bespoke ones: the trial-T-3d **review** (it is the #499
+extended-trial gate, which PostHog cannot unlock) and the per-issue **fix CSAT** (#502).
+
+Both PostHog surveys are type **`api`** and rendered headless in `PostHogSurveyModal.tsx`, not as
+PostHog's popover — a popover answer is a PostHog event and NOTHING ELSE, and the reason LEM asks for
+scores at all is the feedback→auto-work loop. So one answer takes two paths and is counted ONCE: the
+browser emits PostHog's own `survey sent` (native `$survey_response`/`$survey_questions` shape, so it
+shows up in the Surveys product), and the same answer POSTs to `/api/survey/posthog` where it becomes
+a `feedback` row that the classifier files as an issue. `track_survey_response` is deliberately NOT
+emitted on that path — two events per answer would double every response-rate number. A detractor
+(NPS ≤6 / CSAT ≤2) or ANY free text stays `new` so the loop files it; a happy score with no comment
+is stamped `resolved` on the spot rather than burning an LLM call per promoter.
+
+Rendering it ourselves means posthog-js never sees the ask, so `markSurveySeen()` is what advances
+its seen/wait-period bookkeeping — drop it and the 30-day `seenSurveyWaitPeriodInDays` throttle
+silently stops working. Targeting reads person properties the SPA sets at `$identify` off
+`GET /auth/session` (`onboarding_completed_at`, `posts_approved` — counted across
+approved/scheduled/posted, because an approved post moves on and a current-status count would reset
+the tally). `scripts/posthog_surveys.py` is the ONE place the surveys are defined (`--dry-run` diffs,
+`--apply` converges, `--launch` is a separate opt-in so an apply never starts collecting from a
+definition nobody read). The `posthog-surveys-enabled` flag stands the homegrown NPS asks down
+(`select_survey(nps_enabled=False)`) so nobody is prompted twice; the review offer is untouched.
+Full posture: `docs/surveys.md`.
+
 ## CI Gates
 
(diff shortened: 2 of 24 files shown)


gitchrisqueen:

- **`--apply --launch` left a virgin project on two drafts**: `plan_actions` only planned a `launch_survey` for a survey that was already in `existing`, so on the first run both surveys were created as drafts and nothing launched — the `docs/surveys.md` rollout ("`--apply --launch`, then confirm both surveys are **Running**") would fail, and nothing in the output said to run it again. A create now plans its own launch, `apply_actions` resolves the id from the create it just did, and a launch whose create returned no id logs `could not launch survey '<name>'` instead of PATCHing `/surveys/None/`. Tests: first-run launch is planned, both ids are launched on apply, and the create-failed path is asserted.
- **`survey shown` + `markSurveySeen()` fired for an ask the modal never drew**: the suppression rule (`homegrown?.survey`) lived in `PostHogSurveyModal`, but the hook emitted `survey shown` and called `markSurveySeen()` unconditionally. With a bespoke ask due (the trial T-3d review), or simply before `GET /user/survey` had landed, PostHog was told the user saw a survey that was never rendered — which understates the response rate and, because `markSurveyAsSeen` also writes `lastSeenSurveyDate`, spends the 30-day `seenSurveyWaitPeriodInDays` that silences **both** surveys on an ask nobody saw. The gate now sits on the lookup: `usePostHogSurvey(paused)`, with the modal passing `isLoading || !!homegrown?.survey` (`isLoading`, not `isPending` — a query disabled for want of a session token stays pending forever). Flow test added.

Checked and found sound: `posthog-js@1.407` really does expose `markSurveyAsSeen(id, {iteration})` with exactly this contract, and `IN_APP_SURVEY_TYPES` includes `api`, so `getActiveMatchingSurveys` does return these surveys; the CSAT's `repeatedActivation` bypasses the per-survey seen check but **not** the global wait period, so the throttle still holds; the `post_approved` capture precedes the eligibility re-check, so the SDK's event receiver has the trigger before the reload; `_normalize` checks `bool` before `int`; the `posts_approved` tally spans approved/scheduled/posted; the endpoint's per-kind score bounds reject a CSAT 6; and the flag is read at the call site, not at import.

Verified: `pytest tests/unit -q` → **7118 passed**; `tsc -b` clean. (`vitest` needs Node ≥20 and this runner has 18, so the SPA suite is CI's to confirm.)

gitchrisqueen:

PR #691, opened 2026-07-27
feat(analytics): PostHog Endpoints stats panel + deploy-time release annotations (closes #654)

### LINKED ISSUE #654: PH9 — HogQL Endpoints for in-SPA stats + release annotations from the deploy pipeline
## Context
Two cheap wins. (1) PostHog Endpoints (beta, currently free) publishes a HogQL query as a versioned cached HTTP endpoint — the fast path to an in-SPA 'your engagement stats' panel without building a FastAPI reporting layer. (2) PostHog annotations are API-writable and mark releases on every insight graph — LEM ships multiple releases/day and no graph shows when.

## Scope
- Define 2-3 endpoints (posts/engagement by week, comment activity, LLM cost by feature) via the provisioning script from PH5; call them from a small SPA stats panel (auth: project API key server-side proxy via existing FastAPI route — do not expose personal keys to the browser).
- Add an annotation POST to the release pipeline (build-and-push or deploy workflow): 'vX.Y.Z deployed' at deploy time — instantly explains every metric step-change.
- Graceful degradation when endpoints/key absent.

## Acceptance
- Stats panel renders from endpoint data in dev; annotation appears after a deploy; provisioning idempotent.

gitchrisqueen:
## What & why

Two cheap PostHog wins from issue #654:

1. **In-SPA "Live stats" panel via PostHog Endpoints (beta).** `scripts/posthog_provision.py` now
   also provisions three per-user HogQL Endpoints — posts/engagement per week, comment activity per
   week, and 30-day LLM cost by feature (reading `llm_call`, matching the existing money-question
   convention in `docs/llm-analytics.md`) — instead of building a bespoke MySQL reporting layer for
   metrics PostHog already has cached and versioned.

   Every endpoint query is scoped with `distinct_id = {variables.distinct_id}`, resolved against a
   single provisioned `InsightVariable`. This matters because PostHog is **one project shared by
   every LEM account** — an un-scoped query would leak one customer's numbers into another's
   Dashboard. An endpoint is reported `blocked_endpoint` (not creatable) until the variable exists,
   mirroring the existing `plan_alerts` "missing prerequisite" shape.

   `src/cqc_lem/utilities/posthog_endpoints.py` + `GET /user/posthog-stats` is the server-side
   proxy: the personal API key never reaches the browser, and every failure mode (no key, endpoint
   not yet provisioned, PostHog unreachable) degrades to `available: false` for that one panel
   rather than breaking the response or the Dashboard page. `PostHogStatsPanel.tsx` renders nothing
   at all if every panel comes back unavailable.

2. **Release annotations.** `scripts/posthog_annotate.py`, called from `build-and-push.yml`'s
   deploy job right after `deploy.sh`'s own health check passes, posts one project-scoped
   `"vX.Y.Z deployed"` annotation so every insight graph explains its own step-changes — LEM ships
   multiple releases a day and no graph showed when before this. The step runs with
   `continue-on-error: true` and the script itself exits 0 on a missing key, so a PostHog hiccup
   can never fail a release that already shipped.

## Owner action needed

The deploy-annotation step reads a **new** GH Actions secret, `POSTHOG_PERSONAL_API_KEY` (scope:
`annotation` read+write) — distinct from the local-only env var of the same name used by the
provisioning scripts. Until it's added, the step is a harmless no-op (visible in the deploy log).
The existing `POSTHOG_PERSONAL_API_KEY` used to run `scripts/posthog_provision.py --apply` also
needs `endpoint` + `insight_variable` read+write scopes added for the Endpoints pass.

## Testing

- `poetry run pytest tests/unit -q`: 7266 passed (2 pre-existing, unrelated failures in
  `test_error_capture_middleware.py` reproduce on `main` before this branch — not touched here).
- New/updated Python coverage: `tests/unit/test_posthog_provision.py` (endpoint/variable specs,
  diff/plan logic, apply_actions, full `main()` happy path), `tests/unit/test_posthog_annotate.py`,
  `tests/unit/utilities/test_posthog_endpoints.py`, `tests/unit/api/test_api_coverage_misc.py`
  (`GET /user/posthog-stats`). `posthog_endpoints.py` is at 100% line coverage.
- UI: `npx vitest run` (118/118 passed incl. 5 new `PostHogStatsPanel` tests), `tsc -b --noEmit`
  clean, `npm run build` succeeds, `eslint` clean on all changed/new files.
- `scripts/posthog_provision.py --print-sql` and `scripts/posthog_annotate.py --dry-run` both
  verified locally (no network).
- `.github/workflows/build-and-push.yml` validated with `yaml.safe_load`.

Closes #654

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -515,6 +515,11 @@ POSTHOG_EXCEPTION_CAPTURE=true
 # LLM cache-hit rate (§E.2). Optional in the stack: unset just skips the cache-hit check.
 # Also read (scope query:read) by scripts/error_to_issues.sh, the daily error-tracking-issue ->
 # GitHub-issue cron.
+# Also needs endpoint + insight_variable read+write for scripts/posthog_provision.py's Endpoints
+# pass (issue #654, docs/kpi-dashboards.md) and annotation read+write for
+# scripts/posthog_annotate.py, called from build-and-push.yml's deploy job (as the GH Actions
+# secret of the same name) to mark every release on every insight graph. Both degrade gracefully
+# to a no-op without the key — no dashboards break, no deploy fails.
 POSTHOG_PERSONAL_API_KEY=
 # PostHog project the dashboards live in, and the app host used to build their links.
 POSTHOG_PROJECT_ID=475262
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -223,6 +223,7 @@ jobs:
       # passed by this point; a PostHog hiccup here must not fail an otherwise-successful release.
       # The script itself degrades gracefully (exits 0) when POSTHOG_PERSONAL_API_KEY isn't set.
       - name: Checkout (for the annotation script)
+        continue-on-error: true
         uses: actions/checkout@v7
         with:
           ref: ${{ needs.build-and-push.outputs.tag }}
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -77,6 +77,9 @@ jobs:
     env:
       CF_API_TOKEN: ${{ secrets.CF_API_TOKEN }}
       CF_ZONE_ID: ${{ secrets.CF_ZONE_ID }}
+      POSTHOG_PERSONAL_API_KEY: ${{ secrets.POSTHOG_PERSONAL_API_KEY }}
+      POSTHOG_PROJECT_ID: ${{ vars.POSTHOG_PROJECT_ID }}
+      POSTHOG_APP_HOST: ${{ vars.POSTHOG_APP_HOST }}
     steps:
       # SSH to the VPS can intermittently time out ("dial tcp :22 i/o timeout") — the block is
       # upstream of the box (Hostinger network / anti-abuse); on-box sshd + ufw are fine. Ride it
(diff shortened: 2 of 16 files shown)


gitchrisqueen:

- **CI resilience gap**: the new `Checkout (for the annotation script)` step in `build-and-push.yml`'s deploy job had no `continue-on-error`, unlike the `Post release annotation to PostHog` step right after it (and every other non-critical post-deploy step in this job). A transient `actions/checkout` failure would fail the whole "Deploy to Hostinger VPS" job and skip the annotation step entirely — directly contradicting the PR's own stated guarantee ("a PostHog hiccup here must not fail an otherwise-successful release... never a failed release"). Added `continue-on-error: true` to that checkout step so a hiccup there degrades to "no annotation posted" instead of a false-red deploy job.

Everything else held up: the Endpoints provisioning logic (variable-then-endpoint ordering, `blocked_endpoint` gating, drift detection incl. a recreated variable id forcing re-binding), the `/user/posthog-stats` proxy's per-panel degrade-to-`available:false` behavior and distinct_id scoping, and the SPA panel's loading/unavailable states all check out against the acceptance criteria and their own tests. Ran the full unit suite (`poetry run pytest tests/unit -q`) after the fix: 7268 passed, 0 failed.

gitchrisqueen:

gitchrisqueen:

Issue #695, opened 2026-07-27
Cut post impressions/saves over to the member post-stats API (needs r_member_postAnalytics)


gitchrisqueen:
Split out of #645, which settled the API question from the desk.

## Why

`post_stats.impressions` / `post_stats.saves` come from a **Selenium scrape** of
`/analytics/post-summary/<activity-urn>/` (`_post_analytics_counts` in `app/run_automation.py`).
It works, but it breaks on any SDUI relayout and rides the same 429 breaker as every other
Selenium lane — so stats stop whenever engagement automation is paused.

#645 established that LinkedIn **does** expose the same numbers over HTTP for the authenticated
member's own posts, with no browser and no DOM:

```
GET /rest/memberCreatorPostAnalytics?q=entity&entity=(share:urn%3Ali%3Ashare%3A<id>)
    &queryType=IMPRESSION|REACTION|COMMENT|RESHARE|POST_SAVE&aggregation=TOTAL
```

It also established **why LEM cannot read it today**: the endpoint requires the
**`r_member_postAnalytics`** permission, and LEM's authorize call asks for
`openid profile email w_member_social` only (`api/main.py`, `linkedin_auth_init`). The stored
member token has no claim on this surface and will answer `403`.

## What this needs from a human first (agent cannot do it)

1. **LinkedIn app approval** for the Community Management API product that carries
   `r_member_postAnalytics`.
2. **Every connected user re-consents** — adding a scope invalidates the existing grant, so this
   is a re-authorization prompt for real users, not a config flip.

Until both are done, the code change below is unlandable and the scrape stays. Answering #1/#2 is
also the decision on whether this is worth doing at all.

## Scope (once the permission exists)

1. Run the probe that #645 shipped and paste the JSON here:
   ```bash
   sudo docker exec -i celery_worker python - --user-id 1 --post-id <post id> \
       < scripts/linkedin_post_stats_api_probe.py
   ```
   It records HTTP status, response shape and the per-signal API-vs-scrape deltas.
2. Add `r_member_postAnalytics` to `linkedin_auth_init`'s scope list and handle the re-consent
   path for already-connected users.
3. Read the API in `auto_scrape_post_stats` as a **third source**, merged by the existing
   `max(detail, analytics, api)` rule, so a `403`/`426`/outage degrades to today's behaviour
   instead of zeroing a signal.
[...]

Issue #406, opened 2026-07-23
R3 (spike) — Format API feasibility (document + article/newsletter programmatic publish)


gitchrisqueen:
## Goal
Size C1 and de-risk the fragile newsletter Selenium path.

## Tasks
- [ ] Verify LinkedIn's document-post + article/newsletter API surface still supports programmatic publish (or must stay Selenium).

## Output
Feasibility note fixing C1 approach (API vs Selenium) + newsletter-publish risk assessment.

---
_Part of the LEM 2026 Growth Roadmap (gap analysis vs. 2026 LinkedIn algorithm authorities). Follows repo norms: feature branch `feature/claude-<task>`, ≥80% patch coverage, migrations advance from V56, CI gates green before merge._

🤖 Generated with [Claude Code](https://claude.com/claude-code)

gitchrisqueen:
Follow-up filed: #747 — recommendation **1** from `docs/FORMAT_API_FEASIBILITY.md` §4 (emit `log_error` when `_fill_and_publish_article` returns `None`) was never turned into an issue when this spike closed, and is still absent on `main`: both `auto_publish_newsletter_edition` and `auto_publish_edition` return a plain string on that path, so a silent non-publish never reaches PostHog.

Recommendation 2 (fold the article-editor selectors into the F2 #403 sweep) needs a live session and belongs with the next live-validation pass; recommendation 3 is already in the docstring.

Found by a phase-drop audit.

PR #697, opened 2026-07-27
feat(analytics): probe the member post-stats API as a scrape replacement (closes #645)

### LINKED ISSUE #645: Probe LinkedIn 202506 member post stats API as a churn-proof replacement for the analytics scrape
Split out of R1 (#404) per the owner's `3A` decision on PR #641.

## Why
Impressions and saves currently come from a **Selenium scrape** of `https://www.linkedin.com/analytics/post-summary/<activity-urn>/` (`_post_analytics_counts` / `_stacked_counts` in `app/run_automation.py`). That works and is grounded in a live 2026-07-23 capture, but it inherits two structural weaknesses: it breaks on any SDUI relayout, and it rides the same 429 breaker as the rest of the Selenium lanes — so stats stop when engagement automation is paused.

R3 (#406, `docs/FORMAT_API_FEASIBILITY.md`) recorded that LinkedIn's `202506` Community Management version added **member post stats**. If that surface exposes impressions and saves for the *authenticated member's own* posts, it replaces the scrape with an API read that no DOM churn or browser rate limit can break.

This is a **token probe, not a browser session** — no live LinkedIn UI interaction, so the pipeline can work it.

## Scope
1. Read `docs/FORMAT_API_FEASIBILITY.md` for what R3 actually recorded about `202506` member post stats, and confirm the current `LI_API_VERSION` against `scripts/linkedin_version_check.py`.
2. Probe the endpoint with the stored member token (read-only GET) for a known own-post activity URN. Record: HTTP status, whether the app is provisioned for it, and exactly which fields come back.
3. Compare the returned numbers against what `auto_scrape_post_stats` currently stores for the same post.
4. Write the finding into `docs/LIVE_VALIDATION_FORMAT_AND_STATS.md` §2 (replacing the "Follow-up lead (not verified here)" paragraph) with an evidence grade.

## Outcomes
- **Fields present and match** → open a follow-up to make the API the primary source with the scrape as fallback (merge-by-max already exists in `auto_scrape_post_stats`, so a third source slots in cleanly).
- **403 / not provisioned, or fields absent** → record it, keep the scrape, and add the analytics-page rows to the F2 (#403) selector-liveness sweep so drift is caught by cron rather than by a run of zeroes.

## Acceptance
- [ ] Probe run, HTTP status and response shape recorded (no secrets in the issue/doc).
- [ ] Doc §2 updated from lead to verified finding.
- [ ] Either a follow-up issue for the API cutover, or a recorded reason the scrape stays.

## Context
- `docs/LIVE_VALIDATION_FORMAT_AND_STATS.md` §2 (PR #641), `docs/FORMAT_API_FEASIBILITY.md` (#406)
- Scrape today: `app/run_automation.py` — `_ANALYTICS_URL`, `_post_analytics_counts`, `_stacked_counts`, `auto_scrape_post_stats`
- Related: B2 (#387), F2 (#403)

gitchrisqueen:
## What & why

Impressions and saves come from a **Selenium scrape** of `/analytics/post-summary/<activity-urn>/`
(`_post_analytics_counts` in `app/run_automation.py`). It works, but it breaks on any SDUI relayout
and rides the same 429 breaker as every other Selenium lane — so stats stop whenever engagement
automation is paused. R3 (#406) left a lead that LinkedIn's `202506` member post-stats surface might
replace it with an API read no DOM churn or browser rate limit can break.

**The lead is settled, and the answer is "yes, but blocked on a permission."**

`GET /rest/memberCreatorPostAnalytics?q=entity&entity=(share:…)&queryType=…` returns exactly the
signals LEM stores for the authenticated member's own posts — over plain HTTP, no browser:

| Stored signal | `queryType` | Live since |
|---|---|---|
| `impressions` | `IMPRESSION` | `202506` (the `q=entity` finder) |
| `reactions` / `comments` / `reposts` | `REACTION` / `COMMENT` / `RESHARE` | `202506` |
| `saves` | `POST_SAVE` | **`202604`** — later than the finder |

It requires the **`r_member_postAnalytics`** permission. `linkedin_auth_init` (`api/main.py`) asks
for `openid profile email w_member_social`, so **the stored member token will answer `403`** — an
in-repo verifiable fact, not a code defect. Granting it needs Community Management API approval on
the app *and* re-consent from every connected user (a scope change invalidates the grant), neither
of which an agent can do. **So the scrape stays**, and the cutover is #695 behind an owner decision.

## What shipped

- **`scripts/linkedin_post_stats_api_probe.py`** — read-only, stdlib-only, **no browser** (holds no
  Selenium lane, 429-immune). It separates the three answers that demand different fixes: `403`
  (permission), `426` (retired `LI_API_VERSION`), and a `400` naming the `queryType` (the pin
  predates the metric — LinkedIn's own documented error shape), then compares whatever it *does*
  get against the latest stored `post_stats` row.
- **`db.get_latest_post_stats`** — reads the scrape's own numbers back for that comparison. All SQL
  stays in `db.py`.
- **`docs/LIVE_VALIDATION_FORMAT_AND_STATS.md` §2a** — the "Follow-up lead (not verified here)"
  paragraph becomes a graded finding, plus the TL;DR row, §4 pointer, §5 scope update and sources.
- **#695** filed (`needs-human`, `risk:product-decision`) for the permission + cutover.

Four traps the doc and the probe both record, because each would have cost a debugging session:

- **The entity URN is the OPPOSITE of the scrape's.** The finder takes a `share`/`ugcPost` URN; the
  analytics *page* keys off the **activity** URN `_post_analytics_counts` resolves via the redirect.
  The ids are not interchangeable, so the probe **refuses** an activity URN rather than guessing.
- **One `queryType` per request** — five signals is five GETs per post, not one.
- **`metricType` changed shape in `202605`** (object → bare string). The probe parses both; a
  single-shape parser reads a good response from the other side of that line as empty.
- **`POST_SAVE` needs `202604`+.** LEM's `LI_API_VERSION` default `202606` clears it; a rollback
  would silently `400` saves.

## Honesty note — read this before merging

**The live HTTP status is not recorded.** This pass ran headless with no access to a member token
(the runbook forbids `docker`, prod paths and secrets), so acceptance item 1 is *not* met by this
PR. What it ships instead is the instrument that makes recording it a one-liner:

```bash
sudo docker exec -i celery_worker python - --user-id 1 --post-id <post id> \
    < scripts/linkedin_post_stats_api_probe.py
```

Every row in §2a carries an evidence grade — **Documented** (`li-lms-2026-07`) or **in-repo
verified** — and none is presented as a live observation. The expected result today is `403`; the
probe records that as evidence rather than leaving it a hunch.

## Testing notes

- `tests/unit/test_linkedin_post_stats_api_probe.py` — 59 tests over URN handling (including the
  activity-URN refusal), version floors, both `metricType` shapes, status classification, the
  API-vs-scrape comparison (an unknown side is **ungraded**, never a mismatch and never zero),
  verdicts, and that the token never reaches the report.
- `tests/unit/utilities/test_db_latest_post_stats.py` — the new reader, including user scoping and
  DB-error degradation.
- Full suite: `7424 passed` locally.

Closes #645

### DIFF
--- a/docs/LIVE_VALIDATION_FORMAT_AND_STATS.md
+++ b/docs/LIVE_VALIDATION_FORMAT_AND_STATS.md
@@ -119,7 +119,7 @@ things LEM cannot do for itself: the LinkedIn app approved for the Community Man
 **every connected user re-consenting**, because a scope change invalidates the existing grant.
 That is an owner decision, tracked as the follow-up below.
 
-**Four details that will bite whoever does the cutover:**
+**Five details that will bite whoever does the cutover:**
 
 - **The entity URN is the OPPOSITE of the scrape's.** This finder takes a `share` or `ugcPost`
   URN (`entity=(share:…)` / `entity=(ugc:…)`); the analytics *page* keys off the **activity** URN
@@ -133,12 +133,18 @@ That is an owner decision, tracked as the follow-up below.
 - **`202506` is not enough for saves.** `POST_SAVE` only exists from `202604`. LEM's
   `LI_API_VERSION` default is `202606` (`env_constants.py`), which clears every floor above — but a
   pin rolled back to satisfy something else would silently drop saves as a `400`.
+- **`aggregation=TOTAL` is the only one that serves every signal.** Under `DAILY`, LinkedIn does
+  not serve `IMPRESSION` for a post entity (nor `MEMBERS_REACHED`, `LINK_CLICKS`,
+  `FOLLOWER_GAINED_FROM_CONTENT`, `PROFILE_VIEW_FROM_CONTENT` at all) — and it says so with the
+  **same `400` "query type … metric type" shape as a version gap**, so the two are easy to confuse.
+  The probe flags the combination itself and names the aggregation ahead of the version, and the
+  cutover wants `TOTAL` anyway: `post_stats` stores lifetime counts.
 
 **The probe ships even though it is blocked:** `scripts/linkedin_post_stats_api_probe.py` is
-read-only, stdlib-only and needs no browser, and it separates the three answers that demand
-different fixes — `403` (permission), `426` (retired `LI_API_VERSION`), `400`/absent metric
-(version predates the metric) — then compares whatever it *does* get against the latest stored
-`post_stats` row:
+read-only, stdlib-only and needs no browser, and it separates the answers that demand different
+fixes — `403` (permission), `426` (retired `LI_API_VERSION`), `400`/absent metric (version predates
+the metric) and the `400` that is really an unsupported `aggregation` — then compares whatever it
+*does* get against the latest stored `post_stats` row:
 
 ```bash
 sudo docker exec -i celery_worker python - --user-id 1 --post-id <post id> \
--- a/docs/LIVE_VALIDATION_FORMAT_AND_STATS.md
+++ b/docs/LIVE_VALIDATION_FORMAT_AND_STATS.md
@@ -3,6 +3,8 @@
 **Spike:** R1 / issue #404 — "Live LinkedIn validation: document post path + saves/impressions scraping"
 **Grounds:** C1 (#390, native document/PDF posts) and B2 (#387, capture reposts/impressions/saves)
 **Date:** 2026-07-26 · **Probe:** `scripts/linkedin_live_validation.py` (read-only)
+**Updated:** 2026-07-27 (#645) — §2a settles the member post-stats API lead ·
+**Probe:** `scripts/linkedin_post_stats_api_probe.py` (read-only, no browser)
 
 ## TL;DR
 
@@ -13,6 +15,7 @@
 | Does a published document render as a **document card**? | Unconfirmed — needs one live post. | **OPEN** (C1's own acceptance) |
 | Are **saves** scrapeable? | **Yes**, own posts only, on `/analytics/post-summary/<activity-urn>/`. | **Live grab 2026-07-23** (encoded in `_stacked_counts`) |
 | Are **impressions** scrapeable? | **Yes**, same page (Discovery hero, value-first layout). Blank on the post detail view. | **Live grab 2026-07-23** |
+| Can an **API** replace that scrape? | **The endpoint exists** (`memberCreatorPostAnalytics`, impressions + saves + more) — but it needs `r_member_postAnalytics`, which **LEM's token does not request**. Scrape stays. | **Documented** (`li-lms-2026-07`) + **in-repo verified** (§2a) |
 
 **Net scope change: none for B2, one live confirmation left for C1.** Both features are built and
 merged; this spike's job was to say whether their assumptions hold, and they do. What is genuinely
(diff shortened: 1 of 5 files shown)


gitchrisqueen:

Reviewed as the merge gate: full diff, the surrounding scrape/token/log paths, the issue's acceptance criteria, and **LinkedIn's own docs** (I fetched `li-lms-2026-07` rather than trusting the PR's citation). The API surface the doc asserts is real and accurately transcribed — endpoint name, `r_member_postAnalytics`, the `q=entity` finder, `(share:…)`/`(ugc:…)` entity encoding, `POST_SAVE` from `202604`, and the `metricType` object→string change at `202605` all check out, as does the in-repo `linkedin_auth_init` scope claim that makes the 403 a fact rather than a guess. #695 exists with the right labels, and the "analytics-page rows folded into F2" outcome is backed by a real scope comment on #403.

- **A `--aggregation DAILY` 400 was indistinguishable from a version gap** (the defect): LinkedIn serves `IMPRESSION` under `TOTAL` only for a post entity — and `IMPRESSION` is in the probe's default metric set — but answers it with the *same* `"query type … metric type"` 400 shape `classify_status` uses to detect a version gap. So a DAILY run reported "some metrics failed", and on a pin below the floor `verdict()` actively told the reader to bump `LI_API_VERSION` over a flag they set themselves. Separating causes is the whole point of this script. Added `aggregation_supports()` over the documented DAILY-unsupported queryTypes, recorded it per metric, and made `verdict()` name the aggregation cause **ahead of** the version one (they can only collide on a rolled-back pin, and the flag is the cause the reader controls).
- **The report never recorded which aggregation it ran under**: it's meant to be pasted into #695, and without that field a metric that came back empty can't be told apart from one LinkedIn refuses to serve that way. `build_report` now carries it.
- **The 403 verdict named only the missing permission**: LinkedIn also 403s a post that isn't the authenticated member's own, which is the likelier cause once the permission *is* granted. Reason widened.

Also checked and found clean: no datetime leaks from `get_latest_post_stats` into the JSON report (`captured_at` is read but never serialised); the token never reaches stdout; `_resolve_from_db` only touches the DB for what wasn't passed in; the activity-URN refusal doesn't break the documented `--post-id` path, since the logged permalink stores the `/rest/posts` share/ugcPost URN, not the activity one; and `scripts/` is outside both the Codecov flag paths and ruff's `src/ tests/` scope, so the `print()`-to-stdout design is consistent with the other probes.

Not a finding, worth knowing at cutover time: `post_stats.reactions/comments/reposts/saves` are `NOT NULL DEFAULT 0`, so on those four columns "never captured" and "genuinely zero" are the same stored value — the ungraded-never-a-mismatch guarantee in `compare_counts` only really bites for `impressions`, the one nullable column. That's the existing schema, not this PR.

The honesty note is accurate and I left it standing: the live HTTP status genuinely is not recorded here, and every §2a row is graded Documented or in-repo verified rather than observed.

Tests: 71 in the two new files, full unit suite `7432 passed`.

gitchrisqueen:

PR #699, opened 2026-07-27
fix(automation): stop the missing-Connect-button error paging the cron (closes #571)

### LINKED ISSUE #571: bug(errors): Failed to find more or connect button (1x/24h)
**PostHog error (last 24h).** Logged message: `Failed to find more or connect button` — occurred **1** time(s) (first `2026-07-25T12:10:44Z`, last `2026-07-25T12:10:44Z`).

Auto-filed by the PostHog error-scan cron (seed run). Dedup marker (do not remove): `posthogerr-78f8bd86705d`

gitchrisqueen:
## What & why

PostHog filed `Failed to find more or connect button` as an error. It is not one — it is an ordinary outcome of the invite send path: the profile offered **no Connect affordance at all** (an invite is already pending, LinkedIn only shows Follow/Message on that profile, or the SDUI selector drifted). It already degrades gracefully — `invite_to_connect_now` returns `(False, reason)`, `send_connection_request` stores that reason on the `connection_requests` row and log_warnings it — so per the issue's own triage guidance it is now a `log_warning`.

**The real defect the error was hiding.** After the miss, the flow fell through into the "Add a note" / "Send without a note" steps *with no Connect dialog open*. Those can only fail, and their own `log_error` overwrote `result` — so a profile with no Connect option was recorded (and shown in the Connections review UI) as an *add-a-note* failure, and every such invite produced two error events instead of one honest one.

## Changes

- New `_click_connect_affordance(driver, wait, user_id)` — the direct `Invite ` button first, the More-actions overflow as the fallback (same two selectors as before, unchanged). A miss is one `log_warning` and `False`.
- A miss now **stops the invite** with a named reason (`NO_CONNECT_BUTTON_MESSAGE`, added beside `ALREADY_CONNECTED_MESSAGE` in `db.py`) and one `ENGAGED`/`FAILURE` log row — mirroring the #623 1st-degree early exit — instead of cascading into the note/send steps.
- No selector, cadence, cap or rate-limit behaviour changed; `LinkedInRateLimited` still propagates from login as before.

## Testing

- New `tests/unit/app/test_invite_connect_affordance.py` (6 tests): no-affordance → named reason + one log row; note/send steps never attempted after a miss; warning-not-error; direct button still sends; the More menu is still the fallback; a More menu without a Connect item is still a miss.
- `tests/unit/app/test_connection_requests.py` updated to the new failure-reason constant.
- `poetry run pytest tests/unit -q` → **7441 passed** locally.

Closes #571

### DIFF
--- a/src/cqc_lem/app/run_automation.py
+++ b/src/cqc_lem/app/run_automation.py
@@ -30,7 +30,7 @@ from cqc_lem.utilities.ai.content_alignment import humanize_text, split_link_for
     append_link_to_comment, resolve_artifact_delivery, ARTIFACT_KIND_LEAD_MAGNET
 from cqc_lem.utilities.date import convert_viewed_on_to_date
 from cqc_lem.utilities.db import get_user_password_pair_by_id, get_user_id, insert_new_log, LogActionType, \
-    CONNECTION_REQUEST_SENT_MESSAGE, ALREADY_CONNECTED_MESSAGE, \
+    CONNECTION_REQUEST_SENT_MESSAGE, ALREADY_CONNECTED_MESSAGE, NO_CONNECT_BUTTON_MESSAGE, \
     get_engagement_preferences, count_comments_today, get_recent_engagers, upsert_engager, \
     get_newsletter_settings, mark_newsletter_published, record_newsletter_subscriber_stat, \
     get_newsletter_edition, mark_edition_published, mark_edition_failed, \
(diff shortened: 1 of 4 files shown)


gitchrisqueen:

Reviewed the full diff plus the surrounding invite path (`invite_to_connect_now`, `send_connection_request`, `invite_to_connect`, `click_element_wait_retry`, `wait_for_ajax`, `count_invites_sent_today`, the Connections review UI). Hunted for a real defect at each seam and did not find one worth blocking on. What I specifically checked and cleared:

- **Early return vs. `try/except/else/finally`** — `return` inside the `try` skips the `else:` (which would have written a second `insert_new_log`) but still runs `finally: quit_gracefully(driver)`. Exactly one ENGAGED/FAILURE row per miss; `test_a_profile_with_no_connect_option_stops_with_a_named_reason` asserts `insert_new_log` is called once. Mirrors the pre-existing #623 1st-degree exit.
- **`LinkedInRateLimited` is not swallowed** by the new broad `except Exception` in `_click_connect_affordance` — every raise site is inside `login_to_linkedin` (`helper.py:459/466/528/536/619`), none is reachable from `click_element_wait_retry`. The docstring's re-raise contract still holds, and `send_connection_request` still defers (→ `APPROVED`) rather than marking FAILED.
- **No "clicked but reported as a miss" regression** — the only post-click call is `wait_for_ajax`, which swallows all exceptions (`selenium_util.py:613-619`), so a successful Connect click can't raise and be re-classified as absent. The old fall-through into the note/send steps was therefore never doing useful recovery work; removing it loses nothing.
- **Daily cap accounting unaffected** — `count_invites_sent_today` counts `ENGAGED` + `SUCCESS` + `message = CONNECTION_REQUEST_SENT_MESSAGE`, so the new FAILURE row can't inflate or deflate the invite budget. `record_action(ACTION_INVITE)` still only fires on a real send (`test_human_pacing_integration` green).
- **Reason plumbing** — `NO_CONNECT_BUTTON_MESSAGE` (51 chars) fits `connection_requests.failure_reason VARCHAR(512)` and renders as-is in `ConnectionRequests.tsx:182`; no string literal of the old message survives anywhere in the repo.
- **Fallback ordering** — tests pin that the overflow menu is tried only after the direct button (`_clicked(click)[:3]`), and that a More menu without a Connect item is still a miss rather than a half-open state.
- **Conventions** — `log_warning(..., exc=, user_id=, action_type=)` matches the CLAUDE.md logger contract; the constant sits beside `ALREADY_CONNECTED_MESSAGE` in `db.py`; no new SQL, enums, selectors, caps or rate-limit behaviour. (`myprint` and the untyped `driver`/`wait` params match the surrounding module, 112 `myprint` call sites — not worth churning.)
- **Acceptance criteria of #571** met: the `log_error` for this message is gone, the path degrades gracefully, and the second misleading error (`Failed to Add a note`) it used to cascade into is gone too.

Ran locally: `test_invite_connect_affordance.py`, `test_connection_requests.py`, `test_network_activation.py`, `test_human_pacing_integration.py` → 66 passed.

CI note: the red **E2E Tests** check is pre-existing and unrelated — `continue-on-error: true` (informational, does not block merge), failing on `test_carousel_video_post_e2e.py` with `redis.exceptions.ConnectionError: Error 111 connecting to localhost:6379`. The same job is failing on `main` (run 30234103764). All required gates are green.

gitchrisqueen:

Issue #634, opened 2026-07-26
Re-run the Selenium load test now that per-user stagger has landed (verify the 84% prediction)


gitchrisqueen:
## Context
Owner decision `1A` on PR #610: merge the load-test harness, implement the golden-hour stagger, **then re-run the load test**. The stagger shipped in the meantime — #554 closed COMPLETED via PR #607 (`perf(scheduler): stagger the fixed-time engagement fan-outs per user`), released in 0.95.3. So the remaining half of 1A is the verification pass.

## What
Once PR #610 merges, re-run `python -m cqc_lem.utilities.selenium_load_test` against the current code and confirm or refute the harness's own prediction that staggering takes **50 users from 57.7% → 84.0% on-time and sessions-needed from 14 → 11**.

## Scope
- Run `simulate` at 10 / 50 / 100 users; capture the table (and `--json`).
- Verify the harness's workload model actually reflects the shipped stagger (if the model still assumes a single 13:00 UTC fan-out, fix the model — that's the real deliverable, since a stale model silently invalidates every sizing number).
- Update the measured curve in `docs/scaling-plan.md` §5c and `docs/SELENIUM_GRID.md` with post-stagger numbers, replacing PR #610's pre-stagger table and noting both.
- If the prediction does NOT hold, say so plainly in the docs and open a follow-up — the point is a trustworthy number, not a confirmed one.
- Feed the resulting sessions-needed figures to the cost-comparison spike (companion issue) so its pricing math uses post-stagger reality.

## Acceptance
- Post-stagger curve committed to the docs; any model correction covered by unit tests (the harness is at 98% coverage — keep it there).
- Blocked until #610 merges.

Issue #554, opened 2026-07-25
infra: stagger fixed-time engagement fan-outs (per-user golden-hour offset)


gitchrisqueen:
## What
Replace the single `13:00 UTC` `daily-golden-hour-engagement` crontab with a **per-user offset** so N users don't all land one 15-minute `automate_commenting` loop on `se_engage` in the same minute. See docs/scaling-plan.md (PR #551) for the full rationale and measured numbers.

## Why
`auto_daily_engagement` currently fans out to all active users at one crontab. `se_engage` throughput is ~8 users/hour, so 50 users take ~6 hours to drain and most run hours after their intended peak window. Staggering is the single highest-leverage, lowest-cost change for on-time behavior.

## Proposed approach
- Derive a per-user offset from the user's local peak window (Login Location / timezone) or a hashed spread across a configurable window.
- Apply the same staggering pattern to other single-crontab fan-outs (appreciation DMs, group engagement) as needed.

## Acceptance
- Golden-hour dispatch spread across the window instead of a single spike (verify via Flower / queue depth).
- Per-user `QueueOnce` and daily caps unchanged.

## Risk
`risk:product-decision` — changes when engagement fires for every user; confirm it still hits each user's intended peak window.

Issue #553, opened 2026-07-25
infra: dedicated se_prepost lane so pre-post commenting never queues behind golden-hour loop (#547)


gitchrisqueen:
## What
Split a dedicated `se_prepost` Celery lane + worker for the ETA-based pre-post `automate_commenting` task dispatched by `auto_check_scheduled_posts`, so it can't queue behind a running 15-minute golden-hour loop on `se_engage`. See docs/scaling-plan.md (PR #551) for the full rationale and measured numbers.

## Why
`auto_check_scheduled_posts` enqueues `automate_commenting` with `eta = post_time − 15 min` onto `se_engage` (concurrency 2). When ≥3 users have posts in the same window, the pre-post burst fires late — the symptom in #547.

## Proposed changes
- Register `Queue('se_prepost')` in `celeryconfig.py`.
- Add `celery_worker_selenium_prepost` service (`SELENIUM_QUEUES=se_prepost`, concurrency 2) in both compose files.
- Route only the pre-post dispatch (`run_scheduler.py`) to `se_prepost` via `queue=`; leave daily golden-hour on `se_engage`.
- Bump `SE_NODE_MAX_SESSIONS` to cover the extra lane (coordinate with the session-cap issue).

## Acceptance
- Under contention (3+ near-simultaneous scheduled posts), pre-post commenting starts within its window.
- Existing golden-hour behavior unchanged.

Related: #547.

Issue #633, opened 2026-07-26
R4 (spike) — Cost analysis: hosted/cloud Selenium Grid options vs self-managed nodes


gitchrisqueen:
## Context
Owner decision on #556 / PR #610 was `2C` — *decide nothing on hardware now*, and instead research the alternatives before committing to a growth path. The load test in PR #610 measured today's ceiling: **100% on-time at 10 users, 57.7% at 50, 17.7% at 100**, needing ~14 concurrent sessions at 50 users and ~27 at 100 (32.4 GB Chrome, 28.5 vCPU) — beyond one 8-vCPU/31-GiB VPS. The two options documented so far are 'upgrade to 16 vCPU / 64 GB' or 'second VPS running Grid nodes'. This spike adds the third category: **someone else runs the browsers.**

## Questions to answer
- **AWS**: is there a managed Selenium Grid offering today, and what actually fits? Evaluate the realistic paths — ECS/Fargate-hosted Grid nodes, EC2 spot-backed nodes, AWS Device Farm's desktop-browser testing — with current on-demand pricing.
- **Commercial browser grids / Selenium-as-a-service**: BrowserStack Automate, Sauce Labs, LambdaTest, TestingBot, Browserless, Steel.dev, Browserbase, Zenrows/ScrapingBee-style browser APIs. For each: does it accept plain Selenium/WebDriver sessions (so `get_docker_driver()` just repoints at a remote hub URL), and does it allow **long-lived, logged-in, cookie-persisted sessions with residential egress** — or is it test-only infrastructure that would break LEM's anti-bot posture?
- **Deal-breakers to check explicitly** (these disqualify most test-focused vendors for our use case): datacenter egress IPs (see the HeyReach ban wave — LinkedIn enforces on infrastructure signature, and LEM deliberately routes per-user residential proxies via `utilities/proxy.py`); per-minute session billing vs our multi-minute logged-in sessions; ToS clauses forbidding social-media automation/scraping; ability to load our MV3 proxy-auth extension (`selenium_util.py`).
- **Self-managed baselines for comparison**: Hostinger 16 vCPU / 64 GB upgrade; second Hostinger VPS as a nodes-only box (`docker-compose.grid-node.yml` already exists); Hetzner/DigitalOcean equivalents.

## Output
[...]

PR #700, opened 2026-07-27
perf(infra): open the appreciation-DM window half its width early (closes #696)

### LINKED ISSUE #696: se_outreach: staggered appreciation-DM tail can bleed into the pre-post profile-viewer window
## Context
Issue #634 taught `selenium_load_test.py`'s workload model the REAL #554 stagger (each fan-out's own
shipped window — golden hour 180 min, appreciation DMs 120 min — instead of the pre-#554 single
13:00 UTC fan-out it was still assuming). Re-running the curve with the corrected model refuted the
docs' `84.0%` / 11-sessions prediction for 50 users: the actual measured result is **53.7% on-time,
15 sessions needed** — flat-to-worse than the pre-#554 baseline (57.7%, 14 sessions), not the
predicted improvement. Full numbers: `docs/scaling-plan.md` §5c, `docs/SELENIUM_GRID.md` §4.

## What's actually happening
`se_outreach` carries two independent job streams: `appreciation_dms` (now staggered across an
8:00–10:00-local window) and `profile_viewer_engagement` (post-anchored, arriving in the user's own
afternoon peak). Spreading `appreciation_dms`' arrivals over 120 minutes does **not** shrink the
total processing time its burst needs at a given concurrency (workload ÷ concurrency is fixed) — it
can instead push the batch's tail later in real (shared-clock) time, into the window
`profile_viewer_engagement` starts arriving in. Pre-#554, the single-instant DM batch happened to
fully drain before that window opened, hiding the contention; post-#554 it doesn't always.

This is a property of `se_outreach` being a **single shared lane** across the whole active fleet:
even though no individual user's own DM slot and post slot coincide, the fleet-wide DM backlog and
the fleet-wide pre-post backlog compete for the same 2 session slots.

Confirmed not a modelling bug: `tests/unit/utilities/test_selenium_load_test.py::TestRequiredTopology::test_the_shipped_stagger_can_shift_appreciation_dm_contention_onto_profile_viewer`
pins the isolated-lane arithmetic that produces it.

## Ask
Investigate and fix (pick one or combine, whichever the numbers support):
- Shorten `APPRECIATION_DM_WINDOW_MINUTES` and/or move `APPRECIATION_DM_ANCHOR_HOUR` earlier so the
  batch's tail reliably clears before `profile_viewer_engagement`'s window opens at scale.
- Raise `se_outreach` lane concurrency (costs a Chrome session slot — price it against §5c/§4).
- Split `appreciation_dms` onto its own lane so a DM backlog can never delay pre-post profile-viewer
  engagement (mirrors why `se_prepost` got its own lane in #553).

Re-run `python -m cqc_lem.utilities.selenium_load_test --users 10,50,100` after whichever change
lands and update the curve in `docs/scaling-plan.md` §5c / `docs/SELENIUM_GRID.md` §4 again.

## Acceptance
- `se_outreach`'s 95%-on-time sessions-needed at 50 users is no worse than the pre-#554 baseline (2).
- Covered by unit tests analogous to the ones added in #634.

gitchrisqueen:
## What & why

#634 measured the shipped #554 stagger and found `se_outreach` came out **flat-to-worse**: 53.7% on-time / 15 sessions at 50 users, with the lane needing **3** where the pre-#554 fixed-time fan-out needed **2**. The lane carries both the staggered `appreciation_dms` and the post-anchored `profile_viewer_engagement`.

The cause is that **a window OPENS on its anchor**. Anchoring at 08:00 — the hour the old crontab fired — does not leave the batch where it was: it moves the batch's midpoint half a window later and its **tail a full window later**, into the window `profile_viewer_engagement` starts arriving in. Spreading the burst never shrank the processing time it needs (workload ÷ concurrency is fixed), it only moved it; pre-#554 the single-instant batch happened to drain before that window opened.

So `APPRECIATION_DM_ANCHOR_HOUR` moves **08:00 → 07:00**, which puts the 120-minute window's *midpoint* back on the 08:00 off-peak hour and gives the batch back the drain time it had. `APPRECIATION_DM_MIDPOINT_HOUR` pins `anchor + window/2 = 08:00`, so widening the window later cannot silently re-create the collision, and the load-test model now derives the DM anchor from the shipped tuple instead of a literal — production and the model disagreeing about where this batch starts is what the issue was about.

**Users' experience:** appreciation DMs now go out 07:00–09:00 user-local instead of 08:00–10:00, so the *average* DM lands on the same 08:00 hour it does today rather than an hour after it.

## Re-measured — `python -m cqc_lem.utilities.selenium_load_test --users 10,50,100`

| Users | On-time (before → after) | Sessions needed | `se_outreach` |
|---|---|---|---|
| 10 | 100% → 100% | 5 → 5 | 1 → 1 |
| 50 | **53.7% → 64.3%** | **15 → 14** | **3 → 2** ✅ |
| 100 | 21.6% → 21.6% | 28 → 27 | 5 → 4 |

Acceptance met: `se_outreach` at 50 users needs **2**, the pre-#554 baseline. It is now no worse staggered than unstaggered at *every* scale from 10 to 200 users (pinned per-scale), and the stagger is finally a net win on the whole-fleet curve rather than something a lane-mate pays for.

## The two options not taken (priced in the docs)

- **Raise `se_outreach` concurrency** — buys a Chrome slot on a box already at its ceiling, and does not change the sessions-needed answer the SLO search reports.
- **Give `appreciation_dms` its own lane** (the #553 shape) — needs **4** slots across the two lanes at 50 users where the shared lane needs 2. A DM backlog could never delay profile-viewer engagement, but you pay for a browser that idles all afternoon.

## Testing

- `tests/unit/utilities/test_selenium_load_test.py`: the #634 finding test is replaced by `test_the_shipped_dm_window_costs_se_outreach_nothing_against_its_own_baseline`, parametrized over 10/25/50/75/100/150/200 users; plus the acceptance criterion at 50 users and a pin that the model reads its anchor off `STAGGER_APPRECIATION_DM` rather than a literal.
- `tests/unit/utilities/test_engagement_window.py`: the approved-defaults pin updated, plus the `anchor + window/2 == 08:00` invariant and a behavioural check that the mean dispatch minute across 500 users is still the approved hour.
- `poetry run pytest tests/unit -q` → **7445 passed**.

## Docs

`docs/scaling-plan.md` §5c/§5d and `docs/SELENIUM_GRID.md` §4/§5/§6 carry the new curve and the generalizable rule: *a fan-out that shares a lane with a tighter-tolerance job opens its window `window/2` early.* `docs/celery-flower-guide.md`'s beat row was still claiming a fixed 8:00 AM crontab (stale since #554) and now describes the staggered slot.

Closes #696

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -284,7 +284,11 @@ DEFAULT_POSTING_HOURS=16,15,16,17,11,10,18
 GOLDEN_HOUR_ANCHOR_HOUR=9
 GOLDEN_HOUR_WINDOW_MINUTES=180
 GOLDEN_HOUR_ANCHOR_TZ=local
-APPRECIATION_DM_ANCHOR_HOUR=8
+# Appreciation DMs open at 07:00 so the 2-hour window's MIDPOINT is the approved 08:00 off-peak hour
+# (issue #696): opening at 08:00 pushed the batch's tail two hours later, into the window the
+# post-anchored profile-viewer engagement arrives in on the same se_outreach lane. Move the anchor
+# and the window together — anchor = midpoint − window/2.
+APPRECIATION_DM_ANCHOR_HOUR=7
 APPRECIATION_DM_WINDOW_MINUTES=120
 APPRECIATION_DM_ANCHOR_TZ=local
 GROUP_ENGAGEMENT_ANCHOR_HOUR=12
--- a/.env.example
+++ b/.env.example
@@ -287,7 +287,9 @@ GOLDEN_HOUR_ANCHOR_TZ=local
 # Appreciation DMs open at 07:00 so the 2-hour window's MIDPOINT is the approved 08:00 off-peak hour
 # (issue #696): opening at 08:00 pushed the batch's tail two hours later, into the window the
 # post-anchored profile-viewer engagement arrives in on the same se_outreach lane. Move the anchor
-# and the window together — anchor = midpoint − window/2.
+# and the window together — anchor = midpoint − window/2. An EXISTING deployment whose .env still
+# says 8 does not get this fix from a release: change it to 7 there too (a mismatched pair logs a
+# WARNING naming the hour it actually centres on).
 APPRECIATION_DM_ANCHOR_HOUR=7
 APPRECIATION_DM_WINDOW_MINUTES=120
 APPRECIATION_DM_ANCHOR_TZ=local
--- a/docs/SELENIUM_GRID.md
+++ b/docs/SELENIUM_GRID.md
@@ -176,7 +176,11 @@ cohort onboarding without parsing the table. `--json` emits the whole thing mach
 
 ### Pre-#554 baseline (2026-07-26, today's topology: 8 slots, lanes 3/2/2/1)
 
-`--stagger-hours 0` reproduces the ORIGINAL single-13:00-UTC-fan-out behaviour, exactly:
+`--stagger-hours 0` reproduces the ORIGINAL single-crontab-minute behaviour, exactly — every user
+lands on their fan-out's anchor minute instead of a slot inside a window. (#696 moved the
+appreciation-DM anchor an hour earlier; this table is unchanged by that, because with no window the
+batch drains long before the post band opens either way. It is the control for what the WINDOW
+costs, not a historical replay of the 08:00 UTC crontab.)
 
 | Users | On-time % | Late jobs | Delay p95 | Sessions needed | Chrome mem | Host CPU | Verdict |
 |---|---|---|---|---|---|---|---|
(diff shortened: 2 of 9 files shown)


gitchrisqueen:

Held (`needs-human`, risk: product-decision). The fix works and CI-relevant tests pass, but it changes a default **you signed off on in PR #607** (`1A 2A 3A`): appreciation DMs anchored at 08:00 local. `tests/unit/utilities/test_engagement_window.py::TestApprovedFanoutDefaults` says in as many words that changing these numbers "needs the same sign-off again" — so it's yours to make, not mine.

Reply one letter — or `ok` to take the recommendation.

### 1. How should `se_outreach` stop paying for the appreciation-DM stagger?

- **A. Ship this PR — anchor 08:00 → 07:00, window stays 120 min** — DMs go out 07:00–09:00 user-local, so the *average* DM still lands on the approved 08:00 hour instead of an hour after it. `se_outreach` at 50 users: **3 → 2** (the pre-#554 number). Whole fleet: 53.7% → **64.3%** on-time, 15 → **14** sessions. No new Chrome slot, no new lane. Earliest a user could be DM'd is 07:00 their time instead of 08:00.  ✅ *recommended*
- **B. Keep the 08:00 anchor, narrow the window to 30 min instead** — nobody is DM'd before 08:00. Also gets 50 users to 2 sessions, but 30 min is two beat ticks, so it is barely a stagger at all, and at 100 users it needs **5** where the pre-#554 baseline needed 4 — it re-breaks one scale up.
- **C. Keep the 08:00 anchor, give `appreciation_dms` its own lane** (the #553 shape) — structurally guarantees a DM backlog can never delay pre-post profile-viewer engagement. Costs a Chrome session slot: **4** slots across the two lanes at 50 users where the shared lane needs 2, on a box §5c already calls "at ceiling", plus a compose lane + an `SE_NODE_MAX_SESSIONS` bump.
- **D. Change nothing** — send times stay exactly as approved; `se_outreach` keeps needing 3 at 50 users and the #554 stagger keeps costing a session on that lane. #696 closes as won't-fix.

**My recommendation: `1A`.** It is the only option that costs nothing and holds at every scale from 10 to 200 users, and it doesn't really move the DM send time — 08:00 was the window's *opening* hour, which meant the average DM already went out at 09:00 and the last one at 10:00; anchoring at 07:00 puts the average back on 08:00 where you approved it.

If you pick **B** or **C** I'll implement it on this branch (a `MODE=revise` tick) and re-run the curve; **D** and I'll close the PR and comment the reasoning on the issue.

gitchrisqueen:
1A

gitchrisqueen:
Addressed your review: taking `1A` — ship this shape, anchor 08:00 → 07:00, window stays 120 min. B/C/D declined, nothing about the fix itself changed.

**No code changes were needed** — `1A` is the diff that was already on the branch. What this tick actually did was get it mergeable and keep it honest against what landed on `main` while it was parked.

### Rebased onto `main` (it had gone `CONFLICTING`, 34 commits behind)

The conflict was #633's hosted-grid spike, which merged into the same two docs sections after this PR was written. Both halves are integrated rather than one clobbering the other:

- **`SELENIUM_GRID.md` §5** — #633's conclusions win where they contradict mine. My pre-rebase draft argued Option A (16 vCPU / 64 GB) was still the cheaper answer at 50 users; #633 found Hostinger has no tier above the current box, so A is a full cross-provider migration and **Option B is the default**. That paragraph is gone; #633's stands, now carrying the #696 curve (14 sessions / 64.3%) instead of #634's 15 / 53.7%.

- **The AWS cost tables — worth a look, this is the one non-obvious consequence.** #633 recomputed `scaling-cost-options.md` §1/§2b and `scaling-plan.md` §5f against #634's 15/28 curve, and that crossed a **discrete instance boundary**: 15 sessions needs a third c5.2xlarge ($745/mo) where 14 fits on two ($496/mo). **#696 removes that session again**, so those figures are restored to what they were originally priced at — Fargate `$454–578/mo` @ 50 users and `$826–1,115/mo` @ 100, EC2 `$496/mo` on 2 instances — and the sessions-needed table gains the 5/14/27 column. Left alone, this PR would have silently made a doc backing a purchasing decision overstate the 50-user AWS tier by ~$250/mo. **No verdict moves** — self-managed Option B is still cheapest at every tier, and none of #633's decisions are touched.

### Re-verified against rebased `main`, not just re-asserted

`python -m cqc_lem.utilities.selenium_load_test --users 10,50,100` on the rebased branch: **5 / 14 / 27** sessions at **100% / 64.3% / 21.6%** on-time, `se_outreach` back to **2** — the acceptance criterion and the pre-#554 number. Identical to the PR body's table, so nothing that landed meanwhile moved the curve. Full unit suite: **7495 passed**.

One thing to know rather than act on: `E2E Tests` was red on the previous head. It is not one of the four required gates and is unrelated to this diff (no Selenium/browser code changes here) — the force-push re-runs it, and I'll pick it up on a `MODE=fix` tick if it fails for a real reason.

Handing forward to merge.

gitchrisqueen:

Verified independently first: `python -m cqc_lem.utilities.selenium_load_test --users 10,50,100` on this branch reproduces **every** number the PR body and the three docs claim — 100% / 64.3% / 21.6% on-time, 5 / 14 / 27 sessions, `se_outreach` 1 / **2** / 4, 16.8 GB / 15.5 vCPU at 50 users. I also re-derived the "no worse at any scale" claim against the FULL fleet workload (the parametrized test checks the lane in isolation): staggered == unstaggered for `se_outreach` and for the total cap at 10/25/50/75/100/150/200, and staggered on-time is ≥ unstaggered at every one of those scales (57.7→64.3 at 50, 26.7→32.8 at 75, 17.7→21.6 at 100). The claims hold. `tests/unit` → **7500 passed**.

- **The pinned midpoint was not enforced anywhere at runtime — and the highest-probability outcome of merging this was that production never got the fix.** `APPRECIATION_DM_MIDPOINT_HOUR` was pinned only by a unit test on the code defaults, but the anchor and window are env-resolved at call time, and `.env.example` lists `APPRECIATION_DM_ANCHOR_HOUR` *explicitly* — so every existing deployment's own `.env` carries the pre-#696 `=8`, the release ships, and the batch goes on colliding with `profile_viewer_engagement` with zero signal. Same silent shape for the documented retune knob: `APPRECIATION_DM_WINDOW_MINUTES=240` without pulling the anchor back re-creates the exact regression the PR body says the pin prevents. That "nothing fails, the numbers just quietly move" mode is how #634 happened. **Fixed** in `engagement_window.py`: `PINNED_MIDPOINT_HOURS` + a check in `stagger_config` on the RESOLVED pair, logging one WARNING per process per distinct pair (it resolves once per user per 15-min tick, so an unmemoized line would be thousands a day) naming the hour the batch actually centres on and what to set. It warns, it never corrects — an explicit ops override still wins. 6 new tests: stale `=8` anchor caught, widened window caught, a matching retune (05:00 + 6h) stays silent, all three shipped defaults stay silent, memoized to one line. Deploy note added to `.env.example` and `SELENIUM_GRID.md` §4, and the two docs sentences that overclaimed the pin now describe what it actually does.

Looked at and deliberately left alone:
- **The load-test model reads the anchor off the code constant, not `stagger_config`** — so an env retune still leaves the model on the default's curve, which is a mild version of the "model and production disagree" problem the PR names. Left as-is on purpose: making a documented, reproducible curve depend on the ambient env is a worse trade, and the new runtime warning covers the case that actually matters (drift in production). Golden hour's anchor is a hard literal for the same reason.
- **`POST_BAND` / golden hour's `13 * 60` single-timezone convention** — pre-existing, and the one relationship #696 turns on (DM window vs post band) is self-consistent under it.
- Slot-claim behaviour across the deploy: the Redis claim is keyed by local date, so a user whose 08:xx slot already fired today is not re-dispatched at the new 07:xx slot, and a user past their slot catches up on the next tick. No double-send.

gitchrisqueen:

PR #701, opened 2026-07-27
fix(automation): stop an empty profile-views page failing the viewer task (closes #572)

### LINKED ISSUE #572: bug(errors): Error while engaging with profile viewers (1x/24h)
**PostHog error (last 24h).** Logged message: `Error while engaging with profile viewers` — occurred **1** time(s) (first `2026-07-24T22:51:14Z`, last `2026-07-24T22:51:14Z`).

Auto-filed by the PostHog error-scan cron (seed run). Dedup marker (do not remove): `posthogerr-d2d53b07fbff`

gitchrisqueen:
## What & why

`automate_profile_viewer_engagement` walks LinkedIn's profile-views analytics page with
`get_elements_as_list_wait_stale`, which polls `find_elements` until it returns a **non-empty**
list. An empty list is falsy, so a page with no recent viewers never resolves the wait — it
retries to a `TimeoutException`, which fell through to the task's outer handler and was logged
as `log_error("Error while engaging with profile viewers")`.

Confirmed against the PostHog log entry behind this issue (2026-07-24T22:51:14Z, user 1) — the
captured stacktrace is exactly that: `TimeoutException: Message: Finding Profile Viewers` raised
out of `get_elements_as_list_wait_stale` after its 3 retries. "Nobody viewed the profile" was
being reported as a task failure, and the error-scan cron filed it as a bug.

## Changes

- **Timeout / stale list finding the viewers** → `log_warning` + end the walk with zero viewers;
  the task completes normally (`Profile Viewer DMs Completed. Engaged with 0 viewers`). The dead
  `else: break` branch it replaces could never fire, since an empty list never returns.
- **The walk's own last-row read** (name + viewed-on, used to decide whether to keep scrolling)
  stops on a stale/missing node instead of failing the task — the list re-renders while we scroll it.
- **Name/URL extraction is now per-row**, so one unreadable row is skipped with a warning rather
  than aborting the whole batch; rows with no `href` are dropped.
- Anything genuinely unexpected still logs at ERROR — this is not a blanket downgrade.

## Testing notes

New `tests/unit/app/test_profile_viewer_engagement.py` (5 cases): timeout is a warning not an
error and dispatches nothing, stale list behaves the same, a stale row is skipped while the
readable one still queues `engage_with_profile_viewer`, and a non-Selenium failure still logs at
ERROR. Full `poetry run pytest tests/unit` passes (7446 tests).

Closes #572

### DIFF
--- a/src/cqc_lem/app/run_automation.py
+++ b/src/cqc_lem/app/run_automation.py
@@ -4396,6 +4396,7 @@ def automate_profile_viewer_engagement(self, user_id: int, loop_for_duration: in
         viewed_on_xpath = './/div[contains(@class,"artdeco-entity-lockup__caption ember-view")]'
 
         viewer_elements: list[WebElement] = []
+        previous_viewer_count = -1
 
         while True:  # Keep looping until we find a viewed on date out of range
             # Get Each Viewer within the last day (or time of dm run via database log)
@@ -4407,7 +4408,8 @@ def automate_profile_viewer_engagement(self, user_id: int, loop_for_duration: in
                 # An empty analytics page never resolves the wait — find_elements returns [] and the
                 # helper polls to a timeout. Nobody viewed the profile (or the list didn't render) is
                 # nothing to do, not a task failure, so it must not page the error cron.
-                log_warning("No profile viewers found to engage with",
+                log_warning(f"Could not read the profile viewers list — ending the walk with "
+                            f"{len(viewer_elements)} viewer(s)",
                             exc=e, user_id=user_id, task_name="automate_profile_viewer_engagement")
                 break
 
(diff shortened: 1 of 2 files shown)


gitchrisqueen:

- **Unparseable "viewed on" caption still failed the task**: `convert_viewed_on_to_date` raises `ValueError` (via `get_datetime`), and the call at the walk's date check sat OUTSIDE the new `try` (which only caught `StaleElementReferenceException`/`NoSuchElementException`). A caption LinkedIn renders in a shape `dateparser` can't read would hit the outer handler and log the exact `log_error("Error while engaging with profile viewers")` this PR exists to stop. Wrapped the caption read + parse; it now ends the walk with a warning.
- **All-or-nothing date filter leaked out-of-range viewers into the DM loop**: the `[e for e in viewer_elements if ...]` comprehension was wrapped in one `try/except Exception` that only logged — so a single unreadable caption abandoned the filter and left `viewer_elements` UNFILTERED. The walk breaks *on* an out-of-range viewer, so that row is always present, and the filter is the only thing keeping it out of `engage_with_profile_viewer` — i.e. an outreach DM to someone who viewed weeks ago. Rewrote it per-row; a row whose date can't be read is dropped, not engaged.
- **`while True` had no exit for the in-range case**: if every returned row is within 24h and scrolling stops adding rows (a low-volume account whose viewers fit on one screen — the same population as the empty page this issue is about), the loop spun forever, holding a Selenium session slot with a 2s sleep per turn. Added a growth check: when a scroll adds no new rows, the walk ends.
- **Misleading warning**: the new list-read handler logged "No profile viewers found to engage with" even on a mid-walk timeout where viewers had already been collected and were about to be DM'd. It now reports the count it is ending with.

Verified the core of the PR otherwise: `get_elements_as_list_wait_stale` does re-raise `TimeoutException`/`StaleElementReferenceException` after its 3 retries (so the new catch is the right one); `viewer_elements` is pre-initialised so the early `break` can't leave it unbound; a 429/auth-wall still surfaces as an ERROR upstream in `get_current_profile`, so this is not a blanket downgrade; and the outer `except` still errors on anything unexpected.

Tests: 4 new cases (walk termination on a non-growing list, mid-walk timeout keeping viewers already found, unparseable caption, and the filter not leaking the out-of-range row) — 9 in the file. Full `tests/unit` green: 7450 passed.

gitchrisqueen:

PR #702, opened 2026-07-27
fix(automation): make the connection-request note best-effort (closes #573)

### LINKED ISSUE #573: bug(errors): Failed to add a note to connection request (1x/24h)
**PostHog error (last 24h).** Logged message: `Failed to add a note to connection request` — occurred **1** time(s) (first `2026-07-25T12:11:21Z`, last `2026-07-25T12:11:21Z`).

Auto-filed by the PostHog error-scan cron (seed run). Dedup marker (do not remove): `posthogerr-88fb5dba94ef`

gitchrisqueen:
## What & why

`Failed to add a note to connection request` (issue #573) was a `log_error` on a path that did **not** degrade gracefully. Every step of the note branch — the "Add a note" button, the `#custom-message` textarea, typing, and the Send click — sat inside one `try`, and any miss abandoned an **already open** Connect dialog. The invite was never sent, the run recorded an ENGAGED/FAILURE row, and a plain send-without-note (which was right there) would have worked.

Two concrete causes behind the filed occurrence:

- LinkedIn hides the note affordance once a free account's personalized-invite quota is spent — the dialog then offers only "Send without a note".
- The note was typed **raw**. ChromeDriver's `send_keys` raises `WebDriverException` on non-BMP characters (the documented emoji gotcha), and an AI-drafted connection note routinely carries one. Every other composer in `run_automation.py` already goes through `_strip_non_bmp`; this one didn't.

## The change

- **`_add_connect_note`** — attaches the note to an open dialog. Strips non-BMP characters, refines an over-long note down to `CONNECT_NOTE_MAX_CHARS` (300, now a named constant instead of a repeated literal) and hard-caps it, and returns `False` on any miss with a **`log_warning`** — the invite still goes out, bare. That is the fix for the paging error: it now genuinely degrades.
- **`_submit_connect_invite`** — clicks Send, trying **both** aria-labels (`Send invitation` / `Send without a note`), preferred one first, because a partial note attempt can leave the dialog in either state. Failing *both* loses the invite outright, so that stays a `log_error` and reports the new `INVITE_NOT_SENT_MESSAGE` — which `send_connection_request` (#398) stores as the request's `failure_reason` instead of the old raw-exception f-string.
- `invite_to_connect_now` drops to three lines for the send: note (optional) → submit → result.

Behaviour note for review: an approved proactive request (#398) whose note can't be attached is now **sent without the note** rather than failed. It's still the invite the human approved, it still counts against the daily invite budget (`record_action` / `count_invites_sent_today` both key off `CONNECTION_REQUEST_SENT_MESSAGE`), and the dropped note is a WARNING in the logs.

## Testing

New `tests/unit/app/test_invite_connect_note.py` (7 cases): missing note affordance still sends; a `send_keys` BMP failure still sends; non-BMP stripped before typing; over-long note refined + capped; happy path types the note and clicks the right Send; a no-note invite never opens the composer; a dialog with no Send button at all reports `INVITE_NOT_SENT_MESSAGE`, spends no budget, tries both labels, and logs an error.

`poetry run pytest tests/unit -q` → **7457 passed**.

Closes #573

### DIFF
--- a/src/cqc_lem/app/run_automation.py
+++ b/src/cqc_lem/app/run_automation.py
@@ -31,6 +31,7 @@ from cqc_lem.utilities.ai.content_alignment import humanize_text, split_link_for
 from cqc_lem.utilities.date import convert_viewed_on_to_date
 from cqc_lem.utilities.db import get_user_password_pair_by_id, get_user_id, insert_new_log, LogActionType, \
     CONNECTION_REQUEST_SENT_MESSAGE, ALREADY_CONNECTED_MESSAGE, NO_CONNECT_BUTTON_MESSAGE, \
+    INVITE_NOT_SENT_MESSAGE, CONNECT_NOTE_MAX_CHARS, \
     get_engagement_preferences, count_comments_today, get_recent_engagers, upsert_engager, \
     get_newsletter_settings, mark_newsletter_published, record_newsletter_subscriber_stat, \
     get_newsletter_edition, mark_edition_published, mark_edition_failed, \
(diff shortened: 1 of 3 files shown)


gitchrisqueen:

- **Lost-invite error was invisible to the error→issue loop**: `_submit_connect_invite`'s `log_error(...)` passed no `exc=`. Per `logger.py`, an exc-less `log_error` writes to PostHog Logs but never calls `_capture`, so it produces no fingerprinted `$exception` — and `scripts/posthog_error_issues.py` (the cron that filed #573) only files off ACTIVE PostHog *issues*. The one failure this PR deliberately keeps as an error, the one where the invite is genuinely lost, would therefore never page anyone. Fixed by keeping the last click exception from the two-label loop and passing it as `exc=`; test asserts it.

Checked and found clean:
- `LinkedInRateLimited` cannot be swallowed by the new broad `except Exception` blocks — every raise site is inside `login_to_linkedin`, which runs before the dialog is opened, so the re-raise contract at the top of `invite_to_connect_now` still holds.
- The `noted` short-circuit can't skip the send: the early `return` paths (1st-degree, no Connect affordance) are both before it, and the `else:` clause still logs FAILURE + withholds `record_action` on `INVITE_NOT_SENT_MESSAGE`.
- Both Send-label orderings are safe when the note half-fails: `wait_for_ajax` swallows its own exceptions, so a successful click can't fall through and re-click a second label.
- `INVITE_NOT_SENT_MESSAGE` (54 chars) is well inside `connection_requests.failure_reason` VARCHAR(512), and it replaces the old raw-exception f-string that could have hit that cap.
- `_strip_non_bmp` before `send_keys` matches every other composer in the file; the `[:300]` hard cap is the same truncation the textarea's own `maxlength` applied before.
- Behaviour change (an approved #398 note that can't be attached now sends bare and spends the invite budget) is explicitly stated in the PR body — flagging, not blocking.

`pytest tests/unit/app/test_invite_connect_note.py test_invite_connect_affordance.py test_connection_requests.py test_network_activation.py test_human_pacing_integration.py -q` → 73 passed.

gitchrisqueen:

PR #704, opened 2026-07-27
fix(ui): Dashboard Activity Feed fills its card then inner-scrolls (closes #583)

### LINKED ISSUE #583: fix(ui): Dashboard Activity Feed should fill the card height then inner-scroll (not fixed max-h-80)
## Problem
On the SPA **Dashboard**, the **Activity Feed** card doesn't expand to the card's height — the activity list is hard-capped at a fixed `max-h-80` (~320px). When the card is taller than that (it stretches to match its taller sibling in the grid row), there's **dead space below the list**, and when there are more than a few entries the extra items are hidden with **no obvious scroll affordance**. Users can't tell there's more activity.

## Where
`src/cqc_lem/ui/src/pages/Dashboard.tsx`:
- Card wrapper (~line 208): `<div className="bg-white rounded-lg shadow-sm border border-gray-200 p-5">` — no height, doesn't lay out as a fillable column.
- List (~line 215): `<ul className="space-y-2 max-h-80 overflow-y-auto pr-1">` — the fixed `max-h-80` is the bug.

## Fix
Make the Activity Feed **fill the card's available height, then scroll internally** when content overflows:
- Make the card a full-height flex column: `flex flex-col h-full` on the card wrapper (the dashboard grid already stretches cells; confirm `items-stretch`/`h-full` propagation so the card takes the row height).
- Replace the list's fixed `max-h-80` with **`flex-1 min-h-0 overflow-y-auto`** so it grows to fill the remaining card height and scrolls only when the content exceeds it. (`min-h-0` is required for a flex child to shrink and scroll.)
- Add a subtle **scroll affordance** so it's obvious more activity exists — e.g. a bottom fade/gradient when scrollable, or a small "scroll for more" cue — so overflow isn't invisible.
- Keep the empty state and the existing row markup unchanged; this is layout-only.

## Acceptance
- The Activity Feed fills its card's height (no dead space below the list when the card is tall), scrolls internally when entries overflow, and shows a visible cue that there's more to scroll. Verified at a few viewport heights; SPA `tsc -b` clean; no change to data/API.

gitchrisqueen:
## What & why

The Dashboard **Activity Feed** list was hard-capped at `max-h-80` (~320px). Because the card stretches to match its taller `Planned Tasks` sibling in the grid row, that produced **dead space under the list**, and with more than ~5 entries the rest were hidden with **no scroll affordance** at all.

Layout-only fix (`Dashboard.tsx` + one new hook):

- The card is a full-height flex column (`flex flex-col h-full`), and the list is a `flex-1 min-h-0 overflow-y-auto` scroll region — it grows into whatever height the row gives the card, and scrolls only when entries overflow.
- The card also carries `max-h-[calc(100vh-8rem)]`. Without a cap, dropping `max-h-80` means the 15-entry feed just makes the grid row taller and **never** inner-scrolls; the viewport cap is what keeps "scrolls when it overflows" true on short viewports.
- New `useScrollAffordance` hook tracks whether content is hidden above/below the scroll container, and the feed renders a bottom **fade gradient + "Scroll for more ↓"** cue while there is more below. It measures in the ref callback (commit time, children mounted) so the cue is correct on the **first paint**, then re-measures on scroll, on window resize, and via a `ResizeObserver` on the list — the feed re-polls every 30s, which changes what overflows without firing a scroll or resize event, and the container's own box never changes because it is height-constrained. No `ResizeObserver` (jsdom, ancient browser) degrades to scroll+resize only.
- The scroll region gets `tabIndex={0}` + `role="region"` so a keyboard user can reach content the container hides; the cue itself is `aria-hidden` decoration.
- **Unchanged:** the empty state, the row markup (re-indented only), and all data/API behaviour. Most of the diff's line count is that re-indent.

## Testing

- `src/hooks/useScrollAffordance.test.tsx` — 6 vitest cases: content fits (no cue), overflows on first measure, scroll → mid-list shows both cues → bottom drops the down cue, viewport resize re-measures, sub-pixel overflow ignored, unmount clean with no `ResizeObserver`.
- Full SPA suite: **142 passed** (13 files). `npx tsc -b` clean. `eslint` clean on the three touched files.
- Verified reasoning against both layouts: tall sibling → list fills the card (no dead space); 15 entries on a short viewport → capped card, internal scroll with the cue visible.

Closes #583

### DIFF
--- a/src/cqc_lem/ui/src/hooks/useScrollAffordance.test.tsx
+++ b/src/cqc_lem/ui/src/hooks/useScrollAffordance.test.tsx
@@ -1,4 +1,4 @@
-import { describe, expect, it, afterEach } from 'vitest'
+import { describe, expect, it, afterEach, beforeEach } from 'vitest'
 import { useCallback, useRef } from 'react'
 import { act, cleanup, render, screen } from '@testing-library/react'
 import { useScrollAffordance } from './useScrollAffordance'
@@ -42,6 +42,31 @@ function Probe({ metrics }: { metrics: Metrics }) {
 }
 
 const flag = (id: string) => screen.getByTestId(id).textContent
+
+type GlobalWithRO = { ResizeObserver?: unknown }
+const globalRO = globalThis as unknown as GlobalWithRO
+
+class FakeResizeObserver {
+  static instances: FakeResizeObserver[] = []
+  observed: Element[] = []
+  disconnected = false
+  callback: () => void
+  constructor(callback: () => void) {
+    this.callback = callback
+    FakeResizeObserver.instances.push(this)
+  }
+  observe(target: Element) {
+    this.observed.push(target)
+  }
+  unobserve(target: Element) {
+    this.observed = this.observed.filter((o) => o !== target)
+  }
+  disconnect() {
+    this.disconnected = true
+  }
+}
+
+const latestObserver = () => FakeResizeObserver.instances[FakeResizeObserver.instances.length - 1]
 
 afterEach(cleanup)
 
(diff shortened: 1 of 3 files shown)


gitchrisqueen:

- **Test gap: the `ResizeObserver` path shipped untested** (`src/cqc_lem/ui/src/hooks/useScrollAffordance.ts:59-63`). It is the load-bearing branch — the 30s activity re-poll changes what overflows *without* firing a scroll or resize event, and the height-constrained container's own box never moves, so only the `node.children` observation reports it. jsdom has no `ResizeObserver`, and the suite *pinned* that (`expect(typeof ResizeObserver).toBe('undefined')`), so deleting the `observe(child)` loop left all 6 tests green — I verified that mutation passes on the old suite and fails on the new one. Fixed in 9146d76: added a stubbed-observer block (observes container **and** children · re-measures when the list grows under an unchanged `clientHeight` · drops the cue when the list shrinks back to fitting · disconnects on unmount), and rewrote the no-observer case to own the global rather than assert jsdom's current shape, so jsdom shipping `ResizeObserver` one day can't silently drop that branch. SPA suite **146 passed** (13 files), `tsc -b` exit 0, `eslint` exit 0.

Checked and found clean (no change needed):
- `useScrollAffordance` is called unconditionally at Dashboard top level with no early return before it — no conditional-hooks hazard; `attach` is `useCallback([])`-stable so React never detaches/reattaches per render, and its block body returns `undefined` so React 19 can't mistake it for a ref cleanup function.
- Listener symmetry (`passive` isn't part of `removeEventListener` matching), observer teardown on node change, and the `attach(null)` detach path.
- No `ResizeObserver` feedback loop: the cue is `absolute` and sits *outside* the observed subtree, so showing it can't resize what's being observed.
- No stale cue across an empty→populated feed: `attach` re-measures on every (re)attach.
- `max-h-[calc(100vh-8rem)]` — Tailwind v4 normalizes math-operator whitespace in arbitrary values, and the same construction already ships in `ContentStudio.tsx` / `NewsletterQueue.tsx`.
- No CLAUDE.md surface touched (no logger/SQL/enum/tier-alias/Selenium), no data or API change; the `isHttpUrl(entry.post_url)` gate and `rel="noopener noreferrer"` are carried through the re-indent unchanged.

Notes, not blocking:
- `E2E Tests` is red on a pre-existing infra failure (`redis.exceptions.ConnectionError: Error 111 connecting to localhost:6379` in `tests/e2e/test_carousel_video_post_e2e.py`) — unrelated to this UI-only diff and not a required gate.
- Layout trade-off worth knowing: with a full 10-task `Planned Tasks` sibling the row can exceed `calc(100vh-8rem)`, so the capped Activity card leaves the row slightly uneven. Still strictly better than the `max-h-80` it replaces, and capping is what makes "inner-scrolls instead of growing the row" true — not worth blocking.

gitchrisqueen:

PR #705, opened 2026-07-27
fix(ui): Save All no longer hidden behind the feedback widget (closes #596)

### LINKED ISSUE #596: fix(ui): Save All button hidden behind Send feedback dialog
## Why
The "Save All" button on the automation settings page (/account?tab=automation) is now hidden behind the "Send feedback" button and dialog, preventing users from accessing it.

Auto-filed from in-app feedback #1 by the feedback→auto-work loop (`docs/launch-and-marketing-plan.md` §B.2/B.3). Classifier: bug/medium, risk `none`, confidence 0.92.

## Scope
- Address the reported bug in the `ui` area of the product.

## Files
- `src/cqc_lem/ui/src/`
- `tests/unit/`

## Acceptance
- All required CI gates pass.

gitchrisqueen:
## What & why

The Settings **Save All** bar (`SaveAllBar`) and the **Feedback / Report a bug** widget were each independently `fixed bottom-4 right-4 z-40`. Same corner, same z-index — so the one rendered later in `Layout` won, and that is the feedback widget. On `/account?tab=automation` the Save All button sat under the launcher pill permanently, and was completely buried once the feedback panel was opened.

This adds **one** `FloatingDock` in `Layout` that owns the bottom-right corner: a `flex-col-reverse` stack, so the first child (the always-present feedback launcher) stays anchored at the bottom and anything portaled in later stacks *above* it. `FeedbackWidget` gives up its own positioning and becomes that first child; `SaveAllBar` portals into the dock, which also pushes it clear when the feedback panel expands instead of being drawn under it.

`SaveAllBar` keeps its old `fixed` positioning when no dock is present, so a settings card rendered outside `Layout` (and the existing standalone tests) still gets its button.

## Changes

- `components/FloatingDock.tsx` (new) — the shared stack; gaps are `pointer-events-none` so the container never swallows clicks on the page behind it.
- `components/Layout.tsx` — wraps `<FeedbackWidget />` in `<FloatingDock>`.
- `components/FeedbackWidget.tsx` — drops `fixed bottom-4 right-4 z-40`; the panel gains `relative` so its close button stays anchored.
- `pages/account/SettingsSaveContext.tsx` — `SaveAllBar` portals into the dock when one exists, self-positions otherwise.

## Testing

- `components/FloatingDock.test.tsx` (new, 3 cases): Save All lands inside the dock and is ordered above the feedback launcher; no control pins itself to the corner any more (regression guard for the overlap itself); Save All stays a sibling of the *open* feedback panel rather than under it.
- `SettingsSaveContext.test.tsx`: added the no-dock fallback case.
- `npm test` — 150 passed (14 files). `tsc -b` clean. `vite build` succeeds.

Closes #596

### DIFF
--- a/src/cqc_lem/ui/src/components/FeedbackWidget.tsx
+++ b/src/cqc_lem/ui/src/components/FeedbackWidget.tsx
@@ -111,7 +111,7 @@ export default function FeedbackWidget() {
           ensureSessionRecorded()
           capture(EVENTS.feedbackOpened, { route: location.pathname })
         }}
-        className="fixed bottom-4 right-4 z-40 bg-blue-600 text-white text-xs font-semibold px-3.5 py-2 rounded-full shadow-lg hover:bg-blue-700 transition-colors"
+        className="bg-blue-600 text-white text-xs font-semibold px-3.5 py-2 rounded-full shadow-lg hover:bg-blue-700 transition-colors"
         aria-label="Feedback / Report a bug"
       >
         Feedback / Report a bug
@@ -120,7 +120,9 @@ export default function FeedbackWidget() {
   }
 
   return (
-    <div className="fixed bottom-4 right-4 z-40 w-[22rem] max-w-[calc(100vw-2rem)] bg-white border border-gray-200 rounded-xl shadow-xl p-4">
+    // Positioned by FloatingDock (issue #596); `relative` keeps the close button anchored to the
+    // panel now that the panel is no longer the fixed element.
+    <div className="relative w-[22rem] max-w-[calc(100vw-2rem)] bg-white border border-gray-200 rounded-xl shadow-xl p-4">
       <button
         onClick={close}
         className="absolute top-2 right-3 text-gray-400 hover:text-gray-600 text-xl leading-none"
--- a/src/cqc_lem/ui/src/components/FloatingDock.test.tsx
+++ b/src/cqc_lem/ui/src/components/FloatingDock.test.tsx
@@ -42,39 +42,64 @@ function harness() {
 
 afterEach(cleanup)
 
+// The dock id marks the portal SLOT; the visual stack is its parent.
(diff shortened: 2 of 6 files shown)


gitchrisqueen:

- **Dock stacking order was last-render-wins, not structural** — `FloatingDock` mixed a React-managed child (`FeedbackWidget`) with portal-appended ones (`SaveAllBar`) in the *same* container. React re-appends a host node it re-renders to the END of its parent, so the first time the feedback panel opened (and on every close after), the widget became the last child and, under `flex-col-reverse`, rendered **above** the Save All bar — inverting the order the whole fix is built on. Reproduced in jsdom: dock children went `[launcher, saveAll]` → `[saveAll, panel]` → `[saveAll, launcher]`. Fixed by giving the dock a dedicated portal **slot** element rendered first inside a normal `flex-col` stack (`empty:hidden` so it contributes no gap when unused), so portal nodes and `children` never interleave. The ordering tests now assert across a full open/close of the panel — all three fail against the previous implementation.

Also checked, no findings: hook order in `SaveAllBar` (all hooks precede both early returns); the mount-time `getElementById` lookup (Layout renders the dock unconditionally in the same commit, so it is always present when the effect runs, and the no-dock fallback covers standalone use); the `pointer-events-none` / `[&>*]:pointer-events-auto` pair and `empty:hidden` are both emitted by Tailwind v4 in the production bundle (`.[&>*]:pointer-events-auto>*{pointer-events:auto}`, `.empty:hidden:empty{display:none}`), so the dock is not click-through; `z-40` still sits under the `z-50` modals; no other component pins itself to `bottom-4 right-4`; the panel's `relative` correctly re-establishes the containing block for its absolute close button.

Verified locally on Node 22: `npm test` 151 passed (14 files), `tsc -b` clean, `vite build` succeeds.

gitchrisqueen:

PR #707, opened 2026-07-27
fix(feedback): attach labels/assignees after issue creation (closes #598)

### LINKED ISSUE #598: fix(feedback): auto-filed issues land label-less — fine-grained PAT drops labels at creation
## Symptom
The feedback→auto-work loop went live 2026-07-25 (v0.92.1 + `FEEDBACK_GITHUB_TOKEN`) and filed its first issue **#596** from feedback row #1 — but with **zero labels**. Every future auto-filed feedback issue will be the same: no `agent:ready`, no `feedback-loop`, no category/severity. Result: the pipeline never picks them up and the loop is effectively non-functional (issues are invisible). #596 was hand-labeled as a stopgap.

## Root cause
`create_github_issue()` (`src/cqc_lem/utilities/feedback/issue_service.py:522`) passes labels in the **issue-creation payload**:
```python
payload = {"title": title, "body": body, "labels": list(labels or [])}
if assignees: payload["assignees"] = list(assignees)
data = github_request("POST", "issues", payload)
```
`FEEDBACK_GITHUB_TOKEN` is a **fine-grained PAT (Issues: R/W)**, and GitHub **silently drops `labels`/`assignees` supplied in the create payload** for fine-grained tokens — the issue is created, but unlabeled/unassigned, with no error. `labels_for_issue()` is correct and all target labels exist in the repo (`agent:ready`, `feedback-loop`, `priority:*`, `bug`/`feature`/…, `risk:*`); they just never attach.

## Fix
Apply labels and assignees in **separate calls after creation** (these honor `Issues: write` reliably for fine-grained PATs):
```python
data = github_request("POST", "issues", {"title": title, "body": body})
number = data.get("number") if isinstance(data, dict) else None
if number is None:
    return None
if labels:
    github_request("POST", f"issues/{number}/labels", {"labels": list(labels)})
if assignees:
    github_request("POST", f"issues/{number}/assignees", {"assignees": list(assignees)})
```
(`POST /issues/{n}/labels` adds to a freshly-created issue, so ordering is fine. Keep it tolerant of a failed labels call — log a warning, don't lose the issue.)

## Acceptance
- A unit test asserts `create_github_issue` makes the follow-up `POST /issues/{n}/labels` (and `/assignees` when assignees given), not just the create.
- After deploy, the next auto-filed feedback issue carries `feedback-loop` + `agent:ready` (or `needs-human` + assignee for risky) with category/severity labels.

## Note
Also add a `risk:security` handling check: risky items get `needs-human` + owner-assigned, so the assignee path must work too (same fix covers it).

gitchrisqueen:
## What

`create_github_issue()` sent `labels` (and `assignees`) inside the **issue-creation payload**. GitHub silently drops both when the caller authenticates with a **fine-grained PAT** — which `FEEDBACK_GITHUB_TOKEN` is. The issue is created, no error is returned, and it lands with **zero labels**.

Consequence: every auto-filed feedback issue was invisible to the pipeline (no `agent:ready`, no `feedback-loop`, no category/severity) — the feedback→auto-work loop was effectively non-functional. #596 was hand-labeled as a stopgap.

## Why this fix

`POST /issues/{n}/labels` and `POST /issues/{n}/assignees` honour a fine-grained PAT's `Issues: write` reliably, so labels/assignees are attached in separate calls right after creation. `labels_for_issue()` was already correct and every target label exists in the repo — they just never attached.

The attach calls are **tolerant**: a failure logs a warning via `log_warning` and `create_github_issue` still returns the number. The issue is already on GitHub at that point, and returning `None` would strand the feedback row (it would be re-filed on the next pass as a duplicate).

All three callers benefit from the one-place fix — `file_feedback_issue` (feedback loop), `faq_service` (FAQ issues), and `capacity_alerts`. The `risk:security` path noted in the issue is covered too: that path is exactly the `needs-human` + owner-assignee case, which now goes through the working `/assignees` endpoint.

## Testing

`tests/unit/utilities/test_feedback_issue_service.py`:
- `test_create_issue_attaches_labels_and_assignees_after_creation` — asserts the exact three-call sequence (create with `{title, body}` only, then `/issues/777/labels`, then `/issues/777/assignees`) with the right payload on each.
- `test_create_issue_skips_the_attach_calls_when_there_is_nothing_to_attach` — no labels/assignees means exactly one call.
- `test_failed_label_attach_still_returns_the_issue` — a 403 on both attach calls still yields the issue number.

`poetry run pytest tests/unit/utilities/test_feedback_issue_service.py tests/unit/utilities/test_feedback_faq_service.py tests/unit/utilities/test_capacity_alerts.py -q` → **289 passed**.

Post-deploy acceptance (the issue's second criterion) is observable on the next auto-filed feedback issue: it should carry `feedback-loop` + `agent:ready` plus its category/severity labels.

Closes #598

### DIFF
--- a/src/cqc_lem/utilities/feedback/issue_service.py
+++ b/src/cqc_lem/utilities/feedback/issue_service.py
@@ -513,16 +513,29 @@ def github_request(method: str, path: str, payload: dict = None) -> Optional[obj
 
 def create_github_issue(title: str, body: str, labels: list,
                         assignees: list = None) -> Optional[int]:
-    """Open an issue; returns its number or None."""
-    payload = {"title": title, "body": body, "labels": list(labels or [])}
-    if assignees:
-        payload["assignees"] = list(assignees)
-    data = github_request("POST", "issues", payload)
+    """Open an issue; returns its number or None.
+
+    Labels and assignees are attached in SEPARATE calls after creation (issue #598): GitHub
+    silently drops both when a fine-grained PAT supplies them in the create payload, so every
+    auto-filed issue landed label-less and the pipeline never picked it up. A failed attach is a
+    warning, never a lost issue — the row is already on GitHub by then."""
+    data = github_request("POST", "issues", {"title": title, "body": body})
     number = data.get("number") if isinstance(data, dict) else None
     if number is None:
         return None
+    number = int(number)
+    label_list = [str(label) for label in (labels or [])]
+    if label_list and github_request("POST", f"issues/{number}/labels",
+                                     {"labels": label_list}) is None:
+        log_warning(f"Auto-filed issue #{number} could not be labeled {label_list}",
+                    api_provider="github")
+    assignee_list = [str(assignee) for assignee in (assignees or [])]
+    if assignee_list and github_request("POST", f"issues/{number}/assignees",
+                                        {"assignees": assignee_list}) is None:
+        log_warning(f"Auto-filed issue #{number} could not be assigned to {assignee_list}",
+                    api_provider="github")
     log_info(f"Auto-filed feedback issue #{number}: {title}", api_provider="github")
-    return int(number)
+    return number
 
 
 def comment_on_issue(issue_number: int, body: str) -> bool:
--- a/tests/unit/utilities/test_feedback_issue_service.py
+++ b/tests/unit/utilities/test_feedback_issue_service.py
@@ -456,23 +456,52 @@ class TestGithubIO:
         response.json.return_value = payload if payload is not None else {}
         return response
 
-    def test_create_issue_posts_title_body_labels_and_returns_the_number(self, monkeypatch):
+    def test_create_issue_attaches_labels_and_assignees_after_creation(self, monkeypatch):
+        # Issue #598: a fine-grained PAT silently drops labels/assignees sent in the create
+        # payload, so they MUST ride on their own endpoints afterwards.
         svc = _mod()
         monkeypatch.setenv("FEEDBACK_GITHUB_TOKEN", "tok")
         monkeypatch.setenv("FEEDBACK_GITHUB_REPO", "o/r")
         requests = MagicMock()
-        requests.request.return_value = self._response(201, {"number": 777})
+        requests.request.side_effect = [self._response(201, {"number": 777}),
+                                        self._response(200, [{"name": "bug"}]),
+                                        self._response(201, {"number": 777})]
         with patch.dict("sys.modules", {"requests": requests}):
-            got = svc.create_github_issue("t", "b", ["bug"], assignees=["gitchrisqueen"])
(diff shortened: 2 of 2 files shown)


gitchrisqueen:
🔎 Claude adversarial review — PASS (no blocking findings)

Reviewed the diff against issue #598's acceptance criteria, the surrounding `github_request` seam, and all three call sites (`file_feedback_issue`, `faq_service.hold_for_human`, `capacity_alerts.file_capacity_issue`).

What I checked and found clean:
- **Injection / path safety** — `number` is `int()`-coerced before it reaches the `issues/{n}/labels` and `issues/{n}/assignees` paths, and label names are never free text: `labels_for()` maps enum taxonomy through fixed `CATEGORY_LABELS`/`SEVERITY_LABELS`/`RISK_LABELS` dicts, so an LLM classification cannot smuggle a label name.
- **Error paths** — `github_request` returns a `list` on a successful labels POST and `{}` on an unparseable body, so the `is None` success test is correct for both attach calls (a naive truthiness check would have mis-read the empty-list case). Both failures already log at ERROR inside `github_request`, so a silent attach failure still reaches PostHog; the added `log_warning` is the human-readable half.
- **Tolerant-return decision** — returning the number after a failed attach is right: the issue exists on GitHub at that point, and `file_feedback_issue` would otherwise leave the row `new` and re-file a duplicate on the next pass.
- **Acceptance criterion 1** — `test_create_issue_attaches_labels_and_assignees_after_creation` pins the exact three-call sequence and per-call payloads, which is what would catch a regression back to payload-based labels. The nothing-to-attach and 403-tolerance cases are both covered.
- **The `risk:security` note** — that path is `route == AUTO_WORK` + `is_agent_ready() == False` → `needs-human` label + `assignees=[issue_assignee()]`, already asserted at `test_feedback_issue_service.py:612`; `issue_assignee()` can never return empty (falls back to `DEFAULT_ASSIGNEE`), so no `{"assignees": [""]}` call is possible.
- No CLAUDE.md violations (structured logger with `api_provider`, type hints intact, no raw SQL, no secrets).

Tests: `tests/unit` full suite — **7481 passed**.

Two non-blocking observations, deliberately not changed (out of this PR's scope):
- `capacity_alerts._apply_issue_labels()` is now a redundant second labels POST; its own docstring anticipates this ("stays harmless once #598 fixes this at the source") and it is idempotent, so it costs one extra API call per capacity issue and nothing else.
- If the fine-grained PAT's label drop was caused by a permission gap rather than a create-payload quirk, `/issues/{n}/labels` would 403 instead of silently succeeding. That fails loudly (ERROR log + warning) rather than silently, which is the right failure mode — but the post-deploy check in acceptance criterion 2 is what actually confirms the fix.

gitchrisqueen:

PR #708, opened 2026-07-27
fix(feedback): reject unknown triage statuses before the ENUM write (closes #668)

### LINKED ISSUE #668: bug(errors): Could not update feedback triage for 1 (2x/24h)
**PostHog error (last 24h).** Logged message: `Could not update feedback triage for 1` — occurred **2** time(s) (first `2026-07-25T19:20:02.542433Z`, last `2026-07-25T21:35:11.756556Z`).

Auto-filed by the PostHog error-scan cron. Dedup marker (do not remove): `posthogerr-92bd9a76a6dc`

gitchrisqueen:
## What & why

The PostHog error-scan cron filed #668 for `Could not update feedback triage for 1` (2× on 2026-07-25). Pulling the captured stack trace out of PostHog Logs gives the real cause:

```
mysql.connector.errors.DatabaseError: 1265 (01000): Data truncated for column 'status' at row 1
  File "/app/src/cqc_lem/utilities/db.py", ... in update_feedback_triage
```

`feedback.status` is a MySQL ENUM (`new`/`triaged`/`clustered`/`issue_created`/`resolved`/`dismissed`), but `update_feedback_triage` accepted any `status` and pushed it through `str()` into the UPDATE. An out-of-vocabulary value therefore:

1. failed only at the database, as an opaque 1265 that names no value and no caller, and
2. took the **whole statement** down with it — a call that also carried `cluster_id`, `github_issue_number` or `embedding` lost all of them, leaving the row unstamped for the next pass to redo.

Both prod occurrences were isolated writes against row 1 with no surrounding task logs (the beat passes at `:00`/`:30` all succeeded, and the queue drained to `0 row(s)` right after — so the deployed ENUM does accept every `FeedbackStatus` member). The defect is that the function has no vocabulary check at all, so an ad-hoc or future caller passing a raw string turns into an unactionable production error.

## Changes

- `update_feedback_triage` coerces `status` through `FeedbackStatus` **before** opening a connection. Whitespace/case tolerant (`" Issue_Created "` → `issue_created`), so nothing that worked under MySQL's case-insensitive ENUM matching stops working.
- An unknown status is refused up front and logged by **name**, next to the allowed vocabulary — an actionable message instead of a truncation error. Still `log_error`: a caller passing a status that doesn't exist is a bug worth paging for.
- The `mysql.connector.Error` handler now names the columns that were in flight, so the next 1265 says which one rejected the value.
- Build-time guard: a unit test pins `FeedbackStatus` to the ENUM declared in `V20260725063146__add_feedback.sql`. Adding a member without a migration currently fails *only* in production, and only as this same 1265.

No migration and no schema change — the deployed ENUM is already correct.

## Testing

- `tests/unit/utilities/test_db_feedback.py` — unknown status never opens a connection and names the offending value; spelling variants normalize; every `FeedbackStatus` member round-trips to its ENUM value; the DB-error log names the columns.
- `tests/unit/api/test_feedback_api.py` — enum↔migration parity.
- `pytest tests/unit/utilities/test_db_feedback.py tests/unit/api/test_feedback_api.py` → 39 passed; the feedback/FAQ/survey service suites (268 tests) still pass.

Closes #668

### DIFF
--- a/src/cqc_lem/utilities/db.py
+++ b/src/cqc_lem/utilities/db.py
@@ -7950,7 +7950,8 @@ def count_feedback_filed_by_user(user_id: int, hours: int = 24) -> int:
         connection.close()
 
 
-def update_feedback_triage(feedback_id: int, status: "FeedbackStatus" = None,
+def update_feedback_triage(feedback_id: int,
+                           status: Optional[Union["FeedbackStatus", str]] = None,
                            cluster_id: int = None, github_issue_number: int = None,
                            embedding: list = None, sentiment: str = None) -> bool:
     """Stamp the auto-triage result back onto a feedback row (issue #498). Only the arguments you
@@ -7965,10 +7966,12 @@ def update_feedback_triage(feedback_id: int, status: "FeedbackStatus" = None,
     if status is not None:
         try:
             status = FeedbackStatus(str(status).strip().lower())
-        except ValueError:
+        except ValueError as err:
+            # exc= is what turns this into a grouped PostHog $exception (issue #648) — the 1265 it
+            # replaces was one, so without it the refusal would page nobody and only land in Logs.
             log_error(f"Refusing to write unknown feedback status {str(status)!r} for feedback "
                       f"{feedback_id} — expected one of "
-                      f"{', '.join(s.value for s in FeedbackStatus)}")
+                      f"{', '.join(s.value for s in FeedbackStatus)}", exc=err)
             return False
         updates.append("status=%s")
         params.append(str(status))
--- a/src/cqc_lem/utilities/db.py
+++ b/src/cqc_lem/utilities/db.py
@@ -7955,10 +7955,21 @@ def update_feedback_triage(feedback_id: int, status: "FeedbackStatus" = None,
                            embedding: list = None, sentiment: str = None) -> bool:
     """Stamp the auto-triage result back onto a feedback row (issue #498). Only the arguments you
     pass are written, so the filer can save an embedding on one pass and the cluster/issue on the
-    next without clobbering anything."""
+    next without clobbering anything.
+
+    `status` is checked against `FeedbackStatus` BEFORE the write (issue #668): the column is a
+    MySQL ENUM, so an out-of-vocabulary value used to surface as an opaque `1265 Data truncated for
+    column 'status'` AND roll back the cluster/issue/embedding travelling in the same UPDATE."""
     updates: list = []
     params: list = []
     if status is not None:
+        try:
+            status = FeedbackStatus(str(status).strip().lower())
+        except ValueError:
+            log_error(f"Refusing to write unknown feedback status {str(status)!r} for feedback "
+                      f"{feedback_id} — expected one of "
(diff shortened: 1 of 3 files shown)


gitchrisqueen:

Verified the diagnosis independently before reviewing the fix: both prod occurrences are in PostHog Logs with `_mysql_connector.MySQLInterfaceError: Data truncated for column 'status' at row 1` inside `update_feedback_triage`, one from a Celery-env process and one from `unknown_service` (an ad-hoc run) — so the PR is fixing the right thing, and no in-repo caller can produce it (every call site passes a `FeedbackStatus` member). `FeedbackStatus` is a stdlib `StrEnum`, so `str(member)` is the ENUM value on both 3.12 and prod's 3.13 — the new coercion round-trips every member rather than rejecting them.

- **The refusal path never reaches error tracking** (`db.py`): `log_error(...)` was called without `exc=`, so it emits no `$exception` — but the 1265 it replaces did. The PR body says a caller passing a nonexistent status is "worth paging for"; as written it would only land in PostHog Logs, invisible to `scripts/posthog_error_issues.py` and to alerting. Fixed: `except ValueError as err` → `log_error(..., exc=err)`, so the refusal groups as an issue exactly like the truncation did. Test asserts the `exc` kwarg.
- **The ENUM parity guard traps the next author into breaking Flyway** (`tests/unit/api/test_feedback_api.py`): it compared `FeedbackStatus` against the `add_feedback` CREATE TABLE only. Adding a member the *correct* way — a new `ALTER TABLE feedback ... MODIFY status ENUM(...)` migration — would still fail the test, and the obvious way to make it pass is editing an already-applied migration, which breaks Flyway's version+checksum validation on the prod deploy. Fixed: the guard now takes the **last** `feedback.status` declaration across all migrations in version order (statement-scoped to the `feedback` table, so `faq_entries.status` can't be picked up).
- **Signature no longer matched the contract** (`db.py`): `status` is now deliberately string-tolerant (`" Issue_Created "` normalizes), so it's typed `Optional[Union["FeedbackStatus", str]]`.

Checked and found clean: no other writer touches `feedback.status` (all SQL is in `db.py`, `insert_feedback` takes the ENUM default); every caller in `issue_service`/`faq_service`/`surveys` passes a real member; the deployed vocabulary matches the enum; `columns` in the DB-error handler is derived after the `not updates` early return; and refusing the whole UPDATE loses nothing the old 1265 didn't already lose.

Not changed (out of scope, flagging only): `update_dm_followup_status` pushes `str(status)` into the `dm_followups.status` ENUM with the same unchecked shape, and `update_feedback_triage(None, ...)` still silently no-ops on `WHERE id=NULL`.

`pytest tests/unit/utilities/test_db_feedback.py tests/unit/api/test_feedback_api.py tests/unit/utilities/test_feedback_issue_service.py tests/unit/utilities/test_feedback_faq_service.py tests/unit/utilities/test_surveys_posthog.py tests/unit/utilities/test_surveys_fix_csat.py` → 307 passed.

gitchrisqueen:

PR #720, opened 2026-07-27
feat(ops): scan Ollama Cloud retirements + new models before they bite (closes #716)

### LINKED ISSUE #716: Model-health check v2: advance retirement warnings + new-model discovery for Ollama Cloud
## Context
The weekly model-health cron (scripts/weekly_model_check.sh + model_health_check.py, Sun 09:00 UTC) is REACTIVE: it detects retirements only when a configured model already 410s. Ollama Cloud now publishes an authoritative retirement schedule IN ADVANCE, and new models appear ~monthly — we should know before a sunset bites and when a better model ships. (Research 2026-07-27: July batches retired qwen3-coder, deepseek-v3.x, glm-4.x, gemma3, ministral-3, kimi-k2, minimax-m2.x cloud tags; minimax-m2.5/kimi-k2.5 die 2026-07-31 — none in our current config, which is clean today.)

## Scope (extend scripts/model_health_check.py + weekly_model_check.sh; keep the plan/apply split and unit-tested pure functions)
1. **Advance retirement scan**: fetch https://docs.ollama.com/cloud.md (plain-markdown mirror of the cloud docs; index at docs.ollama.com/llms.txt) and parse the retirement table (Date | Model | Alternative). For any CONFIGURED model with a future retirement date: alert immediately with the date + suggested replacement, and auto-append the mapping to .litellm/model_upgrades.yaml via the existing PR path so the swap is pre-approved before the 410 ever happens.
2. **New-model discovery**: GET https://ollama.com/api/tags with Authorization: Bearer $OLLAMA_CLOUD_API_KEY (source of truth for the live catalog); diff against a committed snapshot (.litellm/ollama_catalog_snapshot.json). New tags → open ONE agent:ready evaluation issue listing them (with capabilities pulled from the tags payload) titled 'Evaluate new Ollama Cloud models: <names>'; update the snapshot in the same PR. Removed tags that we don't use → snapshot update only.
3. Alerting stays on the existing email path; scrape failures degrade gracefully (log + skip — the 410 probe remains the backstop).
4. Keep it deploy-flow-correct: all repo-visible changes (upgrade map, snapshot) land via the existing open-PR mechanism, never as silent box edits.

## Acceptance
- Unit tests for the markdown-table parser (real cloud.md fixture), snapshot diff, and mapping-append logic; scrape/API calls mocked.
- A dry-run mode proving: future-dated retirement of a configured model → alert + map PR; new tag → evaluation issue body rendered.

## Addendum (owner requirement, 2026-07-27): major-version upgrade tracking
Retirement-reactive is not enough — e.g. minimax-m3 has been available while our config sat on
minimax-m2.7, and nothing flagged it. Add a **family-version scan** to the weekly check:
- Parse configured Ollama model names into (family, version) — e.g. `minimax-m2.7` → (minimax, 2.7),
  `qwen3.5:397b` → (qwen, 3.5), `glm-5.1` → (glm, 5.1), `gpt-oss:20b` → (gpt-oss, size-tagged/unversioned).
- From the /api/tags catalog, find the newest version per configured family. When the configured
  version trails the newest **major** version, auto-file (or append to) ONE consolidated
  `agent:ready` issue: "Ollama model family upgrades available: <family old→new, ...>" including
  each candidate's usage level vs the current model's — a version bump that jumps usage tiers
  (m2.7 is Medium, m3 is High = more quota burn) is an evaluation, never an auto-swap.
- Minor/patch bumps within the same major can be listed in the same issue at lower urgency.
- Dedup: don't re-file while an open upgrade issue already names that family+version (title-marker
  match, not label match — see #598).
- Unit tests for the version parser across the real current catalog names (minimax-m2.7/m3,
  qwen3.5, glm-5.1/5.2, kimi-k2.6/k2.7-code, gemma4, deepseek-v4-flash/pro, nemotron-3-nano/super/ultra).

gitchrisqueen:
## What & why

The weekly model-health cron (`scripts/weekly_model_check.sh` + `model_health_check.py`, Sun 09:00 UTC) was purely **reactive**: it found a retirement only once the model already returned 410, and it never noticed that a newer model in a family we run had shipped — `minimax-m3` has been available while `.litellm/config.yaml` sat on `minimax-m2.7`. Ollama publishes both facts in advance, so the check now reads them.

Three new scans in `scripts/model_health_check.py`, all pure + unit-tested, all degrading to a skip when a source is unreachable (the 410 probe remains the backstop):

### 1. Advance retirement scan
Parses `docs.ollama.com/cloud.md`'s retirement tables — the "Upcoming" rows carry their own date, the "Past" ones inherit their `<Accordion title="…">` date. For a **configured** model with a **future** date: alert now with the date + published alternative, and pre-append that mapping to `.litellm/model_upgrades.yaml` so the existing swap path applies it *before* the sunset.

Load-bearing choices:
- A date already **past** produces nothing — that's a live 410 the probe owns, and re-reporting it weekly would bury the actionable one.
- A retirement with **no published alternative** alerts but writes **nothing**. Writing `REMOVE` would make the check label its own unhandled retirement "known" and stop asking.
- An alternative that isn't in the live catalog is never auto-mapped.

### 2. New-model discovery
Diffs `ollama.com/api/tags` (Bearer `$OLLAMA_CLOUD_API_KEY` when set) against the committed `.litellm/ollama_catalog_snapshot.json`. New tags → ONE `agent:ready` evaluation issue; the snapshot is refreshed in the same PR. Removed tags we don't use → snapshot only. A **missing** snapshot is SEEDED, never reported as a catalog of 18 new tags.

### 3. Family-version scan (owner addendum)
`parse_model_version` handles the three spellings the real catalog uses — fused (`qwen3.5:397b`), own dash-part (`glm-5.1`, `nemotron-3-nano:30b`, `mistral-large-3:675b`) and one-letter marker (`minimax-m2.7`, `deepseek-v4-flash`). A variant suffix stays **in the family key** (`kimi-k2.7-code` → `kimi-code`), so a coding variant is never offered as an upgrade for plain `kimi-k2.6`. A size tag is not a version, so `gpt-oss:20b` comes back unversioned and stays out of the comparison.

When a configured model trails its family, ONE consolidated `agent:ready` issue is filed, major bumps first and minor/patch below. Each side's **usage level** is scraped from its `ollama.com/library/<name>` page: `minimax-m2.7` is *medium*, `minimax-m3` is *high*, and a tier jump means more quota burn — so this is always an **evaluation**, never an auto-swap, and the body says so. An unread usage page is `None`, never "no jump". A candidate that is itself scheduled to retire is never recommended.

### Dedup (#598 lesson)
Literal marker match across the **title, body AND comments** of open issues carrying the matching title prefix — comments count because that is where an append lands. Issues are **listed and filtered client-side** rather than handed to `--search` (the search index tokenizes and lags; a miss there is a duplicate every week), and the lookup **fails closed**: an unreadable listing skips the filing instead of re-filing.

### Orchestrator
`weekly_model_check.sh` runs the scan after the probe, emails the retirement notices on the existing path, opens the map/snapshot changes as a PR, and files the issues. `open_pr` is now parameterised over branch/title/body/mutation (the swap PR is unchanged in behaviour beyond a force fallback on a stale bot branch). It also `git pull --ff-only`s the cron clone first — the map the run READS is the clone's copy, so without that a mapping pre-approved and merged last week would never reach the run that would apply it.

## Current state, run live against this branch

```
$ python3 scripts/model_health_check.py --catalog-scan --today 2026-07-27
Ollama catalog scan (2026-07-27): docs=ok catalog=ok tags=18
  UPGRADE (major) minimax-m2.7 -> minimax-m3 — usage medium (level 2) -> high (level 3)
```

Exactly the gap the addendum named. No configured model is scheduled to retire and the committed snapshot matches the live catalog, so nothing spurious would be filed.

## Testing

- `113 passed` in `tests/unit/test_model_health_check.py` (was 34); full suite `7597 passed`.
- Fixtures are **real captures** committed under `tests/unit/fixtures/ollama/`: `cloud.md`, `tags.json`, and three model pages (`minimax-m2.7` medium / `minimax-m3` high / `deepseek-v4-pro` extra high).
- Covered: the markdown-table parser (upcoming + accordion dates, empty alternatives, header/separator rows, tables *outside* the Retirements section), notice/mapping planning, comment-preserving + idempotent map append, snapshot parse/render/diff, the version parser across every name the addendum lists, family-upgrade planning, usage-level parsing, issue rendering + title truncation, dedup planning, the `gh` client (fail-closed), and every fetch's degradation path.
- **Dry-run proof, offline:** `--catalog-scan --catalog-fixture tests/unit/fixtures/ollama` drives the whole CLI with the network replaced. `TestCatalogScanCLI` asserts a future-dated retirement of a configured model → alert + map addition, a new tag → rendered evaluation issue, `--catalog-apply` writes and is idempotent, and dry-run writes nothing.

No `src/` code, no migration, no secrets. `scripts/` stays stdlib-only (urllib, not `requests`) — it runs from a host ops cron with no app venv guaranteed.

Closes #716

### DIFF
--- a/.litellm/ollama_catalog_snapshot.json
+++ b/.litellm/ollama_catalog_snapshot.json
@@ -1,0 +1,114 @@
+{
+  "_comment": "Committed snapshot of ollama.com/api/tags, refreshed by scripts/model_health_check.py --catalog-apply. New tags here vs the live catalog become ONE agent:ready evaluation issue; do not hand-edit.",
+  "updated": "2026-07-27",
+  "models": {
+    "deepseek-v4-flash": {
+      "modified_at": "2026-04-24T00:00:00Z",
+      "size": 140000000000,
+      "parameter_size": "",
+      "family": ""
+    },
+    "deepseek-v4-pro": {
+      "modified_at": "2026-04-24T00:00:00Z",
+      "size": 1600000000000,
+      "parameter_size": "",
+      "family": ""
+    },
+    "gemma4:31b": {
+      "modified_at": "2026-04-02T09:00:00-08:00",
+      "size": 62546177752,
+      "parameter_size": "",
+      "family": ""
+    },
+    "glm-5.1": {
+      "modified_at": "2026-04-07T08:00:00-07:00",
+      "size": 1507728316928,
+      "parameter_size": "",
+      "family": ""
+    },
+    "glm-5.2": {
+      "modified_at": "2026-06-16T08:00:00-07:00",
+      "size": 0,
+      "parameter_size": "",
+      "family": ""
+    },
+    "gpt-oss:120b": {
+      "modified_at": "2025-08-05T00:00:00Z",
+      "size": 65290180781,
+      "parameter_size": "",
+      "family": ""
+    },
+    "gpt-oss:20b": {
+      "modified_at": "2025-08-05T00:00:00Z",
+      "size": 13780162412,
+      "parameter_size": "",
+      "family": ""
+    },
+    "kimi-k2.5": {
+      "modified_at": "2026-01-26T00:00:00Z",
+      "size": 1118481408000,
+      "parameter_size": "",
+      "family": ""
+    },
+    "kimi-k2.6": {
+      "modified_at": "2026-03-31T00:00:00Z",
+      "size": 595148192736,
+      "parameter_size": "",
+      "family": ""
+    },
+    "kimi-k2.7-code": {
+      "modified_at": "2026-06-12T00:00:00Z",
+      "size": 595148192736,
+      "parameter_size": "",
+      "family": ""
+    },
+    "minimax-m2.5": {
+      "modified_at": "2026-02-12T00:00:00Z",
+      "size": 230000000000,
+      "parameter_size": "",
+      "family": ""
+    },
+    "minimax-m2.7": {
+      "modified_at": "2026-03-18T00:00:00Z",
+      "size": 480836588544,
+      "parameter_size": "",
+      "family": ""
+    },
+    "minimax-m3": {
+      "modified_at": "2026-06-01T00:00:00Z",
+      "size": 0,
+      "parameter_size": "",
+      "family": ""
+    },
+    "mistral-large-3:675b": {
+      "modified_at": "2025-12-02T00:00:00Z",
+      "size": 682000000000,
+      "parameter_size": "",
+      "family": ""
+    },
+    "nemotron-3-nano:30b": {
+      "modified_at": "2025-12-15T00:00:00Z",
+      "size": 32645090390,
+      "parameter_size": "",
+      "family": ""
+    },
+    "nemotron-3-super": {
+      "modified_at": "2026-03-11T00:00:00Z",
+      "size": 230500000000,
+      "parameter_size": "",
+      "family": ""
+    },
+    "nemotron-3-ultra": {
+      "modified_at": "2026-06-04T00:00:00Z",
+      "size": 0,
+      "parameter_size": "",
+      "family": ""
+    },
+    "qwen3.5:397b": {
+      "modified_at": "2026-02-16T00:00:00Z",
+      "size": 397000000000,
+      "parameter_size": "",
+      "family": ""
+    }
+  }
+}
--- a/scripts/model_health_check.py
+++ b/scripts/model_health_check.py
@@ -35,6 +35,9 @@ CLI:
   --catalog-apply        Write the planned map additions + snapshot refresh to --map / --snapshot.
   --file-issues          File (or append to) the GitHub issues the catalog scan planned.
 Options:
+  --plan-file PATH       Act on a plan already emitted by --catalog-json ("-" = stdin) instead of
+                         re-scanning. Use it with --file-issues so the filing acts on the SAME
+                         reading the decision to file was made from.
   --config PATH          LiteLLM config to read (default .litellm/config.yaml).
   --map PATH             Upgrade map (default .litellm/model_upgrades.yaml).
   --snapshot PATH        Committed catalog snapshot (default .litellm/ollama_catalog_snapshot.json).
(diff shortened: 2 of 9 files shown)


gitchrisqueen:

- **Remote doc content injected extra entries into `model_upgrades.yaml`** (`scripts/model_health_check.py`): `parse_retirements` took the "Recommended alternative" cell from `docs.ollama.com/cloud.md` verbatim, and `append_upgrade_mappings` writes it as a `"key": "value"` YAML line — so a cell containing a quote/newline injected a second, attacker-chosen mapping into the file the reactive half applies to the live LiteLLM config. Reproduced locally before the fix. Added `is_model_tag()` and enforced it in BOTH places: a non-tag model row is dropped at parse, a non-tag alternative degrades to "no published alternative", and the writer refuses any pair it did not check.
- **A transient outage could lose the evaluation issue permanently** (`scripts/weekly_model_check.sh`): the orchestrator decided to file from the `--catalog-json` plan, then ran `--file-issues`, which re-scanned `cloud.md` + `/api/tags` from scratch. If that second scan failed to fetch, it filed nothing and said nothing — and since the snapshot-refresh PR opened moments earlier makes those tags no longer "new", the issue would never be filed at all rather than retried next week. `--file-issues` now accepts `--plan-file PATH|-` and acts on the same reading; `--plan-file` + `--catalog-apply` is refused loudly (the emitted plan strips `_catalog`, so the snapshot write would silently no-op). Same block: `|| log "issue filing failed (non-fatal)"` fired on every SUCCESSFUL run, because 2/3 are the planner's own actions/alerts exit codes — the cron log claimed failure exactly when filing worked. Now checks the real status.

Verified clean (no change needed): the version parser across every live catalog name (`kimi-k2.7-code` → family `kimi-code`, `gpt-oss:20b` unversioned, `deepseek-v4-flash` → `deepseek-pro`/`deepseek-flash` split); alternatives in the fixture match `/api/tags` keys exactly, so the `alternative_in_catalog` gate is not silently off; a same-version-different-size sibling is never proposed; `gh issue create` labels (`agent:ready`/`priority:medium`/`infrastructure`) all exist in the repo, and the `--limit 200` dedup listing comfortably covers the 13 open issues; the `gh`/email paths pass remote strings as argv, not through a shell.

Tests: 119 in `tests/unit/test_model_health_check.py` (was 113 — added the injection cases, the tag-shape parser, and three `--plan-file` CLI cases that fail if the catalog is re-scanned). Full unit suite `7603 passed`.

gitchrisqueen:

Issue #723, opened 2026-07-27
ops(feedback): regrant FEEDBACK_GITHUB_TOKEN with Issues: Read and write (root cause of #718)


gitchrisqueen:
Split out of #718 — the code half of that issue is fixed in PR (error escalation + a repair sweep that re-attaches dropped labels). **This is the root cause, and only you can do it**: an agent may never touch secrets or `/opt/lem`.

## What is broken
`POST /repos/.../issues` succeeds, but `POST /issues/{n}/labels`, `/assignees` and `/comments` all return
`403 Resource not accessible by personal access token`. Auto-filed feedback issues therefore land with no
`agent:ready`/`needs-human`, no `priority:*`, no `feedback-loop`, no assignee and no Decision Comment — the
agent pipeline selects on labels, so it never sees them (#596, #638/#600, #713 were all hand-labeled).

## What to do
1. Reissue/regrant `FEEDBACK_GITHUB_TOKEN` as a fine-grained PAT with **Issues: Read and write** (+ **Metadata: Read**)
   on `christopherqueenconsulting/linkedin_engagement_manager`.
2. Update `/opt/lem/.env` and restart the celery workers.
3. Verify on the next `file-feedback-issues` beat (every 30 min): a newly filed issue should arrive with its
   full label set and its tracking/Decision Comment.

## What happens once you do
The repair sweep shipped for #718 runs on that same beat, so **the already-broken backlog re-labels itself** —
no hand-labeling needed. Until then, every failure is logged at ERROR (so it reaches PostHog) naming this
exact grant, instead of the old silent WARNING.

Related: #598 (the separate-calls fix that this token cannot execute), #590 (the token-narrowing window this
may have regressed in).

gitchrisqueen:
Done — owner regranted the PAT on 2026-07-27 and it is **verified working in place**: a live `POST /repos/.../issues/{n}/labels` call with the token in `/opt/lem/.env` returned **200** (previously 403), and the running celery workers carry the current value. The code half shipped in #718. Auto-filed feedback issues can now label themselves.

Issue #713, opened 2026-07-27
feat(content-generation): Add rejection reason field and prompt to regenerate rejected posts


gitchrisqueen:
## Why
When a post is rejected, the user should be able to add a reason why it was rejected, be prompted with an option to regenerate the post, and the regeneration should consider the rejection reason to avoid the same issue.

Reported by 1 distinct user(s) across 1 feedback item(s).

Auto-filed from in-app feedback #4 by the feedback→auto-work loop (`docs/launch-and-marketing-plan.md` §B.2/B.3). Classifier: feature/medium, risk `none`, confidence 0.90.

## Scope
- Address the reported feature in the `content-generation` area of the product.
- Keep the change minimal and scoped to this report — no unrelated refactors.

## Files
Likely starting points (verify before editing; add/extend the unit tests):
- `src/cqc_lem/app/run_content_plan.py`
- `src/cqc_lem/utilities/ai/ai_helper.py`
- `src/cqc_lem/utilities/ai/content_framework.py`
- `tests/unit/`

## Acceptance
- The behavior the reporter described is fixed/implemented.
- Unit tests cover the new/changed logic (≥80% patch coverage).
- All required CI gates pass.
- A human has answered the Decision Comment before this is merged.

Issue #638, opened 2026-07-26
fix(other): Fix input fields blocking comma and space entry


gitchrisqueen:
## Why
Many settings asking for comma separated values don't allow the user to enter commas or spaces. Something is blocking when trying to type in the fields.

Reported by 1 distinct user(s) across 1 feedback item(s).

Auto-filed from in-app feedback #3 by the feedback→auto-work loop (`docs/launch-and-marketing-plan.md` §B.2/B.3). Classifier: bug/high, risk `none`, confidence 0.90.

## Scope
- Address the reported bug in the `other` area of the product.
- Keep the change minimal and scoped to this report — no unrelated refactors.

## Files
Likely starting points (verify before editing; add/extend the unit tests):
- `tests/unit/`

## Acceptance
- The behavior the reporter described is fixed/implemented.
- Unit tests cover the new/changed logic (≥80% patch coverage).
- All required CI gates pass.

PR #724, opened 2026-07-27
fix(feedback): repair auto-filed issues whose labels never landed (closes #718)

### LINKED ISSUE #718: fix(feedback): FEEDBACK_GITHUB_TOKEN 403s on issue labels/assignees/comments — #598 fix runs but auto-filed issues still land label-less
## Symptom
The #598 fix (attach labels/assignees in SEPARATE calls after issue creation) is deployed and executing in prod — but every one of those follow-up calls now fails with **403 "Resource not accessible by personal access token"**. Auto-filed feedback issues STILL land label-less (no `agent:ready`/`needs-human`, no `priority:*`, no `feedback-loop`), unassigned, and without the tracking comment — invisible to the agent pipeline until a human hand-labels them.

## Evidence (prod, 2026-07-27 13:00 UTC beat run)
Feedback row id 4 (widget, user 1, created 12:34 UTC — "Add rejection reason field + regenerate rejected posts") was classified and auto-filed as issue #713 at 13:00:15 UTC. `celery_worker` log / PostHog ERROR entries at 13:00:15–16 UTC:

```
ERROR: GitHub POST issues/713/labels returned 403: {"message":"Resource not accessible by personal access token", ...}
WARNING: Auto-filed issue #713 could not be labeled ['feature', 'priority:medium', 'feedback-loop', 'needs-human']
ERROR: GitHub POST issues/713/assignees returned 403: {"message":"Resource not accessible by personal access token", ...}
WARNING: Auto-filed issue #713 could not be assigned to ['gitchrisqueen']
ERROR: GitHub POST issues/713/comments returned 403: {"message":"Resource not accessible by personal access token", ...}
```

`POST /issues` (creation) succeeds; `POST /issues/{n}/labels`, `/assignees`, and `/comments` all 403. #713 sat label-less until hand-labeled at 13:51 UTC — the third hand-fix (#596, #638/#600 era, now #713).

## Root cause (likely)
`FEEDBACK_GITHUB_TOKEN` in `/opt/lem/.env` is a fine-grained PAT whose permission set allows issue creation but not the label/assignee/comment sub-resource endpoints. Fine-grained PATs need **Issues: Read and write** on this repo for all four endpoints — the current token's grants (or its repo access selection) don't cover the sub-resource calls. Note the recent token narrowing work (#590 / GHCR_PAT scope reduction) as a possible regression window.

## Fix
1. Reissue/regrant `FEEDBACK_GITHUB_TOKEN` as a fine-grained PAT with **Issues: Read and write** (+ Metadata: Read) on `christopherqueenconsulting/linkedin_engagement_manager`, update `/opt/lem/.env`, and restart the celery workers.
2. Verify on the next `file-feedback-issues` run (every 30 min) that a filed issue arrives with its full label set and the tracking comment.
3. Optional hardening: escalate the "could not be labeled" path from WARNING to ERROR (it currently reads as a soft warning while the loop is effectively broken), or have the beat task retry label attachment on subsequent passes for `issue_created` rows whose issues are label-less.

## Impact
Until fixed, every auto-filed feedback issue requires manual labeling before the autonomous pipeline can see it — the feedback→auto-work loop is only half-functional.

gitchrisqueen:
## What & why

`POST /issues` kept succeeding while `POST /issues/{n}/labels`, `/assignees` and `/comments` all 403'd with
`Resource not accessible by personal access token`. Every auto-filed feedback issue therefore landed with no
`agent:ready`, no `priority:*`, no `feedback-loop`, no owner and no Decision Comment — and the agent pipeline
selects on labels, so it could not see them at all. #596, #638/#600 and #713 each needed a human to hand-label
them before any work could start.

**Scope note:** #718's fix list has three items. Items 1–2 (reissue `FEEDBACK_GITHUB_TOKEN` with
`Issues: Read and write`, update `/opt/lem/.env`, restart workers, verify) are secret/prod-host actions the
RUNBOOK forbids an agent from doing — they are filed as **#723**, assigned to @gitchrisqueen. This PR is item 3,
the code hardening, and it is what makes item 1 self-healing rather than a hand-labeling job.

### 1. A dropped label is an ERROR, not a warning
`log_warning` → `log_error` on both attach paths. Only ERROR reaches PostHog at the default
`POSTHOG_LOG_LEVEL`, so the old warning meant a fully broken loop produced no issue, no alert and no signal —
it read as a soft degradation for weeks. Every message now names the grant that fixes it
(`TOKEN_PERMISSION_HINT`). A held issue whose **Decision Comment** failed to post is an ERROR too: the owner is
being asked to answer something that was never written, and that return value was previously discarded.

### 2. `repair_auto_filed_issues` — the backlog self-heals
Runs on every `file-feedback-issues` beat (every 30 min), so the moment the token is fixed the already-broken
issues get their labels back with no human in the loop.

Nothing in the DB records the intended label set, and recording it would need a schema change — but
`build_issue_body` already writes the classifier verdict into the issue body — the `Classifier:
<category>/<severity>, risk ..., confidence ...` line plus the demand line. `parse_filed_issue` reads that back and
`labels_for_issue` recomputes the exact same set, deterministically, with **no second LLM call**.

Guard rails, because this writes to issues a human may have touched:
- Only ADDS. Never rewrites title/body, never removes a label.
- Only issues carrying our provenance marker (`Auto-filed from in-app feedback`) **and missing the
  `feedback-loop` label**. That label rides in the same call as every other one, so its presence proves the
  attach had permission — anything else on such an issue is a human's curation, not our loss.
- Closed issues are skipped; an unreadable classifier line is skipped and logged rather than guessed at.
- A missing demand line reads as **0** reporters, never 1 — floored up, a first-of-its-kind feature would
  repair itself into `agent:ready` on demand it never had.
- Assignee is only re-attached when there is none; the Decision Comment only when the comment list was
  readable AND holds none (a failed read is never "there isn't one").
- One GET per open cluster, nothing written for a healthy issue, and the sweep **stops at the first refusal** —
  a permission failure is deterministic, so 50 more calls would only bury the one error that says why.

## Testing

- 43 new unit tests: the label set round-trips through the filed body for every risk shape and both feature-demand
  states, every repair decision (label-less / already-labeled / partially hand-labeled / not ours / closed /
  unreadable / held-needs-assignee+comment / already-commented / refused), and the sweep's token gate, issue-number
  dedup, stop-on-refusal and continue-past-an-unreadable-issue behaviour.
- `poetry run pytest tests/unit -q` → **7638 passed**.

Closes #718

### DIFF
--- a/src/cqc_lem/app/run_scheduler.py
+++ b/src/cqc_lem/app/run_scheduler.py
@@ -1705,11 +1705,19 @@ def auto_file_feedback_issues(self, limit: int = 25):
     """Drain the captured-feedback queue into the work pipeline (issue #498, plan §B.2/B.3): classify
     each new report, dedup it against the open clusters (+1 the existing issue) and open ONE
     `MODE=start`-shaped issue per genuinely new problem. QueueOnce so two beat ticks can never file
-    the same report twice."""
-    from cqc_lem.utilities.feedback.issue_service import process_new_feedback
+    the same report twice.
+
+    The same pass repairs already-filed issues whose labels/assignee/Decision Comment never landed
+    (issue #718) — an unlabeled issue is invisible to the agent pipeline, so filing alone is not the
+    job being done."""
+    from cqc_lem.utilities.feedback.issue_service import (
+        process_new_feedback, repair_auto_filed_issues)
 
     result = process_new_feedback(limit=limit)
-    return f"Feedback filing: {result['processed']} row(s) — {result['counts'] or 'nothing to do'}"
+    repair = repair_auto_filed_issues()
+    return (f"Feedback filing: {result['processed']} row(s) — "
+            f"{result['counts'] or 'nothing to do'}; repair: {repair['scanned']} checked, "
+            f"{repair['repaired']} repaired, {repair['failed']} still broken")
 
 
 @shared_task.task(bind=True, base=QueueOnce, once={'graceful': True})
--- a/src/cqc_lem/utilities/feedback/issue_service.py
+++ b/src/cqc_lem/utilities/feedback/issue_service.py
@@ -11,6 +11,10 @@ category/severity/risk verdict. This stage decides what the world sees:
                                               ├─ similar to an OPEN cluster → +1 ITS issue
                                               └─ new problem → open ONE issue, stamp it back
 
+A filed issue is only useful if its LABELS land — the agent pipeline selects on them — so every
+filing beat also runs `repair_auto_filed_issues`, which re-attaches the labels/assignee/Decision
+Comment on any already-filed issue whose post-create calls were refused (issue #718).
+
 Dedup is the whole point: one recurring problem must be one issue, so the Nth report bumps demand on
 the existing issue instead of spamming the backlog. Similarity is embedding cosine when an embedding
 is available and falls back to `content_framework.text_similarity` (the deterministic token-overlap
@@ -40,6 +44,7 @@ Env:
 
 import json
 import os
+import re
 from typing import Optional
 
 from cqc_lem.utilities.ai.content_framework import as_vector, cosine_similarity, text_similarity
@@ -49,7 +54,7 @@ from cqc_lem.utilities.db import (
 )
 from cqc_lem.utilities.feedback.classifier import (
     MAX_BODY_CHARS, NEEDS_HUMAN_LABEL, RISK_LABELS, FeedbackCategory, FeedbackClassification,
-    FeedbackRisk, FeedbackRoute, classify_feedback, labels_for,
+    FeedbackRisk, FeedbackRoute, FeedbackSeverity, classify_feedback, labels_for,
 )
 from cqc_lem.utilities.logger import log_debug, log_error, log_info, log_warning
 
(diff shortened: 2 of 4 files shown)


gitchrisqueen:

- **Repair pass parsed the reporter-supplied half of the body**: `parse_filed_issue` used `search()` for both the `Reported by N distinct user(s)` and `Classifier: …` lines, which takes the FIRST match anywhere. Both live *below* `## Why`, whose summary is model-written from the reporter's own words — so a report summarized as "reported by 9 distinct users on my team" would have made the repair recompute its own demand, and demand is exactly what promotes a FEATURE to `agent:ready` (i.e. built unattended by the pipeline). Same shape for a summary quoting a `Classifier: bug/critical … confidence 0.99` line. Fixed: the provenance line is now searched from `FILED_PROVENANCE_MARKER` onward, and the demand line is the LAST match *in front of* that marker — which is the one `build_issue_body` actually wrote. Two tests added; both fail on the pre-fix code.
- **The sweep stopped on a refused write but not a refused read**: with a token that also can't `GET /issues/{n}` (the same permission family #718 is about), the pass burned 50 failed GETs and 50 ERROR logs every 30-minute beat — the exact "bury the one error that says why" the write path guards against. Fixed with `REPAIR_READ_FAILURE_STOP` (3) consecutive read failures; a single unreadable issue still doesn't stop the sweep, since that's just a deleted issue.

Checked and found clean: `labels_for_issue` round-trips exactly (`route`/`labels` are derived properties off category+confidence, both of which survive the body); `POST /issues/{n}/labels` adds rather than replaces, and `missing` can never be empty since `feedback-loop` is the entry condition; `get_open_feedback_clusters(limit)` signature matches; the new `log_error` calls pass no `exc=`, so they reach PostHog Logs without creating `$exception` issues (no feedback loop into `posthog_error_issues.py`); `faq_service`/`capacity_alerts` issues carry no provenance marker, so the repair never touches them; the Decision Comment can't be double-posted (an unreadable comment list skips rather than reposts).

`poetry run pytest tests/unit -q` → **7641 passed**.

gitchrisqueen:

Issue #619, opened 2026-07-26
G4 — Build-receipt & bookmarkable-resource post archetypes (save-optimized)


gitchrisqueen:
## Context
The 'build receipt' (what I built → tools used → what broke → concrete use cases → soft artifact CTA) is the AI-consultant niche's best measured format: Ben van Sprundel's 20-agent-team receipt hit ~8% engagement on a 20K-follower account. Bookmarkable mega-resource posts are Charlie Hills' signature ('content worth saving, not scrolling past'); saves weigh ~5x a like and feed 2026 evergreen redistribution. The owner ships real AI/automation systems weekly (LEM itself) that are never posted as build logs. Current posts have 0 saves ever.

## Scope
- Two new archetypes in content_framework.py shared menus (NOT parallel helpers): 'build_receipt' and 'resource_compendium', with hook menus leading with a real number and structures per above; mark them save-targeted so scheduling can prefer them.
- Ground both archetypes in story-bank entries / user-supplied project facts (depends on G5; until G5 lands, require approval-gated drafts with placeholders clearly marked, never fabricated specifics — #416 no-fabrication policy).
- Wire archetypes into carousel generation too (a build receipt renders naturally as a document post — 1.45x multiplier, 6.6% ER format).
- Rotation: shape-history (V51 posts.archetype) already covers avoidance; ensure new archetypes participate.

## Acceptance
- Unit tests for archetype selection, hook constraints (number-led, mobile ~140-char budget), and no-fabrication guard (drafts with unverifiable specifics and no story-bank anchor are blocked).

PR #733, opened 2026-07-27
fix(content): one story anchor per deck + a reference-value gate on carousels (closes #728)

### LINKED ISSUE #728: Carousels/documents promise 'save-worthy' but deliver nothing worth saving — one anchor per post, real reference value per slide
## Context
Found in a live working session (2026-07-27) while publishing the first native document post (#644). The generated deck — 'The 160-Release Build Receipt', 5 slides — exposed two content-quality defects that make the save-targeted archetypes (#619/G4) a false promise. Saves are weighted ~5x a like and drive 2026 evergreen redistribution, so a deck that ISN'T save-worthy is not just weak, it wastes the format's whole advantage.

### Defect 1 — the deck's slides carry no reference value
The slides restated claims and narrative ('Release count is not the goal', 'You do not need Kubernetes') with nothing a reader would return to: no checklist, no commands, no config snippet, no numbers table, no decision rule, no before/after. The COPY says 'here is the exact stack' but the slides never show a stack. If a post's theme is save-worthy/bookmarkable, each slide must carry something reusable — the reader should be able to act from the deck alone with the post text deleted.

### Defect 2 — too many story-bank anchors in one post
Generation pulled SIX story-bank entries into a single post (160 releases, blue/green, VARCHAR rollback, 429 doom loop, 369 PRs, cost routing). It reads as a greatest-hits reel, spends the account's best receipts at once, and ran 365 words. #620's design intends ONE anchor per post (`select_story` picks one); the multi-entry path is either a carousel/document-specific leak or the fact-anchor list being handed to the writer as usable material rather than as a verification allow-list. `_fact_anchors` deliberately returns EVERY active entry for the fabrication CHECK — the writer prompt must not receive that same list as content.

## Scope
- **One anchor per post**: audit `create_carousel_content` / `create_text_post` / `_fact_anchors` vs `_fact_anchors_for` and `blueprint_directive` — the writer gets ONE selected story; the verification gate keeps the full list. Add a regression test asserting a generated post references at most one story-bank entry's specifics.
- **Reference value per slide** for save-targeted archetypes (build_receipt, resource_compendium and any deck whose copy claims save-worthiness): each slide must carry at least one concrete reusable artifact — a step, command, metric, threshold, config line, checklist item, or decision rule. Encode as a deterministic review gate (like the slop-linter): a deck whose slides are all narrative/claims fails and regenerates with a directive naming what is missing.
- **Copy/deck consistency**: if the post text promises a list/framework/stack, the deck must contain it. Reject decks whose promise isn't fulfilled by the slides (cheap heuristic: promise keywords in body vs artifact presence in slides).
- Consider a slide-template family for reference content (steps, checklist, comparison table, before/after, command block) so the generator has shapes that ARE inherently save-worthy.

## Acceptance
- Unit tests: one-anchor invariant; save-worthiness gate passes a reference deck and fails a narrative-only one; promise/deck consistency check.
- A regenerated sample deck demonstrably carries reusable content per slide (attach the rendered slides to the PR).

gitchrisqueen:
## What & why

The first live native document post ("The 160-Release Build Receipt", 5 slides) exposed two defects
that make the #619 save-targeted archetypes a **false promise**. Saves carry ~5x a like into 2026
evergreen redistribution, so a deck that isn't save-worthy doesn't merely under-perform — it wastes
the whole advantage of the format.

### Defect 2 — one anchor per post

`create_carousel_content` passed `_fact_anchors(user_id)` — **every active story-bank entry** —
straight into the writer prompt via `carousel_blueprint_directive`. `_fact_anchors` deliberately
returns the whole bank for the fabrication CHECK; handing that same list to the writer is what
produced a greatest-hits reel (160 releases, blue/green, VARCHAR rollback, 429 doom loop, 369 PRs,
cost routing) that spent the account's best receipts at once.

Carousels now run the **same two-width split text posts have always had**:

| | writer | checkers |
|---|---|---|
| before | whole bank | whole bank |
| after | the ONE selected entry (`_select_story_for_post` → `fact_sources`) | whole bank (`_report_carousel_fact_grounding`) |

The deck also finally gets the **story itself** (`story_directive`) — it previously received an
allow-list of numbers with no material behind them. And whether a fact-anchored archetype is on the
carousel menu at all now follows the WRITER's anchor rather than the bank's size, so a deck can
never be handed an archetype it has nothing to write it from (its `[[…]]` placeholders would be
baked into slide images that no re-score can edit).

### Defect 1 — reference value per slide

`save_worthy_directive` has always ASKED for a reusable reference; nothing ever checked one came
back. `content_framework.deck_reference_report` is the deterministic counterpart, built exactly like
the slop lint — pure regex/string work, no LLM, no I/O, one plain-English reason per failure, fed
into a bounded regeneration:

- **Per-slide rule** — every BODY slide of a `save_targeted` archetype (build receipt, bookmarkable
  compendium), or of ANY deck whose caption promises a checklist / the exact stack / a framework /
  the numbers, must carry ≥1 reusable artifact: `step`, `command`, `metric`, `threshold`, `config`,
  `checklist`, `decision_rule`, `comparison`. Cover, CTA and testimonial slides are exempt by
  design — a promise and a save ask are not references, and `carousel_creator` already treats those
  as their own layout family.
- **Promise/deck consistency** — the caption's promise keywords are mapped to the artifact kinds
  that would fulfil them; a promise no slide delivers fails the deck on its own, even when every
  slide passed the per-slide rule.
- **Regeneration** — a failure re-drafts once with the exact failing slides and the broken promise
  named (`deck_retry_directive`), bounded by `DECK_REFERENCE_MAX_ATTEMPTS` (default 2 = one retry).
  A deck that still fails ships with the reason logged: rendered slide images have no review queue
  behind them, and a narrative deck beats no post.
- **Slide-template family** — `reference_slide_directive` gives the writer the shapes that ARE
  inherently save-worthy (numbered step, checklist, command/config block, numbers table,
  before/after, decision rule) up front, so "carry a reusable artifact" is a form to copy rather
  than an abstract ask.

Both gates **fail open**: `DECK_REFERENCE_ENABLED=false` restores the exact prior behaviour, and a
deck with no gradeable slide is never failed.

## Worked example (acceptance evidence)

`docs/examples/deck-reference-728/` carries the rejected and accepted decks rendered through the
**unchanged** `create_carousel_slide_images` pipeline — what this PR changes is the TEXT that lands
on the slides. Both graded against the caption *"Here is the exact stack behind it, end to end."*

**Before** (`before/`, the deck that shipped) — `passes=False`

```
empty_slides = ['Release count is not the goal', 'You do not need Kubernetes', 'Automation compounds']
unfulfilled  = ['the exact stack, commands or config']
```

**After** (`after/`) — `passes=True`, every body slide carries an artifact:

| Slide | Artifacts |
|---|---|
| 1. Pin the image tag — set IMAGE_TAG to the release tag, never latest | `step`, `config` |
| 2. Migrate first — run flyway migrate before the app flips; if a migration fails, roll back | `step`, `command`, `decision_rule` |
| 3. Health check under 2 seconds — or the deploy aborts | `step`, `threshold`, `metric` |
| What changed — deploys went from 14 minutes of downtime to 0 | `comparison`, `metric` |

(The slides render without a photo band because they were generated offline with Pexels unreachable.)

## Testing

- `tests/unit/utilities/ai/test_deck_reference_gate.py` (46) — every artifact shape, narrative-only
  slides carrying nothing, a tool version not counting as a metric, cover/CTA/testimonial exemption,
  the promise lexicon, a reference deck passing and a narrative one failing, promise-not-fulfilled
  failing on its own, fail-open cases, the retry directive, the env toggles.
- `tests/unit/utilities/ai/test_carousel_reference_regeneration.py` (9) — the generator regenerates
  with the missing slides named, a good deck costs exactly one call, the retry is bounded and the
  last deck still ships, the budget is configurable, an unpromising ordinary deck is left alone, the
  story directive reaches the prompt.
- `tests/unit/app/test_carousel_one_anchor.py` (10) — the one-anchor invariant against a 3-entry
  bank, rotation, empty bank, a failed bank read, the archetype menu following the writer anchor,
  and the deck fact-grounding check knowing every active entry.
- `tests/unit/app/test_save_targeted_wiring.py` — the #619 test that pinned the old whole-bank
  behaviour is rewritten to pin the invariant instead.
- Full suite: `7842 passed`. Coverage on changed modules: `content_framework` 98%,
  `run_content_plan` 97%, `ai_helper` 96%.

## Notes / follow-ups

- A deck that fails the gate after its retry is **logged, not held**. The slides aren't persisted as
  text, so `evaluate_post_gates` (which runs later, from the DB) has nothing to re-read — the same
  posture the slop lint takes for DMs and newsletter editions. The caption itself still goes through
  every existing post gate.
- `_report_carousel_fact_grounding` is advisory for the same reason: an invented slide number is now
  visible in the logs instead of silent, but there is nothing a hold could fix once images render.

Closes #728


gitchrisqueen:

- **Carousel anchor was read but never rotated** (`run_content_plan.create_carousel_content`): the new one-anchor split selects a story-bank entry but never called `record_story_bank_use`, which `create_text_post` has always done. `select_story` orders by least-used/longest-unused, so every deck got the SAME entry and the next text post re-used it — "one anchor per post, same anchor every post". Fixed: the deck records its anchor after a successful generation (guarded, and only when a deck actually came out of it), with tests for the write, the empty-bank no-op and a bookkeeping failure not blocking the carousel.
- **Narrative slides passed the reference gate as "commands"**: `_ARTIFACT_COMMAND_RE` matched its runner list under `re.IGNORECASE`, so `make` — an ordinary English verb — matched "Make sure the pipeline is boring" and "make it simple". Verified against the very deck this issue is about: those slides came back carrying a `command` artifact. Fixed: the runner alternative is now case-sensitive (`(?-i:…)`) and refuses a following stop word; real invocations (`docker compose up`, `make build`, backticked spans, `--flags`) still count. Tests added both ways.
- **The bounded retry could destroy a shippable deck** (`ai_helper.generate_carousel_content`): the loop overwrote `post_text, carousel_dict` with whatever the retry returned. `_call_llm` re-raises, so a provider error on the SECOND call failed the whole carousel task; and an unparseable retry replaced a usable deck with `{}` plus the generic "Explore our latest insights…" caption — grading as `checked=False`, so it shipped silently. Both contradict the gate's stated fail-open posture. Fixed: a retry that raises or comes back with no gradeable slide keeps the previous draft and logs why; tests for each.
- **Caption-promise lexicon over-triggered**: bare `\bprocess\b` made `required=True` for any deck whose caption said "our hiring process", forcing every body slide through the per-slide rule and a second `lem-complex` draft on decks nobody claimed were references — the exact "permanent regeneration tax" the module's own comment warns against. Dropped; `step-by-step` / `playbook` / `framework` / `workflow` / `steps` still promise.

Acceptance criteria re-checked against the diff: the one-anchor invariant, the save-worthiness gate (reference deck passes / narrative fails), the promise-consistency check and the rendered before/after decks are all present and pinned by tests.

Full unit suite on this head: `7855 passed`.

gitchrisqueen:

Issue #483, opened 2026-07-24
feat(leads): inbound-intent detection & hot-lead routing from comments/replies/DMs


gitchrisqueen:
## What
Detect **buying/intent signals** in comments, replies, and DMs on the user's own posts (and on replies to our automated comments) — e.g. "how much", "pricing", "do you offer/help with X", "can you do Y", "DM me", "need this", "who do you recommend" — and **flag + route them as hot leads** for an immediate, personalized response (draft auto-generated, human-approved).

## Why (inbound BD)
This is captured demand we're almost certainly dropping today: someone raising their hand in a comment/DM is the warmest lead possible, and a slow/no response kills it. Turning those signals into a surfaced, actioned queue converts existing engagement into conversations with zero new outreach.

## How
- A lightweight intent classifier (heuristic keyword set + a cheap `lem-simple` LLM check for ambiguous cases) run over new comments/replies/DMs the system already reads (`_reply_to_comments_on_open_post`, comment follow-ups #478, DM inbound).
- On a hit: record a `lead_signal` (person, source, snippet, score, post/thread ref), tag it hot, and generate an approval-gated personalized reply/DM via the existing template + voice machinery.
- Surface hot signals in the UI (a "Leads" inbox) + optional notification to the owner.
- Dedup per (person, thread) so one conversation isn't re-flagged.

## Acceptance
- `detect_lead_signals` over the existing comment/DM read paths; `lead_signals` table (timestamp migration); UI leads inbox; approval-gated draft response. Unit tests ≥90% patch. No new scraping — rides existing reads.

PR #735, opened 2026-07-27
fix(automation): resolve the LinkedIn message thread through a route ladder (closes #731)

### LINKED ISSUE #731: check_dm_replied is dead on live LinkedIn: Message control is now an <a href='/messaging/compose/…'>, not a <button> — follow-ups never stop after a reply
## Confirmed live 2026-07-27 (F2 / #403 working session)

`check_dm_replied` (`run_automation.py` ~L3971) opens a person's thread by locating a Message **button**:
```python
find_first(driver, wait, [(By.CSS_SELECTOR, "button[aria-label^='Message']"),
                          (By.XPATH, "//button[normalize-space()='Message']")], ...)
```
On live LinkedIn today **neither locator matches**, on a pending-invite profile *or* a 1st-degree connection. The function therefore returns `False` on its `msg_btn is None` path — and `False` means *'no reply — keep sending follow-ups.'*

**Impact:** reply detection is effectively disabled. `process_user_followups` (`run_automation.py` ~L4153) is the only caller; it will keep sending scheduled follow-ups to people who ALREADY replied. That is precisely the spammy behaviour the DM sequencing was designed to avoid, and it fails silently (the defensive `return False` is indistinguishable from a genuine no-reply).

## Evidence
E2E run of the production function against real profiles returned `check_dm_replied → False` with an EMPTY thread tail (zero message events read). A DOM sweep of a 1st-degree profile found the Message control is **an anchor, not a button**:
```json
{"tag": "a", "text": "Message", "href": "/messaging/compose/?profileUrn=urn%3Ali%3Afsd_profile%3A<id>"}
```
(No `aria-label`; the visible label lives in nested `span`/`div` nodes. Obfuscated class names — do not key on them.)

## Fix
1. **Locator**: add anchor forms ahead of the legacy button ones, e.g.
   `(By.CSS_SELECTOR, "a[href*='/messaging/compose/']")` and
   `(By.XPATH, "//a[.//text()[normalize-space()='Message']] | //button[normalize-space()='Message']")`,
   keeping the existing button locators last as fallbacks (other surfaces may still render buttons).
2. **Post-click verification**: after opening, assert the thread actually rendered (message events > 0) before judging. If zero events are readable, that is UNKNOWN, not 'no reply'.
3. **Make the unknown case safe**: today unknown → `False` → follow-up sends. Given the caller, unknown should SKIP the follow-up (or defer it) rather than send blind — a missed follow-up is recoverable, a message to someone who already replied is not. Owner call: default to skip.
4. Verify the message-event JS (`_LAST_SENDER_JS`, `.msg-s-*` classes) against a genuinely opened thread — it read correctly on the messaging page (18 events, names resolved), so the classes appear alive; only the entry point is broken.
5. Add a live-validation check to `scripts/linkedin_live_validation.py` covering this path so the next drift is caught by the probe, not by users.

## Related
- #403 (F2) — this is that issue's second scope item, now answered.
- #236 duplicates the same E2E validation ask; close it in favour of this once fixed.
- The group-composer half of #403 validated CLEAN (see #403 comment).

gitchrisqueen:
## What & why

`check_dm_replied` opened a person's message thread by clicking `button[aria-label^='Message']`. On live LinkedIn (probed 2026-07-27) that control is an `<a href='/messaging/compose/…'>`, so the locator missed on every profile, the function fell through to `return False` — which the caller reads as *"no reply, keep sending"* — and `process_user_followups` kept sending follow-ups to people who had already answered. It failed **silently**: a defensive `False` is indistinguishable from a genuine no-reply.

Per the owner's domain note on the issue, this is **not** a selector swap. LinkedIn varies the messaging entry point by time, profile and connection state, so hardcoding today's anchor would just move the breakage to the next rotation.

## The change

**New: `src/cqc_lem/utilities/linkedin/message_thread.py`** — the ONE way LEM opens and reads a 1:1 thread.

- `open_message_thread(driver, wait, profile_url) -> ThreadOpen` walks **six routes in order**, stopping at the first that yields a verified-open thread:
  1. `anchor` — `a[href*='/messaging/compose/']` (today's live shape) / any anchor whose text is "Message"
  2. `button` — the legacy `button[aria-label^='Message']` forms
  3. `text_node` — tag-agnostic: the nearest clickable ancestor of a node whose trimmed text is exactly "Message"
  4. `overflow` — open each top-card **More** menu and re-run routes 1–3 inside it
  5. `direct_url` — navigate straight to `/messaging/compose/?profileUrn=<urn>`, depending on no rendered control at all
  6. `messaging_search` — `/messaging/`, search the name, open the matching conversation
- **Verification is part of the contract.** A route "succeeds" only when `msg-s-*` message events are readable or a compose form is present — on **either surface**, since opening from a profile may yield the bottom-right `msg-overlay-*` chat rather than `/messaging/`.
- No locator keys on a class name (they are obfuscated hashes) — only `href`, `aria-label` and visible text. The profile URN for route 5 is captured **before** any route navigates away.
- The winning route is logged via `log_info(..., action_type='followup')`, so telemetry shows the rotation over time and which fallbacks actually carry traffic.
- A route that raises never ends the ladder; every failure mode returns "not opened" rather than throwing.

**Three states, not a bool.** `check_dm_replied` now returns `ThreadState.REPLIED / NOT_REPLIED / UNKNOWN`. `UNKNOWN` covers: no route opened a thread, the thread opened but nothing was readable, or we have no self-name to compare the last sender against (that last case previously read as "they replied"). `process_user_followups` treats `UNKNOWN` as **skip**: it does not send, does not `mark_followup`, and leaves the row due so the next run re-reads it — the owner's call, since a missed follow-up is recoverable and a message to someone who already replied is not. The run summary now reports `skipped N unreadable thread(s)`.

The thread readers (`read_last_sender` / `read_last_message`) moved into the shared module so the nurture (#485) and inbound-intent (#483) paths read threads through the same code.

**Live probe:** `scripts/linkedin_live_validation.py --dm-thread-url <profile>` reports which route resolved, on which surface, the event count and what the reply reader sees — read-only (it opens and reads; it types and sends nothing). That is the early-warning signal for the next rotation.

## Testing

- 25 new unit tests (`tests/unit/utilities/linkedin/test_message_thread.py`): every route with a stubbed driver, the verify-before-success contract (a control that clicks but opens nothing is **not** a success), a route that raises mid-ladder, hidden controls, overlay vs page, URN capture before navigation, and the messaging-search name/slug match.
- `test_dm_followups.py`: the three-state verdict plus a sequencer test proving `UNKNOWN` neither sends nor marks the row.
- New probe tests in `tests/unit/test_linkedin_live_validation.py`.
- Full unit suite green locally: **7888 passed**.

## Not covered here

The ladder's selectors are grounded on the owner's 2026-07-27 DOM capture plus the previously-verified `msg-s-*` reader; the ladder itself has not been exercised against a live session (headless agent — no LinkedIn credentials). That is exactly what the new `--dm-thread-url` probe is for. See the decision comment below.

Closes #731


gitchrisqueen:
Held (`needs-human`, risk: **live-linkedin**). The ladder is grounded on your 2026-07-27 DOM capture and the already-verified `msg-s-*` reader, but I run headless with no LinkedIn session, so no route has been walked against a real profile — and this PR also changes what happens when the thread can't be read at all.

### 1. Live grounding before merge?
- **A. Merge now, probe after deploy** — run `sudo docker exec -i celery_worker_selenium python - --user-id 1 --dm-thread-url '<a profile you have DM'd>' < scripts/linkedin_live_validation.py`; it reports which route won. Worst case today's `anchor` route misses too, and reply detection stays where it already is (no false "no reply" — it now skips instead of spamming).  ✅ *recommended*
- **B. Hold the PR until you run the probe** — you paste the JSON, I adjust the locators first. Safest, but reply detection stays dead meanwhile and follow-ups keep going out under the current code.
- **C. Merge, but pause DM follow-ups until the probe passes** — no follow-up traffic at all until a route is proven live.

### 2. An UNKNOWN thread defers forever — bound it?
Today a thread we can never read is skipped and its `dm_followups` row stays due, so every run re-opens it (one Selenium thread-open each) and that person never gets a follow-up.
- **A. Leave it unbounded** — the row keeps retrying; the `Follow-up skipped:` WARNING plus the new `skipped N` run summary is the signal. Self-heals the moment the DOM comes back.  ✅ *recommended*
- **B. Stop the sequence after 3 consecutive UNKNOWNs** — bounds the cost, but silently ends real sequences during an outage.
- **C. Send the follow-up after 3 UNKNOWNs** — guarantees delivery, but reintroduces exactly the "message someone who already replied" risk, just later.

### 3. What if we don't know our OWN name?
The verdict is "is the last sender us?", so with `my_profile.full_name` empty every sender looks like someone else — the old code called that a REPLY. This PR calls it UNKNOWN, which also stops nurture drafting for that user.
- **A. Keep UNKNOWN** — a user with no name on their profile gets no follow-ups until it's filled in; nothing is sent on a guess.  ✅ *recommended*
- **B. Fall back to the profile-slug name** (`/in/chris-queen-123` → "chris queen") and compare against that — keeps the sequencer running, but a slug that doesn't match the display name reads as a reply.
- **C. Treat it as NOT_REPLIED and keep sending** — the pre-#731 behaviour for this one case.

**My recommendation: `1A 2A 3A`.** The change is strictly safer than what's in production right now (it can no longer send a follow-up on a *guess*), the probe is a one-command read-only check that tells us which route is actually carrying traffic, and both bounding options in 2/3 trade that safety back for convenience.

gitchrisqueen:
1A 2A 3A (Make a required field for the SPA settings configuration for the user to enter their name. Make sure this field(s) is aligned with what we need for this feature. If we need separate first and last name and for it to be the same as displayed on LI then make that explicit to the user. user_id=1 full name is Christopher Queen)

gitchrisqueen:
Addressed your review (`1A 2A 3A` + the settings ask). Commit `631da4c`.

**1A / 2A / 3A** — taken as-is, no code change needed: the PR already merges now and probes after deploy (1A), leaves an UNKNOWN row unbounded and due (2A), and keeps "no self-name → UNKNOWN" rather than guessing (3A).

**Your side-instruction — a required name field in Settings.** Built in this PR, since 3A is only safe if the name is actually knowable:

- **One field, not first/last.** LinkedIn labels a message group with the **full display name as one string** (`Christopher Queen`), and that string is the only thing reply detection can compare against — splitting and rejoining it would only invite a mismatch on middle names, suffixes or "Dr.". The card says so, and the requirement is explicit: *"Type your name exactly as it appears at the top of your LinkedIn profile — same spelling, accents and capitalisation LinkedIn shows on your messages."*
- **Where:** Settings → **Setup & Connection** → "Your LinkedIn Name", flagged **Required**, with an inline warning while it's empty (*"DM follow-ups stay paused until this is filled in"*) and a one-click **Use this** for the name LEM already scraped from your profile.
- **Also a required item on `GET /user/account-readiness`** (`linkedin_display_name`), so it appears in the "Finish account setup" checklist on the Account page, not only inside the section.
- **Storage:** `users.linkedin_display_name` (migration `V20260727194245`, identity on `users` like V55's reply token, so a prefs save can't clobber it). The migration **backfills every existing user from the profile LEM already scraped**, so your user 1 lands on `Christopher Queen` without you typing anything — and you can correct it in Settings if LinkedIn renders it differently. I deliberately did **not** hardcode your name into the migration.
- **Runtime:** `resolve_self_name()` is the one resolution — saved value wins, scraped profile is the fallback, `''` means UNKNOWN. Resolved **once per follow-up run**, and when it's empty the run logs exactly why every thread is being skipped instead of going quiet.
- **Live probe:** `--dm-thread-url` now also reports `reply_state` (`replied` / `not_replied` / `unknown`) and the `self_name` it used, so a saved name that doesn't match what LinkedIn renders shows up as a verdict rather than as an absence of follow-ups.

**Tests:** 6 unit tests for `resolve_self_name`, 2 sequencer tests (saved name beats the scraped one and is resolved once; nothing anywhere → every thread skipped), 7 db tests, 2 readiness tests (required item present/absent), 9 integration tests for the two endpoints, 2 new probe tests, and 5 vitest tests for the card (loads, required warning, suggestion, save, empty-save blocked). Unit suite green locally (7904 passed; the 2 failures in `test_error_capture_middleware.py` reproduce on the unmodified branch — pre-existing, not from this change). `npm test` (168) and `npm run build` green on Node 22.

Nothing deliberately left undone.

gitchrisqueen:
🔎 Claude adversarial review — FIXED 4 findings (commit `56fd5a0`)

All four are the same failure class, and it is the one this PR exists to close: the ladder **trusts an opened thread**, so any route that opens the *wrong* thread (or reads a good one as empty) produces a confident verdict, and a wrong `NOT_REPLIED` sends a follow-up to someone who already answered.

- **`profile_urn_from_page` could return a stranger's URN** (`message_thread.py`): the fallback took the *first* `urn:li:fsd_profile:*` in `page_source`. On a live profile page that is routinely the **viewer's own** URN (the global-nav Me menu) or a "People also viewed" card — so the direct-URL route, the issue's "strongest fallback", could compose to the wrong person (possibly ourselves) and then decide *this* person's follow-up from that thread. It is now **slug-scoped**: the URN nearest this person's `/in/<slug>` in the embedded model, `None` rather than a guess — which drops the ladder to messaging search instead of navigating somewhere wrong. Reached exactly in the rotation case the ladder is for (no compose anchor on the page).
- **Messaging search matched a stranger's conversation**: the row filter was `first_name.lower() in item.text.lower()`, so `"Jane"` matched **"Janet Smithers"**. Rows that link to a *different* `/in/<slug>` are now rejected outright rather than falling back to their label, and label matching is whole-word against both the stored name and the slug-derived one.
- **The self-name comparison had the same substring bug** (`check_dm_replied`): a saved display name of `"Chris"` matches `"Christine Baker"`, so her reply read as *our own* last message → `NOT_REPLIED` → follow-up sent. Since #731 makes this a **user-typed** field, that is now reachable by a typo, not just by chance. Both comparisons — and the live probe's `_reply_state`, which was quietly using its own `in` check and could report a verdict the sequencer never reaches — now share `name_matches()`.
- **`_wait_thread_open` returned on the first compose form**: LinkedIn paints the composer before the message list, so a perfectly readable thread was reported with `events=0` → `UNKNOWN` → and under option **2A** that row is deferred *indefinitely*, so the person never gets a follow-up and every run pays another six-route walk. Events now end the wait; a composer-only reading is held and returned only once the budget is spent.

Checked and **not** findings: `check_dm_replied` has exactly one caller (no stale bool-truthiness site — `ThreadState` is a `str` Enum and would have been truthy in all three states); the migration is safe (`profiles.data` is a `JSON NOT NULL` column and `user_id` is UNIQUE, so no invalid-JSON abort and no ambiguous join); `account-readiness` is a UI checklist only, so the new required item gates nothing at runtime.

**Close is honest**: every acceptance item in #731 is in the diff (six routes + verify-before-success, three-valued verdict, caller skips UNKNOWN, per-route unit tests, `--dm-thread-url` probe, shared readers for the nurture/inbound paths). The one remaining item — verifying the `msg-s-*` reader against a genuinely opened live thread — is the owner's `1A` decision (merge now, probe after deploy), not an untracked phase.

Tests: 13 new unit tests (wrong-person URN, no-URN-yields-none, slug-less page, `Janet`/`Christine` prefix cases, mismatched-profile row, composer-before-events race, `name_matches`). Full unit suite green: **7921 passed**.

gitchrisqueen:

PR #739, opened 2026-07-27
fix(outbound): throttle company-page invites into a paced daily drip (closes #732)

### LINKED ISSUE #732: Company-page invites: replace the monthly credit-exhausting blast with a throttled daily/weekly cadence that honors max_invites_per_day
## Problem (confirmed in code, 2026-07-27)

Company-page invitations currently fire as a **once-a-month blast that deliberately drains every remaining credit in one sitting**:

- **Schedule** (`app/my_celery.py`, `invite_to_company_pages`): `crontab(hour='5', minute='0', day_of_month='1')` — the 1st of each month at 05:00, one shot.
- **Volume** (`utilities/linkedin/company_page_inviter.py`, `automate_invitations`): reads `current_credits` from the page and calls `select_connection_checkboxes(driver, wait, current_credits)` — i.e. selects **as many invitees as there are credits** — then, if it couldn't select them all in one pass, **recurses into itself** ("If the selected_count is less than the current_credits, we can continue inviting") until the credits are exhausted.

### The bigger finding: the per-day cap already exists and this path ignores it
`engagement_preferences.max_invites_per_day` is a real, user-configurable setting (user 1 has it at 10). `utilities/human_pacing.py` even maps `ACTION_INVITE → "max_invites_per_day"` and lists invites in `ENGAGEMENT_ACTIONS`/`ENVELOPE_ACTIONS` — but **the company-page inviter never consults any of it**. The cap is honored for connection requests and not for page invites, so a user who set 10/day can still emit hundreds in one morning. That is the actual bug; the cadence is the design question on top of it.

### Why this matters
A burst of dozens-to-hundreds of page invites in one session is the most bot-like pattern in the whole product: identical action, machine cadence, one narrow window. It is exactly the behaviour the 2026 anti-automation enforcement keys on (velocity + regularity), and it undermines the human-pacing work shipped in #626 (read-time delays, jitter, variable daily volumes, rest days) which every other outbound path now flows through.

## Scope
1. **Cadence**: replace the monthly cron with a **daily** beat entry (invites are then paced within the day). Keep a per-user weekly option if the research below favours it, but daily-with-a-small-cap is the expected default. Align the run window with the user's OTHER cadences — same timezone/golden-hour logic the engagement lanes use — rather than a fixed 05:00 UTC for everybody.
2. **Cap + pacing**: route this path through `human_pacing` like every other outbound action — honor `max_invites_per_day`, draw the day's budget with the same variance/rest-day behaviour, and jitter spacing between invites. **Delete the recursion** in `automate_invitations`; it must send at most the remaining daily budget, never 'all credits'.
3. **Credit awareness without exhaustion**: keep reading available credits (a run must stop early when credits hit 0) but treat credits as a CEILING, not a target. Spread the monthly credit pool across the month — e.g. default daily budget ≈ min(max_invites_per_day, remaining_credits / days_left_in_month), so the pool lasts and never front-loads.
4. **Research the right numbers (2026)**: LinkedIn's page-invite credit model (commonly ~250/month, replenished when an invite is accepted), any published/observed per-day guidance, and what invite acceptance rates look like when spread vs blasted. Pick defaults from that and DOCUMENT the source in the PR — do not guess silently. Note the acceptance-replenishment mechanic argues strongly for spreading: accepted invites return credits, so a steady drip can invite MORE people per month than one blast.
5. **Config surface**: expose the invite cadence/daily cap in the engagement-preferences UI next to the other caps, defaulting conservatively; respect the brand-account phase ceilings (`utilities/brand_account.py` already sets `max_invites_per_day` per launch phase).
6. **Observability**: log invites sent/skipped per run with the reason (budget reached / credits exhausted / paused) and emit the existing engagement telemetry so the analytics dashboards can show invite volume over time.

## Acceptance
- Unit tests: daily budget derivation (incl. credits-remaining/days-left spreading), cap enforcement, no-recursion guarantee, rest-day/jitter behaviour, early stop at 0 credits.
- The monthly cron is gone; a daily task exists and is idempotent (a second run the same day sends nothing beyond the budget).
- A user with `max_invites_per_day = N` can never exceed N page invites in a day, verified by test.
- PR documents the researched 2026 numbers behind the chosen defaults.

## Context
Related: #626 (human-pacing engine — the machinery to reuse), `utilities/brand_account.py` (phase ceilings), `docs/engagement-growth-analysis-2026-07.md` (why bot-like velocity patterns are the risk).

gitchrisqueen:
## What & why

Company-page invitations fired **once a month and deliberately drained the Page's entire invitation-credit pool in one sitting**. `automate_invitations` read `current_credits`, selected that many invitees, and then **recursed into itself** ("If the selected_count is less than the current_credits, we can continue inviting") until there were none left — all on one `crontab(hour='5', minute='0', day_of_month='1')`.

It was also the only outbound lane that consulted neither `max_invites_per_day` nor the #626 human-pacing engine. `human_pacing` already maps `ACTION_INVITE → max_invites_per_day` and lists invites in `ENGAGEMENT_ACTIONS`/`ENVELOPE_ACTIONS`; this path just never asked. A user with 10 invites/day could still emit hundreds in one morning — identical action, machine cadence, one narrow window, which is exactly the velocity pattern 2026 enforcement keys on.

### A run is now bounded by the smallest of three ceilings

1. **The user's cap** — new `max_company_page_invites_per_day` (default **5**), clamped by `max_invites_per_day`. The account-wide cap is the harder bound on purpose: it is what `brand_account`'s `PHASE_OUTBOUND_POLICY`/`BRAND_CAP_CEILINGS` already clamp, so the brand account's phase ceiling governs page invites with **no second ceiling to keep in step**. That cap goes through `human_pacing` like every other lane — stable 40–100% daily draw, weekend asymmetry, rest days, and the shared account envelope — minus what today's log rows say was already spent.
2. **The credit spread** — `credits_remaining / days_left_in_month`, so the pool lasts the month instead of being front-loaded. Floored at 1 while any credit remains; 0 credits is 0, never 1.
3. **The live credit count** — a hard stop at 0, read off the page.

**The recursion is deleted.** One run is one batch. `plan_daily_invites` decides the allowance **before** a Chrome session is opened — on most days it is zero, and a browser spent discovering that is a lane an engagement task needed. A paused account stands down: a page invite is discretionary amplification, never a response owed to someone.

**Idempotency is durable, not Redis.** A page invite is a batched UI action (select N, click Invite once), so one log row carries a count; `count_company_page_invites_sent_today` **sums** the trailing number in the `Company page invites sent: <n>` marker rather than counting rows. A second run the same day therefore sends nothing beyond the budget, across worker restarts.

**Cadence** is now the same staggered per-user daily tick the other fan-outs use — `STAGGER_COMPANY_INVITE` (10:00 on the user's own clock, 3h window), dispatched with `dispatch_jitter_seconds()`, ticking every `STAGGER_TICK_MINUTES`. It opens after the appreciation-DM window has drained so the two don't queue on `se_outreach` together.

**Observability**: every run emits `company_page_invite_run` — *including* the ones that send nothing, with the reason (`sent` / `budget_reached` / `disabled` / `credits_exhausted` / `no_candidates` / `no_page` / `failed` / `paused`). A series carrying only sends can't tell "paced down to zero" from "silently broken".

**Config surface**: the cap sits in Account → Volume next to the other daily caps, with an `inform` finding (C29) when it exceeds `max_invites_per_day` and would therefore silently do nothing.

## The researched 2026 numbers behind the defaults

- **LinkedIn Help — [Invitation limits for your LinkedIn Page](https://www.linkedin.com/help/linkedin/answer/a547492)** (official, checked 2026-07-27): "Pages are granted monthly invitation credits… The invitation credit limit renews on the first day of every month." And critically: **"When an invitation is accepted, the credit is earned back and applied back to the Page's balance. It can take up to 72 hours for the credit to reflect in the Page's balance."** The page also says the amounts are *in transition*: "We are changing the available invitation credits for free Pages and Premium Company Pages. This change is gradually being made available." — so the exact number is deliberately **not** hardcoded anywhere in this PR; we read it live off the page and divide.
- **The size of the pool is being cut.** The long-standing figure was 250/month per Page, shared across all admins. Per [The X Concept, March 2026](https://thexconcept.com/2026/03/18/linkedin-just-killed-your-company-page-growth-what-the-50-invitation-limit-really-means-for-your-business/), free Pages are being reduced to **50/month** (Premium 300), in a staggered rollout — corroborated by [The Scott Partnership](https://www.scottpr.com/linkedin-is-reducing-follower-invite-credits-whats-changing-and-what-to-do-now/). A default had to survive **both** pool sizes, which is why the spread is computed from live credits rather than from an assumed 250.
- **The replenishment mechanic argues for spreading, exactly as the issue predicted.** Because an accepted invite refunds its credit (within the same month), a steady drip can invite *more* people per month than a single blast: a blast spends the pool before any acceptance can return anything, whereas a drip keeps re-spending returned credits all month.
- **Per-day pacing.** LinkedIn publishes no per-day page-invite guidance, so the daily number is anchored to the invite guidance that does exist: ~100 connection invites per rolling 7 days across all account tiers, with a commonly-cited safe working pace of **10–25/day** and "never more than 10 in an hour" ([Overloop](https://overloop.com/blog/linkedin-limits), [LeadLoft](https://www.leadloft.com/blog/linkedin-limits), [Taplio](https://taplio.com/blog/linkedin-connection-request-limit)).

**Why 5/day.** It sits well under the 10–25/day safe band for a lighter-weight action (no message, existing connections), and it behaves correctly at both pool sizes: on a **50-credit** Page the credit spread binds first (~1–2/day), on a **250-credit** Page this cap binds at 5, and #626's 40–100% draw pulls the realised average lower still. The number is user-editable (0–50) and 0 turns the lane off.

## Testing

- `tests/unit/app/test_company_page_invites.py` — 39 tests: credit-spread derivation (full pool on the 1st → 8/day not 250, a 50-credit pool → 1/day, thin-pool-late-in-month floor, 0 credits → 0, last-day behaviour), cap enforcement incl. the `max_invites_per_day` ceiling and the brand P0 phase policy, NULL-cap-is-not-zero, the pacing/envelope wiring, **the no-recursion guarantee** (one credit read, one selection pass, one log row when fewer than the budget are selectable), early stop at 0 credits, the countable log marker, the governor hand-off, the selection-pause bounds (and its no-op when pacing is off), the task's no-Chrome-slot and paused paths, the staggered daily beat, and that the monthly `day_of_month` crontab is gone.
- Updated `test_run_scheduler.py`, `test_daily_engagement.py` (the staggered-fan-out cadence invariant now covers `invite_to_company_pages`) and `test_db_engagement_prefs.py` (column coverage).
- Full unit suite green locally: **7895 passed**.

## Migration

`V20260727175938__add_company_page_invite_cap.sql` — additive nullable `engagement_preferences.max_company_page_invites_per_day`. NULL reads as the 5/day default, never as 0 (a NULL read as 0 would silently switch the lane off for every existing row); an explicit 0 IS "off" and is preserved.

Closes #732


gitchrisqueen:

- **Shared pacing budget key with the connection-invite lane** (`company_page_inviter.plan_daily_invites`): the new lane drew its daily budget under `ACTION_INVITE`, and `daily_budget` keys its stored draw on the action name alone — so whichever lane drew first that day wrote the *other* lane's budget. With the page cap at 5 and `max_invites_per_day` at 10, a page-invite draw silently clamped connection requests to roughly half their allowance for the rest of the day (measured across users: 4→2, 6→3, 5→2, 10→5). It also self-inflicts inside one `plan_daily_invites` call, since `remaining_actions` draws the lane budget *before* `account_envelope` re-reads the same key. **Fixed**: added `ACTION_COMPANY_INVITE` to `human_pacing.py` and draw against it; the lane still SPENDS the shared envelope (it keeps `record_action`-ing under `ACTION_INVITE`, which is also its harder ceiling) and deliberately does not enlarge it. Regression test `test_it_draws_its_own_budget_and_never_the_connection_lanes` fails against the pre-fix code.
- **A browser that never comes up emitted no run event** (`run_automation.automate_invites_to_company_page_for_user`): `get_driver_wait_pair()` sat outside the try, so a Chrome-session failure (grid full, container restart) escaped before `track_company_page_invite_run` — the one failure mode the every-run event exists to make visible, and indistinguishable from a day paced down to zero. **Fixed**: session acquisition moved inside the reporting path, emitting a new `session_failed` status, matching the golden-hour sweep's shape. Test added.

Checked and found clean: the three-ceiling arithmetic and its month-boundary/floor cases, the deleted recursion, `count_company_page_invites_sent_today`'s `REGEXP_SUBSTR` sum (no digits precede the count in the marker; no overlap with `CONNECTION_REQUEST_SENT_MESSAGE` rows), the NULL-cap-is-the-default read in `_select_engagement_row` and the clamp in `update_engagement_preferences`, the additive timestamp migration, the `STAGGER_COMPANY_INVITE` beat (page-check-before-slot-claim ordering is right; not a `PINNED_MIDPOINT_HOURS` fan-out), and the API/registry/conflicts surface.

Full unit suite green locally: **7897 passed**.

gitchrisqueen:

PR #740, opened 2026-07-27
fix(ui): let comma-separated settings fields accept commas and spaces (closes #638)

### LINKED ISSUE #638: fix(other): Fix input fields blocking comma and space entry

gitchrisqueen:
## What & why

Reported via in-app feedback: *"Many settings asking for comma separated values don't allow the user to enter commas or spaces. Something is blocking when trying to type in the fields."*

Reproduced. The eight list settings in the Account settings hub bound their `<input>` straight to `csv(list)` while publishing `parseCsv(e.target.value)` on every keystroke:

```tsx
<input value={csv(eng.focus_topics)} onChange={(e) => setEng({ focus_topics: parseCsv(e.target.value) })} />
```

`parseCsv` trims each entry and drops empty ones, so `"ai,"` → `["ai"]` → `"ai"` and `"ai, "` → `"ai"`. The typed character was parsed away and the box re-rendered without it before the next keystroke landed — the comma and the space after it could literally never be entered. The only way to get a second entry in was to paste it.

## The fix

New `settings/CsvInput.tsx` holds the raw text in local state and publishes only the parsed array upward:

- A prop equal to the canonical form of what it last published is its **own echo** and is ignored — that's the write-back that ate the separator.
- A value changed from **outside** (preferences loading, a one-click conflict fix, a preset) still replaces the box.
- On **blur** the text is tidied to the canonical `a, b` form, so `"ai,, sales ,"` settles as `"ai, sales"`.

Wired into all eight affected fields — `include/exclude_topics`, `include/exclude_keywords`, `include/exclude_authors` (Targeting), `focus_topics` (Voice), `connection_target_authors` (Outreach). No API, model or storage change: the same `string[]` is sent on save.

## Testing

- New `CsvInput.test.tsx` (7 tests) drives the component through a harness that owns the parsed array and hands it straight back down — the exact wiring that caused the bug. Covers typing a comma, typing the space after it, trimmed/empty-filtered publishing, rendering a saved list, blur tidying, an external replacement, and a null value.
- Full SPA suite green locally: `npm test` → 16 files, 170 tests passed. `npm run build` (tsc -b + vite build) and `eslint` on the touched files are clean.

Closes #638

### DIFF
--- a/src/cqc_lem/ui/src/pages/account/settings/CsvInput.test.tsx
+++ b/src/cqc_lem/ui/src/pages/account/settings/CsvInput.test.tsx
@@ -1,0 +1,84 @@
+import { useState } from 'react'
+import { describe, expect, it, afterEach, vi } from 'vitest'
+import { cleanup, fireEvent, render, screen } from '@testing-library/react'
+import CsvInput from './CsvInput'
+
+afterEach(cleanup)
+
+// The real wiring: the parsed array is state OWNED BY THE PARENT and handed straight back down,
+// which is what used to erase the separator on every keystroke (issue #638).
+function Harness({ initial = [], onChange }: { initial?: string[]; onChange?: (v: string[]) => void }) {
+  const [values, setValues] = useState<string[]>(initial)
+  return (
+    <>
+      <CsvInput
+        aria-label="topics"
+        value={values}
+        onChange={(v) => {
+          setValues(v)
+          onChange?.(v)
+        }}
+      />
+      <span data-testid="parsed">{JSON.stringify(values)}</span>
+    </>
+  )
+}
+
+const box = () => screen.getByLabelText('topics') as HTMLInputElement
+
+describe('CsvInput', () => {
+  it('lets a comma be typed instead of collapsing it back to the previous entry', () => {
+    render(<Harness />)
+    fireEvent.change(box(), { target: { value: 'ai' } })
+    fireEvent.change(box(), { target: { value: 'ai,' } })
+    expect(box().value).toBe('ai,')
+  })
+
+  it('keeps the space after a comma while the next entry is being typed', () => {
+    render(<Harness initial={['ai']} />)
+    fireEvent.change(box(), { target: { value: 'ai, ' } })
+    expect(box().value).toBe('ai, ')
+    fireEvent.change(box(), { target: { value: 'ai, sales' } })
+    expect(box().value).toBe('ai, sales')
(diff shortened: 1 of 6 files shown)


gitchrisqueen:

Reviewed the diff, the surrounding wiring (`EngagementPrefsContext`, `ConflictsContext`, `Field`, `presets`), and the whole SPA for other CSV round-trip inputs. The core fix is correct — I re-ran the suite under Node 22 locally (17 files / 181 tests green, `tsc -b` and `eslint` clean on the touched files) and probed the echo-guard against StrictMode double-rendering, an external replacement landing mid-typing, a parent that never accepts the published value, and blur-then-keep-typing. All behave.

- **Test gap on the changed behaviour**: `CsvInput.test.tsx` drives a harness written for this PR, so a section that reverts to binding an `<input>` straight to `csv()`/`parseCsv()` would not fail a single test — that is precisely the shape of the reported bug. Added `settings/csvFields.test.tsx`, which renders the three REAL sections and types through all eight list settings (plus blur-tidy, per-field independence, and StrictMode). Verified it fails 11/11 against the pre-fix binding and passes 11/11 with the fix.
- **Silent failure mode in `CsvInput`**: `{...rest}` was spread AFTER `onBlur`, so any future caller passing an `onBlur` would silently replace the tidy handler (`value`/`onChange`/`type` are type-omitted, `onBlur` is not). Moved the spread ahead of the component's own handlers.

Checked and NOT findings: no other CSV round-trip inputs remain in the SPA (`tone`/`comment_style` bind raw strings, so commas already typed fine); `inputClass` is still used in all three sections, so no unused-import build break; no conflict `fix` patch targets a list field, so the echo guard has no live external-write path to fight; publishing a new array on every keystroke does not mark the form falsely dirty (`isDirty` compares JSON).

Unrelated to this PR: the E2E job is red on `redis.exceptions.ConnectionError: Error 111 connecting to localhost:6379` in `test_carousel_video_post_e2e.py` (`auto_second_wave_comment` → celery-once lock). Pre-existing infra, not a required gate, and untouched by a UI-only diff.

gitchrisqueen:

Issue #659, opened 2026-07-26
Activate the marketing engine: owner-only steps (brand account, launch phase, caps)


gitchrisqueen:
## 🧑‍⚖️ Owner checklist — the marketing engine is BUILT but DORMANT
The full launch/marketing build (#496-#507) shipped: feedback→issue loop, NPS/review capture, 60-day early-adopter trial grants, onboarding checklist + nudges, auto-changelog, auto-FAQ, video-tutorial pipeline, and the brand-account dogfooding engine (utilities/brand_account.py, phase-gated caps). None of it can RUN because the inputs only you can provide are missing. Prod state 2026-07-26: users table contains ONLY user 1; BRAND_ACCOUNT_ENABLED unset; TUTORIAL_VIDEOS_ENABLED unset. (FEEDBACK_GITHUB_TOKEN and POSTHOG_PERSONAL_API_KEY are now present — those loops are live.)

Reply with the item numbers you've completed, or ask for any to be prepared further.

### 1. Brand LinkedIn account (unblocks #504's engine)
- Create/designate the LEM brand LinkedIn account (a real account — company pages can't comment/DM the same way).
- Add it as a user in the SPA, store its LinkedIn password (Selenium login) or connect a session via the LinkedIn Connect extension.
- Set in /opt/lem/.env: BRAND_ACCOUNT_ENABLED=True, BRAND_ACCOUNT_EMAIL=<its email>, BRAND_SIGNUP_URL=<trial signup URL>; restart app services (or wait for next deploy).
- brand_account.sync_brand_preferences() then enforces P0 phase caps automatically.

### 2. Confirm LAUNCH_PHASE + P0 outbound caps
- LAUNCH_PHASE defaults to P0 (lowest volume). Confirm the P0 brand caps in brand_account.brand_outbound_policy are the numbers you're comfortable with under LinkedIn ToS — this was #504's risk:product-decision that was never explicitly signed off.

### 3. Video tutorials (unblocks #505's weekly cadence)
- Set TUTORIAL_VIDEOS_ENABLED=True + YouTube API credentials per docs, if you want the weekly tutorial pipeline live.

### 4. Early-adopter program switch
- Confirm early_adopter_slots seeding (P0 cap 25) and the 60-day/coupon terms from the plan's §A.2 — that was also a deferred risk:product-decision.

[...]

Issue #504, opened 2026-07-24
feat(marketing): dogfooding self-marketing — LEM brand account content + comment→connect→DM funnel


gitchrisqueen:
## Why
Flagship channel (docs/launch-and-marketing-plan.md (PR #495) Section C.2): LEM markets LEM by running the product's own automation on a dedicated brand LinkedIn account.

## Scope
- Onboard a LEM brand account as a first-class user (focus_topics = ICP pains) and drive its 30-day content plan (`plan_content_for_user`/`auto_generate_content`), feed commenting (`automate_commenting`), connect targeting (`invite_to_connect`), appreciation/outreach DMs + follow-ups (`build_dm_from_template`, `automate_appreciation_dms_for_user`, `process_user_followups`), newsletter, and company-page invites — all pointing to the trial signup.
- Runs through the SAME per-day caps, 429 backoff (`rate_limit.py`), and per-user proxy as any user. A phase-gated `LAUNCH_PHASE` config sets brand outbound volume.

## risk:product-decision
Outbound at scale + LinkedIn ToS judgment — owner signs off on brand daily connect/DM caps and cadence.

## Files
Config/seed for the brand user; scheduling glue in `app/run_scheduler.py`; no new outreach primitives (reuse existing).

## Acceptance
- Brand account runs the existing tasks under caps; volume is config-gated; tests for the scheduling glue; ≥80% patch coverage.

PR #741, opened 2026-07-27
feat(outbound): resolve the brand account as user 1 by convention (closes #736)

### LINKED ISSUE #736: Brand account = user 1 by convention (drop BRAND_ACCOUNT_* env vars)
## Owner decision (2026-07-27)
**user_id=1 IS the LEM brand account, permanently** — it is the owner's own account and always will be. The current design (`BRAND_ACCOUNT_ENABLED` / `BRAND_ACCOUNT_EMAIL` in `.env`, resolved by `brand_account.get_brand_user_id()`) is unnecessary indirection that has kept the whole self-marketing engine (#504) dormant because two env vars were never set on the box.

## Scope
- `utilities/brand_account.py`: resolve the brand user as **user_id 1** by convention. Remove `BRAND_ACCOUNT_EMAIL` and `BRAND_ACCOUNT_ENABLED` from `env_constants.py`, `.env.example`, `.env.prod.example` and every call site; delete the email→id lookup.
- Keep an explicit override ONLY if it costs nothing (e.g. an optional `BRAND_USER_ID` honored when set) — but the DEFAULT must work with zero configuration. No env var may be required for the brand engine to run.
- `is_brand_user(user_id)` becomes `user_id == brand_user_id()`; `sync_brand_preferences()` keeps enforcing the phase caps/focus topics on that user exactly as today.
- **Guard the collision**: user 1 is also the owner's ordinary LEM account. Brand-preference syncing must not silently overwrite settings he tuned by hand — reconcile with the existing 'seeded fields vs user-owned fields' split in `brand_preference_overrides`, and log what it changes. This is the one real risk in the change; cover it with tests.
- `BRAND_SIGNUP_URL` stays (it is a real content input, not a wiring flag) but should default sensibly rather than blocking startup when unset.

## Acceptance
- With NO brand env vars set, the brand engine resolves user 1 and runs under its phase caps.
- Tests: brand resolution without env; override honored when present; preference sync does not clobber owner-tuned fields it does not own.
- `docs/launch-and-marketing-plan.md` §C updated to match.

Related: #659 (owner activation checklist — this removes two of its four steps), #504 (the engine).

gitchrisqueen:
## What & why

The self-marketing engine (#504) has been **dormant in prod since it shipped** — not because of a bug, but because `BRAND_ACCOUNT_ENABLED` / `BRAND_ACCOUNT_EMAIL` were never set on the box. That pair was wiring that could only ever fail closed. Per the owner decision on #736, **user_id 1 IS the LEM brand account, permanently**, so it is now resolved by convention.

### Brand resolution
- `brand_account.brand_user_id() -> int` replaces `get_brand_user_id() -> Optional[int]`. It **never returns None**: unset, blank, non-numeric, `0` or negative `BRAND_USER_ID` all fall back to user 1, so a typo can't silently switch self-marketing off.
- `BRAND_ACCOUNT_ENABLED` / `BRAND_ACCOUNT_EMAIL` are gone from `env_constants.py`, `.env.example` and every call site, along with the email→id DB lookup. `.env.prod.example` never carried them.
- The optional override is `BRAND_USER_ID` — it costs nothing and the **default works with zero configuration**.
- `is_brand_user(user_id)` is now `int(user_id) == brand_user_id()`, and returns `False` rather than raising on a non-numeric argument.
- `auto_sync_brand_account` can no longer return `"Brand account not configured"` (which is what it returned in prod every night at 00:45); the only failure it reports now is a failed write.

### The collision guard (the one real risk)
User 1 is also the owner's **ordinary** LEM account, and the nightly sync used to *assert* the phase's caps and connect posture over anything he tuned by hand. The phase policy is now applied as a **ceiling, not an assignment**:

- `max_comments_per_day` / `max_dms_per_day` / `max_invites_per_day` are only ever pulled **down** to the phase number (`_capped`). A hand-tuned `3` survives P1's `15`; a hand-tuned `999` still clamps to `15`. `0` is honoured as a real choice; a missing or unreadable value takes the phase's.
- `connection_request_mode` / `connection_targeting_mode` are only ever **tightened** (`_strictest`, strictness order `pre_review > auto_approve` and `off > suggest > auto_queue`). An unknown saved value isn't comparable, so the phase's wins.
- Honouring a stricter value can never widen outbound, so the volume gate is intact — this only stops the sync stomping settings it doesn't own.
- Seeded content fields (`focus_topics`, `business_goals`) keep the existing fill-only-when-empty split, and only the policy fields are still sent to the upsert (the #639 abort is preserved).
- **Nothing is silent:** `preference_changes()` diffs before/after and `sync_brand_preferences` logs every field it actually changes in one INFO line (`max_dms_per_day: 20 -> 5`). A settled account logs nothing at INFO — a line that fires whether or not anything moved is not a change log. The diff normalizes string-vs-int columns so a DB `"5"` against a policy `5` isn't reported as a nightly edit.

### `BRAND_SIGNUP_URL`
Stays (it's a real content input), but is now an override rather than a requirement: `attribution.signup_landing()` falls back to `PUBLIC_BASE_URL` — the landing page *is* the signup surface (there is no `/signup` route). `signup_url()` and `referral_url()` both read it, so the brand's trial CTA also needs no extra config. Empty only when neither is set, which every caller still reads as "no CTA link to publish".

### Docs
`docs/launch-and-marketing-plan.md` §C.2 gains a "who the brand account is" paragraph and a "collision guard" block; the status table row and `docs/marketing-attribution.md`'s owned-host / verification notes are updated.

## Testing

`poetry run pytest tests/unit -q` → **7917 passed**.

New/changed coverage in `tests/unit/app/test_brand_account.py`:
- resolution with **no env at all** → user 1; explicit override honoured; parametrized unusable overrides (`nope` / `0` / `-3` / `1.5` / `None`) all fall back; `is_brand_user` across ints, strings, `None` and junk, and following the override.
- `TestOwnerTunedSettingsSurvive`: stricter cap kept, `0` kept, stricter posture kept while a looser one is tightened, unknown mode falls back, and a property check that no phase can ever be raised above its own policy.
- `TestPreferenceChanges`: only real changes reported, `"5"` vs `5` is not a change, list-valued focus topics compare by value, unset field is a change.
- sync-level: syncs user 1 with no config, does not clobber the owner's stricter caps under P2, logs every changed field with `user_id=1`, and logs no edit for a settled account.
- `test_marketing_attribution.py`: signup/referral URLs fall back to `PUBLIC_BASE_URL`; empty only when neither is configured.

Related: #659 (owner activation checklist — this removes two of its four steps), #504.

Closes #736

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -396,11 +396,10 @@ EARLY_ADOPTER_COUPON_ID=
 # P0 private early-adopter / P1 open beta / P2 GA. Sets the brand account's daily
 # comment/DM/connect caps and connect approval posture (see utilities/brand_account.py).
 LAUNCH_PHASE=P0
-# The LEM brand LinkedIn account, onboarded as a normal user (same caps, 429 backoff, proxy).
-# OFF by default — it sends real outbound.
-BRAND_ACCOUNT_ENABLED=False
-BRAND_ACCOUNT_EMAIL=
-# Trial signup URL the brand's content/DMs steer toward. Empty = no CTA seeded.
+# The LEM brand LinkedIn account is user 1 by convention (issue #736) — no configuration needed.
+# Set this ONLY to seat the brand on a different account; blank/invalid = user 1.
+BRAND_USER_ID=
+# Trial signup URL the brand's content/DMs steer toward. Empty falls back to PUBLIC_BASE_URL.
 BRAND_SIGNUP_URL=
 # Extra hosts whose analytics we own (marketing site, blog, docs) — comma separated, bare host or
 # full URL. Only these plus PUBLIC_BASE_URL/BRAND_SIGNUP_URL get UTM-tagged on outbound links.
--- a/docs/launch-and-marketing-plan.md
+++ b/docs/launch-and-marketing-plan.md
@@ -37,7 +37,7 @@ re-file work that already exists. Sections A–D remain the plan of record for e
 | Onboarding/activation checklist + stalled-user nudges (A.3) | **Shipped** | `OnboardingChecklist.tsx`, `onboarding_state` (`V20260725090900__add_onboarding_state.sql`), `onboarding-nudges` beat |
 | Extended-trial endpoint + cohort slots (A.2) | **Shipped** | `POST /trial/extend`, `extend_trial_for_user`, `early_adopter_slots` / `early_adopter_grants`, `EARLY_ADOPTER_*` + `LAUNCH_PHASE` env |
 | Funnel instrumentation (C.5) | **Shipped** | `track_funnel_event` in `observability.py` |
-| Brand-account dogfooding (C.2) | **Partial** | `sync-brand-account` beat (`auto_sync_brand_account`), gated by `BRAND_ACCOUNT_ENABLED`; the brand runs the existing content/engagement/newsletter tasks |
+| Brand-account dogfooding (C.2) | **Partial** | `sync-brand-account` beat (`auto_sync_brand_account`); the brand user is **user 1 by convention** (issue #736, no env var required) and runs the existing content/engagement/newsletter tasks |
 | Passive-signal mining (B.1) | **Planned** | — |
 | SEO, email-nurture, referral, lead-magnet, retargeting and partner/affiliate agents (C.3, C.4) | **Planned** | — |
 | Analytics agent: CAC / channel-ROI rollups + optimization loop (C.5) | **Planned** | — |
@@ -232,9 +232,15 @@ economics, brand-voice guardrails) — never pressing "post."
 
 ### C.2 Dogfooding as the flagship channel — **LEM markets LEM**
 
-The single most credible proof that LEM works is LEM growing its own audience with LEM. A dedicated **LEM brand
-LinkedIn account** is onboarded as a first-class user of the product and runs the exact same automation users
-pay for:
+The single most credible proof that LEM works is LEM growing its own audience with LEM. The **LEM brand
+LinkedIn account** is a first-class user of the product and runs the exact same automation users pay for.
+
+**Who the brand account is: user 1, by convention (issue #736).** The first account on the box is the owner's
+own, it doubles as the brand account permanently, and `brand_account.brand_user_id()` resolves it with **zero
+configuration** — the earlier `BRAND_ACCOUNT_ENABLED` / `BRAND_ACCOUNT_EMAIL` pair was wiring that only ever
+failed closed, and kept this whole engine dormant in prod because two env vars were never set. An optional
+`BRAND_USER_ID` seats the brand elsewhere; blank or invalid falls back to user 1 rather than switching
+self-marketing off. What the brand does:
 
 1. **Content about the problem LEM solves.** The brand account's `focus_topics` are set to the ICP's pains
    ("consistent LinkedIn presence without the grind", "solo-founder pipeline", "AI content that sounds like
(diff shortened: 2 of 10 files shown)


gitchrisqueen:

Held (`needs-human`, risk: `product-decision`). This PR makes the dogfooding engine run with **zero config** — but it does that on **your own live account (user 1)**, and the nightly `sync-brand-account` beat (00:45 UTC) will start writing to its engagement preferences the night this merges. Three calls are yours, not mine.

### 1. How should the nightly sync treat caps/posture you tuned by hand on user 1?
- **A. Ceiling only — it can tighten, never loosen** (what this PR implements) — a phase cap you set *lower* by hand survives; anything above the phase number is pulled down; `pre_review`/`off` survive a phase that says `auto_approve`/`auto_queue`; every change is logged `before -> after`.  ✅ *recommended*
- **B. Hard-assert the exact phase values** (pre-#736 behaviour) — your hand-tuned settings are overwritten every night, but the brand account is guaranteed to sit exactly on the phase's numbers.
- **C. Log-only on the caps** — the sync reports drift and changes nothing; the phase stops being a gate and volume is entirely on you.

### 2. Which `LAUNCH_PHASE` should prod run at once this is live?
Prod has no `LAUNCH_PHASE` set, so it resolves to **P0** today. On merge, user 1's caps get clamped from the product defaults (20 comments / 20 DMs / 10 invites per day, `auto_approve` connects) to the phase's:
- **A. P0 — 8 comments / 5 DMs / 5 invites per day, connects held at `pre_review`, targeting `suggest`** — the conservative warm-up; you approve every connect. Nothing to change on the box.  ✅ *recommended*
- **B. P1 — 15 / 10 / 8, `auto_approve` + `auto_queue`** — connects go out unattended; set `LAUNCH_PHASE=P1` in `/opt/lem/.env`.
- **C. P2 — 20 / 15 / 10, `auto_approve` + `auto_queue`** — full GA volume immediately; set `LAUNCH_PHASE=P2`.

### 3. Where should the brand's trial CTA point when `BRAND_SIGNUP_URL` is unset?
- **A. Fall back to `PUBLIC_BASE_URL`** (what this PR implements) — the landing page *is* the signup surface (there is no `/signup` route), so posts/DMs/newsletter get a real UTM-tagged CTA with no extra config.  ✅ *recommended*
- **B. Keep it empty unless explicitly set** — no CTA link is published at all until you add `BRAND_SIGNUP_URL` to `/opt/lem/.env`; safer against publishing a URL you didn't pick, but re-introduces one of the "never set it, so it never ran" failures this issue is about.

**My recommendation: `1A 2A 3A`.** The ceiling rule keeps the volume gate intact (it only ever tightens) while making your own account safe to share with the brand engine, and P0 means the first night of real self-marketing runs at ~a third of the product's default volume with a human on every connect — you can advance the phase any time with one env var, and no code change.

gitchrisqueen:
1A 2A 3A

gitchrisqueen:
Addressed your review: the PR already implements all three choices — **1A** ceiling-only sync, **2A** P0 default phase, and **3A** fallback to `PUBLIC_BASE_URL` for the brand CTA — so no additional code changes are needed. Handing the PR forward to CI/review and merge.

gitchrisqueen:

gitchrisqueen:
MARK — FIXED 1 findings
- Fixed database error handling in get_engagement_preferences to return None instead of silently returning defaults on DB errors, preventing potential override of user-configured stricter limits

Issue #549, opened 2026-07-25
fix(infra): graceful worker shutdown + maintenance mode so deploys never lose in-flight Celery tasks


gitchrisqueen:
## Problem (HIGH PRIORITY): deploys kill in-flight Celery tasks and lose them
Every deploy recreates the app containers (`scripts/deploy.sh` → `docker compose up -d`), which SIGTERMs the celery workers. Long tasks (feed commenting loops, ~6-min Runway/Veo video generation, reply sweeps, DMs) get **killed mid-flight and lost** — no warm shutdown, no re-queue. This causes the "worker restarts interrupt tasks" behavior noted in the pre-post-commenting investigation (#547).

## Root causes (verified 2026-07-25)
1. **SIGTERM never reaches Celery.** `compose/local/celery/worker/start` and `start-selenium` launch the worker as `su $CELERY_USER -c "… celery … worker …"` — **not `exec`'d**. Docker's SIGTERM goes to PID 1 (the shell/`su`), which does not forward it to the celery child, so **Celery never gets the warm-shutdown signal**.
2. **No `stop_grace_period`** on any worker service (docker-compose.yml/prod.yml) → Docker's default **10s**, then SIGKILL — far shorter than a video/Selenium task.
3. **No global `task_acks_late` / `task_reject_on_worker_lost`** (only a couple of tasks set it) → default acks-on-receipt, so a task killed mid-run is **not re-queued → lost**.
4. **No drain/pause before recreate** — `deploy.sh` recreates workers while tasks are actively running.

### A. Make warm shutdown actually work
- Forward SIGTERM to Celery so `TERM` triggers Celery's **warm shutdown** (stop consuming new tasks, finish running ones, then exit). Use `exec` (e.g. `exec gosu $CELERY_USER … celery … worker …` or run under `tini`/`--init`) so the celery process is PID 1 / receives signals.
- Set a per-worker `stop_grace_period` longer than the longest task (e.g. **6–8m**; video gen ~6m) so Docker lets the warm shutdown finish before SIGKILL.

### B. Never lose a task that IS interrupted
- Set `task_acks_late=True` + `task_reject_on_worker_lost=True` (globally, or on all long/Selenium/media tasks) so a task killed before completion is **re-queued and re-run**.
- Tune `broker_transport_options.visibility_timeout` (Redis) above the longest task so a long-running task isn't re-delivered to a second worker mid-run (the classic acks_late+Redis double-run trap).
[...]

PR #751, opened 2026-07-27
fix(infra): make the Selenium Grid the deployed topology (it was reverted by every deploy)

gitchrisqueen:
**Cut over to the Grid on prod today (2026-07-27) — this makes it survive deploys.**

## The gap
`docker-compose.grid.yml` shipped in #610, but `scripts/deploy.sh` composes only `docker-compose.yml -f docker-compose.prod.yml`. So a manual cutover lasts exactly until the next release, which silently puts the box back on the single standalone container — healthy, and quietly the wrong topology.

## Changes
- `deploy.sh` composes the grid overlay by default. `SELENIUM_TOPOLOGY=standalone` (env or `/opt/lem/.env`) falls back; the overlay parks the standalone behind a profile rather than deleting it, so rollback stays one flag.
- **Transition guard**: evict a running `selenium-chrome` before bringing the grid up.

## Two failure modes found by doing it live (both now guarded + documented)
1. A compose **profile stops a service from being started, not from running** — the standalone kept holding `127.0.0.1:4444`, so the hub could not bind.
2. When that bind fails, Docker leaves the hub **running with no network attached**. It presents as *'hub unhealthy, 0 nodes registered'* with nodes logging `UnknownHostException: selenium-hub` — not as a port error. Recreating the hub fixed it.

## Verified on prod after the cutover
- hub healthy; `/status` → **8 nodes, 8 slots, all UP** (matches the summed lane concurrency invariant)
- a real `get_docker_driver()` session created, navigated and closed cleanly through the hub
- capacity monitor: `session_saturation ok / lane_backlog ok / session_wait ok`
- site `/health` 200; all workers + beat healthy
- 42 deploy/compose unit tests pass

## noVNC — restored by a dedicated node
The hub runs no browser and pool replicas cannot share a host port, so the *pool* is unwatchable.
The overlay therefore runs `selenium-node-debug`: a **ninth**, non-replicated node that publishes
7900 (loopback by default) and is what `lemvnc` points at. It sits on top of the eight the lanes are
sized for, so watching a browser never costs a production slot — and it is excluded from the
capacity monitor's saturation denominator (`capacity_alerts._pool_slots`) for the same reason.
Grid still routes a session to any free matching node, so it is *a spare node you can watch*, not
the node your session lands on; deterministic pinning is follow-up work.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -724,11 +724,23 @@ NGROK_JAEGER_PREFIX=your-app-jaeger
 SELENIUM_HUB_HOST=selenium-chrome
 SELENIUM_HUB_PORT=4444
 SELENIUM_KEEP_VIDEOS_X_DAYS=7
-# Selenium Grid (Phase 2, docker-compose.grid.yml — NOT enabled by default; see docs/SELENIUM_GRID.md).
-# Each node runs one session, so the node count IS the browser-session cap and MUST equal the summed
-# SELENIUM_CONCURRENCY of the lane workers (se_engage + se_prepost + se_outreach + se_content).
+# Selenium Grid (Phase 2, docker-compose.grid.yml — the DEPLOYED topology since 2026-07-27; see
+# docs/SELENIUM_GRID.md). `scripts/deploy.sh` composes the overlay in on every deploy unless this is
+# set to `standalone`, which is the one-flag rollback to the single selenium/standalone-chrome.
+SELENIUM_TOPOLOGY=grid
+# Each POOL node runs one session, so the node count IS the browser-session cap and MUST equal the
+# summed SELENIUM_CONCURRENCY of the lane workers (se_engage + se_prepost + se_outreach + se_content).
 # SELENIUM_HUB_HOST is overridden to `selenium-hub` by the overlay itself — leave it alone here.
 SELENIUM_GRID_NODES=8
+# The watchable debug node is a NINTH node on top of that pool, so it is excluded from the capacity
+# monitor's saturation denominator by the name it registers under. Empty disables the exclusion.
+SELENIUM_DEBUG_NODE_HOST=selenium-node-debug
+# Bind address for the debug node's noVNC (7900). Loopback like every other published port here.
+SELENIUM_GRID_VNC_BIND=127.0.0.1
+# The address an OPERATOR'S BROWSER reaches the Grid on (the tunnel hostname, e.g.
+# https://lemgrid.example.com). Only used to build the Grid UI's live-view link for the debug node;
+# unset just means that shortcut points at a compose-internal host.
+# SELENIUM_GRID_PUBLIC_URL=
 # Bind address for the Grid event bus (4442/4443). Loopback unless nodes run on a SECOND box, in
 # which case set the primary box's PRIVATE (WireGuard/VPC) address — never 0.0.0.0: registering a
 # node needs no credentials.
--- a/docker-compose.grid.yml
+++ b/docker-compose.grid.yml
@@ -9,10 +9,9 @@
 #
 #   docker compose -f docker-compose.yml -f docker-compose.prod.yml -f docker-compose.grid.yml up -d
 #
-# This is the DEPLOYED topology as of the 2026-07-27 cutover: `scripts/deploy.sh` composes this file
-# in on every deploy unless SELENIUM_TOPOLOGY=standalone is set (env, or the box's .env). It is still
-# an overlay, not the base — a bare `docker compose up` (dev) is unchanged and runs the standalone.
-# See docs/SELENIUM_GRID.md for the cutover runbook, the rollback, and the load-test evidence.
+# NOT enabled by default. Prod runs the standalone until the capacity monitor (§5e) says the cap is
+# the operating point; see docs/SELENIUM_GRID.md for the cutover runbook and the load-test evidence
+# that should precede it.
 #
 # INVARIANT (same one the standalone carries): total session slots == the sum of every lane
 # worker's SELENIUM_CONCURRENCY. Here total slots = SELENIUM_GRID_NODES x SE_NODE_MAX_SESSIONS(=1),
@@ -51,6 +50,15 @@ services:
       - SE_SESSION_REQUEST_TIMEOUT=300
       - SE_SESSION_RETRY_INTERVAL=5
     restart: unless-stopped
+    networks:
+      default:
+        aliases:
+          # Answer to the STANDALONE's name too. Everything that predates the Grid points at
+          # `selenium-chrome:4444` — the Cloudflare tunnel's lemgrid hostname (dashboard-managed,
+          # not in this repo), SELENIUM_HUB_HOST in the box .env, and host-side tooling. Without
+          # this alias the cutover 502s the Grid console and silently strips those callers of a
+          # hub. The standalone service is profile-parked, so the name is free.
+          - selenium-chrome
     deploy:
       resources:
         limits:
--- a/docker-compose.grid.yml
+++ b/docker-compose.grid.yml
@@ -31,8 +31,8 @@ services:
       # point for `ports` is that 4444 is not publicly reachable — a default of "all interfaces"
       # here would silently undo that hardening on the cutover.
       # NOTE: there is no 7900 on a hub. noVNC lives on the NODES, and replicas cannot publish the
-      # same host port, so "watch the live browser" is lost on a Grid — attach to one node instead:
-      #   docker compose port <node-container> 7900   (or run one node with an explicit mapping).
+      # same host port — which is why the pool is unwatchable and `selenium-node-debug` below is a
+      # separate, non-replicated node that DOES publish 7900.
       - "${SELENIUM_GRID_HUB_BIND:-127.0.0.1}:${SELENIUM_HUB_PORT}:4444"
       # Event bus. Only nodes speak these, and node-on-the-same-box reaches them over the compose
       # network without any publishing at all — so the default bind is loopback. Set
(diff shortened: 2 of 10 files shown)


gitchrisqueen:

- **Deploy-breaking `set -e` abort (critical)**: `SELENIUM_TOPOLOGY="${VAR:-$(grep -E '^SELENIUM_TOPOLOGY=' "$ENV_FILE" ...)}"` runs under `set -euo pipefail`. grep exits 1 when the key is absent — and it is absent on every box, because the key is new — so the command substitution failed the assignment and killed the script at that line, before `check_env.sh`, on **every** deploy (exit 2 with no `.env` at all). Reproduced under bash. Fixed by resolving through the script's existing `env_value` helper (always exits 0, returns the default), which required moving the block below that helper's definition; an unrecognized value now falls *forward* to `grid` with a warning instead of silently reverting the topology.
- **Eviction happened before the drain**: the transition guard `docker rm -f selenium-chrome` sat at step 3b, ahead of `maint begin`/`maint drain` — so it killed every in-flight Selenium session mid-task, which is exactly the mid-flight kill the #549 maintenance window exists to prevent. Moved to step 4c, after the drain; nothing before it starts a browser service, so the wait costs nothing.
- **No guard in the rollback direction**: with `SELENIUM_TOPOLOGY=standalone`, `selenium-hub` keeps holding 4444 and the standalone can't bind — the documented cutover failure with the roles swapped, i.e. the "one flag" fallback did not actually work. The guard now evicts whichever container the target topology replaces (hub ↔ standalone); the nodes are cleared by step 7's `--remove-orphans`.
- **No tests on the changed behavior** (the 42 that passed were pre-existing): added `TestDeployComposesTheGridTopology`, which **executes** deploy.sh's resolution prefix under real bash — a substring assertion on the source cannot see a `set -e` abort — covering no-`.env`, key-absent, `standalone` fallback, exported-value-wins, typo, plus guard ordering and both eviction directions.
- **Docs contradicted the change**: `docker-compose.grid.yml`'s own header and `docs/SELENIUM_GRID.md`'s status line still read "NOT enabled by default / Prod runs the standalone"; the rollback snippet omitted the `SELENIUM_TOPOLOGY=standalone` flag (so the next deploy silently undid the rollback) and the 4444 collision; and `docs/SELENIUM_DEBUGGING.md` + `DEPLOYMENT.md` still pointed on-call at a `selenium-chrome` container prod no longer runs. All corrected, with the noVNC consequence stated where an operator will hit it.

Checked and OK: the loopback/bus binds and the node-count == Σ-lane-concurrency invariant survive the cutover; `${COMPOSE}` is picked up by every rollback/standby path; `docker-compose.grid.yml` is present in the last three release tags, so the roll-back-one-release path can't reference a missing overlay; no `Closes #N` claim to verify.

Full unit suite: **7970 passed**.

gitchrisqueen:
🔎 Claude adversarial review — FIXED 6 findings

**The head commit reverted the previous review.** `64f9433` ("alias the Grid hub as selenium-chrome") was authored from a pre-review checkout: it landed the new alias but carried the stale copies of all six files with it, undoing `2f5cf77` in the same commit. Everything the earlier marker comment claimed fixed was live again at HEAD. Restored that content and re-applied the alias on top (`a8bd905`) — so this review is the earlier five findings *plus* what the alias itself needs.

Re-broken by the revert, now fixed again:

- **Deploy-breaking `set -e` abort (critical)**: `SELENIUM_TOPOLOGY="${VAR:-$(grep -E '^SELENIUM_TOPOLOGY=' "$ENV_FILE" ...)}"` runs under `set -euo pipefail`. grep exits 1 when the key is absent — and it is absent on every box, because the key is new — so the assignment failed and killed the script at that line, before `check_env.sh`, on **every** deploy. Re-reproduced under bash this pass: **exit 1** with a `.env` lacking the key, **exit 2** with no `.env`, the following line never reached. Resolved through the script's own `env_value` helper (sed-based, always exits 0); an unrecognized value falls *forward* to `grid` with a warning rather than silently reverting the topology.
- **Eviction happened before the drain**: `docker rm -f selenium-chrome` sat at step 3b, ahead of `maint begin`/`maint drain`, killing every in-flight Selenium session mid-task — the exact mid-flight kill the #549 window exists to prevent. Back at step 4c, after the drain.
- **No guard in the rollback direction**: with `SELENIUM_TOPOLOGY=standalone`, `selenium-hub` keeps holding 4444 and the standalone can't bind, so the "one flag" fallback did not work. The guard evicts whichever container the target topology replaces (hub ↔ standalone); `--remove-orphans` clears the nodes.
- **No tests on the changed behavior**: `TestDeployComposesTheGridTopology` **executes** deploy.sh's resolution prefix under real bash — a substring assertion on the source cannot see a `set -e` abort — covering no-`.env`, key-absent, `standalone` fallback, exported-value-wins, typo, plus guard ordering and both eviction directions.
- **Docs contradicted the change**: the overlay header and `docs/SELENIUM_GRID.md`'s status line read "NOT enabled by default / Prod runs the standalone"; the rollback snippet omitted the `SELENIUM_TOPOLOGY=standalone` flag (so the next deploy silently undid the rollback); `SELENIUM_DEBUGGING.md` + `DEPLOYMENT.md` pointed on-call at a container prod no longer runs.

New this pass, on the alias itself (kept — it is load-bearing):

- **The alias was untested and undocumented, and its interaction with the eviction guard was unstated.** Verified it is genuinely needed: `env_constants.py`'s `SELENIUM_HUB_HOST` default, `tools/selenium_mcp_server.py` and `tools/debug_linkedin_login.py` all address `selenium-chrome:4444`, and the tunnel hostname that 502'd isn't in this repo at all — so nothing in-tree fails when the alias is dropped. Added a test for it, documented that it is **DNS, not a container name** (`selenium-chrome:4444` resolves; `docker exec selenium-chrome` does not), and recorded that it is a *second* reason the eviction must happen: two containers answering one alias round-robins half the sessions into the wrong browser. Also corrected the overlay's own rollback comment, which told an operator to `--profile standalone up selenium-chrome` beside a live hub — that collides on both 4444 and the alias.

Checked and OK: `networks.default` resolves against the implicit default network (no file declares a top-level `networks:`), and the overlay parses with the compose `!override`/`!reset` tags; every compose invocation in `deploy.sh` goes through `${COMPOSE}`, so the overlay reaches the rollback/standby paths too; `env_value` is sed-based and cannot abort; the loopback/bus binds and the node-count == Σ-lane-concurrency invariant survive; no `Closes #N` claim to verify.

Tests: `tests/unit/app` + capacity alerts — **1695 passed** (28 in `test_selenium_capacity.py`, 8 of them new).

gitchrisqueen:

Reviewed the full diff plus the code at its seams (`capacity_alerts.py`, the base compose, `docs/SELENIUM_GRID.md`). No linked issue, so nothing to check on the close claim. Full unit suite green: **7976 passed**.

- **The 9th node silently desensitised the capacity signal** (the one that says "the cap is the constraint, buy the second box"): `collect_selenium_capacity` counts every slot the hub reports, so the denominator moved 8 → 9. A fully-claimed pool now reads 0.89 and the sample below it 7/9 = 0.78 — under the 0.85 `CAPACITY_SATURATION_PCT` that breached at 7/8 before. Added `_pool_slots()` to drop the debug node's slot, matched on the URI host it registers under. That required giving it one: it had **no `SE_NODE_HOST`**, so it registered as a per-restart container IP and nothing could identify it in `/status`.
- **The rollback direction had no transition guard** — the half this PR added it for. `SELENIUM_TOPOLOGY=standalone` left the hub running on 4444 (compose only reaps it as an orphan during the step-7 `up`), and the hub answers to the new `selenium-chrome` alias — so that name would resolve to two containers and workers could drive a hub with zero nodes. That path runs during an incident. Made it symmetric (`compose rm -sf` the hub + both node services first).
- **The topology was read by a second hand-rolled `.env` parser** (`grep | cut | tr`), 17 lines above the existing `env_value()` helper that already strips quotes, inline comments and stray CRs — so `SELENIUM_TOPOLOGY=standalone # rollback` resolved to neither value. And any value that wasn't exactly `grid` silently deployed the standalone with **no log line** — precisely the invisible drift this PR exists to end. Now uses `env_value`, normalises case, names an unrecognised value, and logs the resolved topology every deploy.
- **`SE_NODE_GRID_URL` was documented as "marks this node in the Grid UI"** — it doesn't; it's what the UI builds the node's live-view (noVNC) link from, so a compose-internal value points the operator's browser at a host only Docker can resolve. Now `${SELENIUM_GRID_PUBLIC_URL:-…}` with an accurate comment.

Also corrected the record around the last commit: the "standalone is retired by this overlay" comment had been orphaned above the new debug node (it read as describing it); `docs/SELENIUM_GRID.md` still said noVNC "goes away" and that the capacity monitor "needs no change"; `.env.example` still said the Grid was NOT enabled by default and documented none of the new knobs (`SELENIUM_TOPOLOGY`, `SELENIUM_GRID_VNC_BIND`, `SELENIUM_DEBUG_NODE_HOST`, `SELENIUM_GRID_PUBLIC_URL`); and the PR body's "Known consequence: noVNC is gone" contradicted the diff — updated.

New tests: pool exclusion (present / absent / disabled / no false-positive host match), the debug node's registered name pinned against `SELENIUM_DEBUG_NODE_HOST`, and `deploy.sh`'s topology composition, unrecognised-value handling, shared-parser use and both eviction guards.

Not changed, flagging as known: nine concurrent Chrome sessions is +1 vCPU / +1.5 GB / +2 GB shm over the §4 budget the box was sized for — that's the recorded owner decision, not a defect.

gitchrisqueen:

gitchrisqueen:

Issue #754, opened 2026-07-27
SPA new-version awareness: prompt to refresh from /api/app-info — Phase 2 (follow-up of #743)


gitchrisqueen:
## Why

Follow-up of #743, which shipped the two halves that stop a stale-chunk failure from being visible:
old-build asset retention (`api/spa_assets.py`) and a single silent reload with a loop guard
(`ui/src/utils/chunkReload.ts`). Both are **reactive** — they fire once something has already failed
to load.

The remaining piece is the third, explicitly optional item from #743:

> **3. Optional (nice-to-have): new-version awareness** — the SPA already has `/api/app-info`; a
> light poll could show 'New version available — refresh' proactively, turning a surprise error into
> an expected prompt.

Worth doing because releases batch **4x daily** (05/11/17/23 UTC): a long-lived tab is routinely
several builds behind, and a proactive prompt is a better experience than an invisible reload the
user did not ask for — and it also covers the case where a stale tab keeps *working* but is running
old client code against a newer API.

## Scope

- Poll `/api/app-info` on a slow interval (and on tab focus — a backgrounded tab should not poll).
- Compare against the version the bundle booted with; when it differs, render the same
  `NewVersionNotice` surface #743 added (do NOT add a second banner component) with a "Refresh" action.
- Never force a reload: this half is a PROMPT. The automatic reload stays owned by
  `chunkReload.ts`, and the two must not both fire.
- Keep it quiet when `/api/app-info` is unreachable or returns no version — an unknown version is
  never "new".

## Acceptance

- [x] A version change reported by `/api/app-info` surfaces the refresh prompt without any chunk
      having failed.
- [x] An identical version, an unreachable endpoint, or a missing/blank version surfaces nothing.
- [x] Polling is bounded (interval + no polling while the tab is hidden) — unit tested with fake timers.
- [x] The prompt reuses `NewVersionNotice`; no duplicate banner and no double-reload with #743's handler.
- [x] Unit tests (vitest) for the hook and the notice wiring.

Prior art in this repo: `docs/zero-downtime-deploys.md` § "Stale lazy chunks in already-open tabs".

PR #755, opened 2026-07-27
fix(ui): recover from stale lazy chunks after a deploy (closes #743)

### LINKED ISSUE #743: SPA breaks after every deploy: stale lazy-loaded chunks 404 ('Failed to fetch dynamically imported module') — needs reload-on-chunk-error + old-asset retention
## Reproduced from a real owner report (2026-07-27)
Clicking **Train New Avatar** failed with:
```
Failed to fetch dynamically imported module:
https://lem.christopherqueenconsulting.com/assets/jszip.min-BZWDyCXg.js
```

## Diagnosis — server side is healthy; this is the classic SPA-deploy chunk gap
- The requested hash `jszip.min-BZWDyCXg.js` **404s**; the build currently deployed ships `jszip.min-B8bT8jVo.js`, which **200s**.
- Both blue and green colors serve the SAME build (checked) — so this is NOT a blue/green mismatch.
- Cache headers are already correct: `index.html` is `no-store` (`cf-cache-status: BYPASS`), hashed assets are `immutable, max-age=31536000`.
- Cause: the browser tab was loaded BEFORE a deploy. Its in-memory module graph references the previous build's chunk hashes. A lazily-imported chunk (jszip, used by the avatar flow) is only fetched when the user triggers the feature — by which time the new image has replaced the old build's files, so the old hash is gone.

**A hard refresh fixes it, but users should never be expected to know that.** This is now a frequent failure: releases batch **4x daily** (05/11/17/23 UTC) and today alone shipped 4 versions. Any tab open across a release loses every not-yet-loaded lazy chunk — avatar training, carousel/PDF paths, charts, anything code-split. It presents as 'the feature is broken', not 'reload me'.

## Fix (both halves — they solve different windows)
**1. SPA: recover automatically from a chunk-load failure (essential, cheap)**
- Listen for Vite's `vite:preloadError` on `window` and for rejected dynamic imports; on failure, reload once to pick up the new `index.html` (which is `no-store`, so a reload always lands on the current build).
- Guard against reload loops: set a `sessionStorage` marker before reloading and, if it is already set, surface a clear message ('A new version was released — please refresh') instead of looping.
- Route-level `React.lazy` boundaries should use the same handler so a failed route chunk self-heals rather than showing a blank/error panel.

**2. Server: retain the previous builds' assets so in-flight sessions keep working (removes the reload entirely in most cases)**
- Keep the last N releases' `dist/assets` available so an old hash still resolves. Options for the agent to weigh: a shared assets volume that each deploy adds (not replaces) files into, with pruning beyond N releases; or an nginx `try_files`/fallback for `/assets/` misses into an archive directory. Must not break the immutable-cache contract (hashed filenames are unique, so serving an old hash is always safe).
- The deploy script (`scripts/deploy.sh`) is the natural place to sync/prune the archive; keep it idempotent and never let archive maintenance fail a deploy.

**3. Optional (nice-to-have): new-version awareness** — the SPA already has `/api/app-info`; a light poll could show 'New version available — refresh' proactively, turning a surprise error into an expected prompt.

## Acceptance
- Simulate the failure in a test (mock a dynamic import rejection) and assert exactly one reload attempt plus the loop guard.
- After the fix, a tab loaded pre-deploy that triggers a lazy feature post-deploy either loads the chunk (asset retention) or silently self-heals with one reload — no user-visible error.
- Unit tests for the handler; deploy-script change covered by the existing deploy tests.
- Verify no regression to the `no-store` index / immutable-asset headers.

## Context
Frequency is a direct consequence of the (good) 4x-daily zero-downtime release cadence — `docs/zero-downtime-deploys.md`. Prior related work: the Cloudflare stale-`index.html` fix (no-store) already shipped; this is the *chunk* half of the same problem class.

gitchrisqueen:
## What & why

A browser tab holds the **content-hashed chunk filenames of the build it loaded**, and a code-split
chunk (jszip in avatar training, any `React.lazy` route) is only fetched when the user triggers the
feature. Releases batch 4x daily, so a tab open across one asks for a hash the new image no longer
has — the request 404s and it presents as *"the feature is broken"*, not *"reload me"*. That is the
owner-reported `jszip.min-BZWDyCXg.js` failure in #743.

Two layers, because they cover different windows.

### 1. Asset retention — most users see nothing (`src/cqc_lem/api/spa_assets.py`)

Both prod colors mount a named `spa_asset_archive` volume at `SPA_ASSET_ARCHIVE_DIR`. Each FastAPI
container syncs its own `ui/dist/assets` into it at startup, keeps `SPA_ASSET_ARCHIVE_KEEP` builds
(default **5** — more than a day of releases) and serves a live-bundle miss out of the archive
(`ArchivedStaticFiles`). Serving an old hash is safe by construction: a content-hashed name resolves
to exactly one file forever, so the `immutable` cache contract is **preserved, not weakened**.

The archive is identified by the build's own file list, so a restart is not a new generation and a
**rollback re-deploy** of a build the archive already knows is not duplicated. Pruning only deletes
files no retained generation references, so a chunk unchanged across releases survives its first
build being pruned. Manifest is written before the prune, and both file copies and the manifest are
atomic (`os.replace`) — the other color may be serving that directory, and a half-written chunk the
browser caches `immutable` is worse than a missing one.

**Where it does NOT live: `deploy.sh`.** #743 offered a shared volume or an nginx archive fallback,
with deploy.sh syncing/pruning, and required that "archive maintenance never fail a deploy". Doing
it at container startup gets that property for free — it is idempotent, it runs on the new color
*before* the flip, it covers both colors and the rollback path with no shell, and a container that
cannot write the volume logs a warning and serves the live bundle only. So there is **no deploy-script
change** in this PR; the equivalent invariant (both colors mount the same named volume + set the env
var) is asserted in `TestProdComposeWiring` instead — nothing in the app fails when that wiring is
missing, retention just silently stops, and the only symptom would be this bug coming back.

### 2. One silent reload — the fallback for anything older (`src/cqc_lem/ui/src/utils/chunkReload.ts`)

`index.html` is `no-store`, so one reload always lands on the current build. The loop guard is a
`sessionStorage` marker: one reload per minute, and **a tab that cannot PERSIST the marker never
reloads at all** — with no way to remember an attempt there is no way to stop a loop, so it shows the
message instead. A blocked second failure dispatches to `NewVersionNotice` ("A new version was
released — please refresh" + a Refresh button).

react-query and error boundaries **catch** these rejections, so the window-level
`vite:preloadError` / `unhandledrejection` handlers alone would never see the jszip failure — hence
`importWithChunkRecovery` wrapping the avatar-training import, and `lazyWithChunkRecovery` for route
chunks. The error message every browser spells differently is matched on text (Chrome / Safari /
`ChunkLoadError` / `Unable to preload CSS`); an unrelated rejection is left entirely alone.

## Testing

- `tests/unit/api/test_spa_asset_archive.py` — 26 tests: sync, retention across builds, pruning
  (incl. a shared chunk surviving), restart/rollback idempotence, corrupt manifest, unusable archive
  (must never raise — the API has to start), path-traversal rejection, live-bundle vs archive
  serving, and the header contracts (`no-store` shell, `immutable` assets, incl. the archive
  fallback). Full suite: **8005 passed**.
- `src/cqc_lem/ui/src/utils/chunkReload.test.tsx` — 21 tests: exactly one reload then blocked, the
  marker surviving the reload that wipes module state, cooldown expiry, the storage-denied path,
  non-chunk errors untouched, both window listeners, `lazyWithChunkRecovery` rendering and
  recovering, and the notice. Full UI suite: **207 passed**; `npm run lint` clean on the touched
  files; `npm run build` green.

## Deploy note

The new `spa_asset_archive` volume is created by compose on the next deploy — no manual step, no new
key required in `/opt/lem/.env` (the env var is set in `docker-compose.prod.yml`, so `check_env.sh`
is unaffected). Retention starts with the first release that ships this, so the first deploy after
merge still has only one build archived.

## Scope

Closes #743 — every acceptance criterion is covered above. #743's item 3 was explicitly marked
*optional (nice-to-have)* and is not an acceptance criterion; it is filed rather than left in prose.

Follow-up: #754 (proactive "new version available" prompt from `/api/app-info`)


gitchrisqueen:

- **Offline tab reloaded away (`ui/src/utils/chunkReload.ts`)**: a dynamic import that fails because the tab is OFFLINE reports the *same* message a stale chunk does (`Failed to fetch dynamically imported module`), and `vite:preloadError` names no cause at all — so both entry points treated a dropped connection as a deploy. `index.html` is `no-store`, so that reload cannot re-fetch the shell: it replaced a working app with the browser's offline page, strictly worse than the error it was recovering from. Fixed: `navigator.onLine === false` (the only direction that flag is trustworthy) now returns `ignored`, so the original rejection goes back to the caller's error UI (the avatar-training mutation renders it) and heals itself when the connection returns, and the preloadError handler leaves Vite's default rethrow in place exactly as it did pre-#743. +2 tests (209 UI tests pass), `docs/zero-downtime-deploys.md` + the CLAUDE.md bullet updated.

Verified, no finding:
- **Path traversal** — `archived_asset_path` realpath-containment holds behind Starlette's own `normpath`; the manifest is unservable; `..`/absolute/backslash/empty-segment all rejected.
- **Header contracts** — archive fallback still `immutable`, shell still `no-store`; probed a real request: HEAD returns headers-only and `Range` returns 206 through `FileResponse`, and nested `assets/sub/*.js` archives and serves (both work, neither is in the suite).
- **A genuinely missing chunk still 404s** and does NOT fall through to the SPA catch-all — a Mount matches FULL, so the browser never gets HTML for a JS import.
- **Volume permissions** — the #155 assets-volume class of bug does not apply: `web_api_*` runs as root, so the new root-owned named volume is writable. Both colors inherit the mount + env via the `&web_api` anchor.
- **Deploy preflight** — `check_env.sh` reads `.env.prod.example`, not the `.env.example` this PR adds keys to, so it cannot fail a deploy.
- **The archive actually gets the request** — the prod nginx edge proxies `/` wholesale to the active color; nothing serves `/assets` ahead of FastAPI.
- **Rollback ordering** — re-syncing a previously-archived build re-promotes it to newest, which is correct: after `rollback.sh` that build IS live.
- **Close honesty** — `Closes #743` is covered; the optional item 3 is filed as #754, linked in the body and commented on the issue.

Unrelated to this PR: `E2E Tests` fails on `redis.exceptions.ConnectionError: Error 111 connecting to localhost:6379` in `test_carousel_video_post_e2e.py` — an environment gap in a non-required check, untouched by this diff.

gitchrisqueen:

PR #761, opened 2026-07-28
fix(observability): attribute LLM spend to serving model, add shadow cost, vendor price map (closes #752)

### LINKED ISSUE #752: LLM cost attribution is wrong: spend is booked against the tier ALIAS, not the serving model — and there is no shadow cost for subscription traffic
## Two defects, one root
Research 2026-07-27 (LiteLLM model cost map) checked LEM's accounting against the live `$ai_generation` stream and found:

### 1. Spend is attributed to the ALIAS, not the model that actually ran
`ai_helper.py:106` passes `model=model` to `track_llm_call` — but `model` is the **tier alias** (`lem-complex`), never the deployment that served the call after fallback or cost-aware down-routing. `_cost_table()` is therefore keyed on aliases with roughly OpenAI list rates:

| alias | booked in/1K | booked out/1K |
|---|---|---|
| lem-simple | 0.00015 | 0.00060 |
| lem-medium | 0.00060 | 0.00240 |
| lem-complex | 0.00300 | 0.01500 |

Against the models that actually serve those tiers today (priced at the cheapest metered reference in LiteLLM's map), LEM **over-books the Ollama tiers ~2-5×** (e.g. `qwen3.5:397b` ≈ 0.00060/0.00360 vs `lem-complex`'s 0.00300/0.01500 — 5× / 4.2× high). Consequences: the margin report, `cost_ledger`, and every per-feature cost insight are wrong; a fallback to a paid provider is invisible; and the cost-routing experiment (#494) cannot measure its own savings because both arms book the same alias rate.

### 2. Subscription traffic reports $0, so there is no shadow cost
LiteLLM prices from `model_prices_and_context_window.json`, which lists Ollama Cloud models at `0.0`. Verified on the live project (`$ai_generation`, 14 days): **every Ollama call reports $0.00** across ~64k tokens, while `gpt-4o-mini` and embeddings price correctly. $0 is *arguably right* as marginal cost on a flat-rate subscription — but it means nothing in the stack can answer **'what would our usage cost if we left the subscription?'**, which is the entire basis of that decision.

Same 14 days, three different numbers: LiteLLM **$0.0000** · LEM ledger **$0.2066** · realistic shadow **$0.0568**.

## Scope
1. **Attribute to the serving model.** Capture the model LiteLLM actually used (`$ai_model` / response `model` / `_hidden_params`) and pass THAT to `track_llm_call`; keep the requested alias as a separate dimension (`tier`) so per-tier reporting survives. Backfill is out of scope — fix forward.
2. **Vendor the cost map.** Commit a dated snapshot of `model_prices_and_context_window.json` (MIT) or a slim extract of the ~20 models LEM can touch; refresh via PR so price changes are a visible diff. Do NOT fetch live at runtime, and do NOT add `input_cost_per_token` to the Ollama deployments in `.litellm/config.yaml` — that would corrupt `response_cost`, the one number that is currently TRUE (real marginal spend).
3. **Add shadow cost as a distinct signal.** Map each Ollama model to a hand-chosen metered reference (document the choice — it is editorial, not fact) and emit `shadow_cost_usd` as its own property/column that no existing insight sums. Label it 'shadow cost @ <provider> list', never 'cost'. `docs/llm-analytics.md` already warns about two spend streams; this makes three — the doc must state which is which and which one answers which question.
4. **Fix `_cost_table()`** to derive from the vendored map keyed on the serving model, with the current hardcoded values kept only as a last-resort fallback for models absent from the map.
5. **Consider `deprecation_date`** from the same map as an extra signal for the weekly model-retirement cron (#716) — it covers hosted models, not Ollama Cloud.

## Acceptance
- Unit tests: serving-model extraction (incl. fallback and down-routed calls), shadow-cost computation, map-miss fallback, and a test proving `shadow_cost_usd` never lands in `cost_ledger.cost_usd`.
- A dated snapshot committed with its source URL and fetch date.
- `docs/cost-performance-margin-plan.md` + `docs/llm-analytics.md` updated to explain the three streams (real marginal, booked, shadow) and which decision each supports.

## Out of scope
Routing the agent pipeline's `claude -p` through LiteLLM — researched and rejected: the Max-subscription passthrough is unsupported by Anthropic and its failure mode silently converts a flat-rate pipeline to metered billing. Per-agent-run token capture, if wanted later, comes from `claude -p --output-format json` (the runner already captures output) priced with this same map.

gitchrisqueen:
## What & why
Issue #752 found that LEM's LLM cost attribution was booked against the requested tier alias (e.g. ) rather than the model LiteLLM actually served after fallback or cost-aware down-routing, and that Ollama Cloud subscription traffic reported $0 with no shadow cost for the "what if we left?" decision.

This PR:
- Captures the serving model from  /  in  and passes it to  as , while preserving the requested alias as .
- Prices  by the serving model via a vendored  (LiteLLM price map + LEM shadow references), with legacy tier aliases as fallback.
- Emits  for subscription-only Ollama Cloud models, explicitly keeping it out of .
- Adds  and updates  to use the snapshot.
- Updates  and  to document the three cost streams (real marginal / booked / shadow) and which decision each supports.

## Testing
-  — 150 passed.
- New tests cover serving-model extraction (response.model, hidden params, fallback, down-routed), shadow-cost computation, map-miss fallback, and a proof that  never lands in .

Closes #752

### DIFF
--- a/.litellm/model_prices_snapshot.json
+++ b/.litellm/model_prices_snapshot.json
@@ -1,0 +1,171 @@
+{
+  "_comment": "Committed snapshot of BerriAI/litellm model_prices_and_context_window.json plus LEM shadow-reference mappings for Ollama Cloud models. Source: https://github.com/BerriAI/litellm/blob/main/model_prices_and_context_window.json Fetched: 2026-07-28",
+  "models": {
+    "anthropic/claude-sonnet-4-6": {
+      "input_cost_per_token": 3e-06,
+      "output_cost_per_token": 1.5e-05,
+      "deprecation_date": null,
+      "mode": "chat"
+    },
+    "lem-complex": {
+      "input_cost_per_token": 3e-06,
+      "output_cost_per_token": 1.5e-05,
+      "mode": "chat"
+    },
+    "lem-embedding": {
+      "input_cost_per_token": 2e-08,
+      "output_cost_per_token": 0.0,
+      "mode": "embedding"
+    },
+    "lem-image": {
+      "input_cost_per_token": 0.0,
+      "output_cost_per_token": 0.0,
+      "mode": "image_generation"
+    },
+    "lem-medium": {
+      "input_cost_per_token": 6e-07,
+      "output_cost_per_token": 2.4e-06,
+      "mode": "chat"
+    },
+    "lem-research": {
+      "input_cost_per_token": 1e-06,
+      "output_cost_per_token": 1e-06,
+      "mode": "chat"
+    },
+    "lem-router": {
+      "input_cost_per_token": 6e-07,
+      "output_cost_per_token": 2.4e-06,
+      "mode": "chat"
+    },
+    "lem-simple": {
+      "input_cost_per_token": 1.5e-07,
+      "output_cost_per_token": 6e-07,
+      "mode": "chat"
+    },
+    "lem-tts": {
+      "input_cost_per_token": 0.0,
+      "output_cost_per_token": 0.0,
+      "mode": "audio_generation"
+    },
+    "openai/dall-e-3": {
+      "input_cost_per_token": null,
+      "output_cost_per_token": null,
+      "deprecation_date": null,
+      "mode": "image_generation"
+    },
+    "openai/deepseek-v4-flash": {
+      "input_cost_per_token": 0.0,
+      "output_cost_per_token": 0.0,
+      "deprecation_date": null,
+      "mode": "chat",
+      "shadow_reference": "openai/gpt-4o-mini"
+    },
+    "openai/deepseek-v4-pro": {
+      "input_cost_per_token": 0.0,
+      "output_cost_per_token": 0.0,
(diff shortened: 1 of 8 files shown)


gitchrisqueen:
- Serving model attribution correctly implemented: track_llm_call now receives both requested alias (as model_tier) and actual serving model, with costs calculated by the serving model
- Shadow cost properly implemented: estimate_shadow_cost_usd computes shadow cost for zero-priced models and emits as separate property (never added to cost ledger)
- Cost table correctly sources from vendored map: _cost_table() uses model_prices_snapshot.json as primary source with legacy tier fallbacks
- Vendored price map includes source URL and fetch date: .litellm/model_prices_snapshot.json contains proper attribution
- Documentation updated: cost-performance-margin-plan.md and llm-analytics.md explain the three cost streams and their use cases
All acceptance criteria met. Implementation includes appropriate fallbacks and error handling.

gitchrisqueen:

PR #819, opened 2026-07-31
fix(scheduling): never convert a picked time against a guessed timezone (closes #774)

### LINKED ISSUE #774: fix(scheduling): Scheduled post published at incorrect time
## Why
Post #33 was scheduled for 9am EST today yet somehow went out at 5am EST instead. This is the wrong time.

Auto-filed from in-app feedback #10 by the feedback→auto-work loop (`docs/launch-and-marketing-plan.md` §B.2/B.3). Classifier: bug/high, risk `none`, confidence 0.87.

## Scope
- Address the reported bug in the `scheduling` area of the product.

## Files
- `tests/unit/`

## Acceptance
- All required CI gates pass.

gitchrisqueen:
## What happened

Post 33 was scheduled for 9am ET and went out at 5am ET.

**The scheduler was not at fault.** Production logs show it honored the stored instant exactly:

```
2026-07-28T08:40:00Z  Post 33 queued for 2026-07-28 09:00:00+00:00
2026-07-28T09:02:19Z  Pre-post engagement pass complete
2026-07-28T09:22:20Z  Golden-hour reply_sweep on post 33: … 22 min after publish
```

`09:00Z` **is** 5am ET. The stored time was wrong: the wall clock the user picked was read in a
zone that wasn't theirs. Post 37, two days later, stored the same 9am pick correctly as `13:00Z` —
same user, same input, two different interpreting zones.

## Why

`useUserTimezone()` silently substituted `Intl.DateTimeFormat().resolvedOptions().timeZone` — the
**browser's** zone — whenever the user's own zone wasn't available: while `/user/timezone` is in
flight, before the session token hydrates, or if the request fails. Every scheduling surface
converts the typed wall clock with whatever that hook returned at that moment, so a browser zone
that differs from the configured one (travel, VPN, a UTC machine, a slow request) stores the post
offset by the difference.

Rendering a value in the wrong zone for a few milliseconds is cosmetic. **Converting an input
against it is not** — that instant is persisted and published. And ContentStudio's two post pickers
carried no zone label at all, so "09:00" was unfalsifiable to the user.

## What changed

- **`useUserTimezoneState()`** returns `{ timezone, isResolved }`; `isResolved` is true only when
  `/user/timezone` actually answered. `useUserTimezone()` still returns the bare string for
  display-only callers (Dashboard, list rows) where a brief fallback is harmless.
- **Convert-on-change pickers** (ContentStudio editor, NewsletterQueue) are **locked** until the
  zone is known — their `onChange` writes a converted UTC instant into state.
- **Convert-on-submit pickers** (ComposePost, ContentStudio bulk reschedule, ScheduledDMs) **refuse
  the write** and say why.
- **Every picker names the zone it is read in.** Unresolved reads `loading your timezone…` rather
  than naming the guess. The query now retries, so a transient blip can't leave scheduling locked.
- **`_warn_if_naive_schedule()`** logs a WARNING when `/schedule_post/`, `/update_post/` or
  `/posts/bulk_update/` receive an offset-less datetime. It is still interpreted as UTC (that's the
  contract, and legacy callers depend on it) — just no longer *silently*, so the next occurrence is
  visible in logs instead of only in a mistimed post.
- `docs/timezone-contract.md` §4 records the rule.

## Testing

- `src/hooks/useUserTimezone.test.tsx` — 5 cases: in-flight, resolved, failed, no session token,
  and the display-only string hook.
- `src/pages/ContentStudio.timezone.test.tsx` — 3 cases pinned to an unresolved zone: bulk
  reschedule blocked, label says loading (never `UTC`), editor time field locked and empty.
- `tests/unit/api/test_schedule_timezone_breadcrumb.py` — 9 cases across the helper and all three
  endpoints, asserting naive warns and explicit-UTC / absent stays quiet.
- `poetry run pytest tests/unit -q` → **8218 passed**. `tsc -b` clean.
- **vitest could not be run locally** (this box has Node 18; CI uses Node 22 and runs `npm test` in
  the UI Build job) — the new SPA tests are verified by CI, not by me.

## Note on existing rows

This prevents recurrence; it does not retro-correct rows already stored under a wrong zone (post 33
included). Which rows those are isn't recoverable — nothing recorded the zone each write was
converted with — so a blanket re-localize would move correctly-stored posts too. Affected posts
should be rescheduled by hand; the pickers now show the zone, so the value is checkable.

Closes #774

### DIFF
--- a/docs/timezone-contract.md
+++ b/docs/timezone-contract.md
@@ -65,6 +65,22 @@ only the last-resort fallback while the query is in flight. A user whose Login L
 their device (travel, VPN, or deliberately targeting an audience in another zone) must still see and
 set times in their configured zone.
 
+**Never convert against the fallback.** Rendering a value in the browser's zone for the few
+milliseconds before the real one lands is cosmetic; converting a wall clock the user *typed* against
+it is not — that instant is stored, and the post fires by the difference between the two zones
+(issue #774: a post picked for 9am ET was stored as `09:00Z` and published at 5am ET). So every
+surface that WRITES a scheduled time reads `useUserTimezoneState()` instead, which reports
+`isResolved: false` until `/user/timezone` has actually answered — in flight, disabled because the
+session token hasn't hydrated, or failed. While it is false the picker is disabled, labelled
+`loading your timezone…` rather than named with the guess, and the submit path refuses. `ComposePost`,
+the `ContentStudio` editor and bulk-reschedule bar, `ScheduledDMs` and `NewsletterQueue` all hold to
+this; `useUserTimezone()` (bare string, browser fallback) is for display-only callers.
+
+Server-side, a scheduling write that arrives with **no** offset is still interpreted as UTC — that is
+the contract and legacy callers rely on it — but `api.main._warn_if_naive_schedule()` logs a WARNING
+on `/schedule_post/`, `/update_post/` and `/posts/bulk_update/` so the next occurrence is diagnosable
+from logs rather than only from a post that went out at the wrong hour.
+
 `src/utils/datetime.ts` owns all three conversions:
 
 | Helper | Direction |
--- a/src/cqc_lem/api/main.py
+++ b/src/cqc_lem/api/main.py
@@ -266,6 +266,21 @@ def _utc_iso(dt) -> Optional[str]:
     return dt.isoformat().replace("+00:00", "Z")
 
 
+def _warn_if_naive_schedule(dt, endpoint: str, **context) -> None:
+    """A scheduling write whose datetime carries NO offset is a client bug, and a silent one: the
+    storage layer assumes naive means UTC (db.to_naive_utc), so a wall clock the user picked in
+    their own zone is stored verbatim and the post fires offset-hours away from the time they saw
+    (issue #774 — a 9am post published at 5am). We keep interpreting it as UTC (the contract, and
+    what every legacy caller relies on) but leave a breadcrumb, because today the only evidence is
+    the wrong publish time itself."""
+    if dt is not None and getattr(dt, "tzinfo", None) is None:
+        log_warning(
+            f"Naive scheduled_datetime received by {endpoint} — assuming UTC "
+            f"(clients must send an explicit-UTC ISO string; see docs/timezone-contract.md)",
+            **context,
+        )
+
+
 def _public_post_url(value) -> Optional[str]:
     """Only surface real http(s) permalinks. Home-feed comments have no LinkedIn permalink and are
     logged under a synthetic 'feedpost://<hash>' dedup key — never expose that raw string to the UI."""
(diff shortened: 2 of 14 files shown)


gitchrisqueen:

Reviewed the diff against #774's acceptance criteria and the surrounding scheduling code (all five `zonedInputToUtcIso`/`zonedDateToUtc*` call sites, `AuthContext` session hydration, `to_naive_utc`, `_utc_iso`, and the `/schedule_post/` `/update_post/` `/posts/bulk_update/` paths). The diagnosis holds up: `useUserTimezone()` was the only thing standing between a typed wall clock and a stored instant, and a failed or in-flight `/user/timezone` silently substituted the browser zone. Every surface that WRITES a scheduled time is now covered — `ComposePost`, the ContentStudio editor + bulk bar, `NewsletterQueue`, `ScheduledDMs` — and nothing else in the SPA converts user input against a zone. The remaining `useUserTimezone()` callers (Dashboard, list rows, `ConnectionRequests`) are display-only, which is correct.

Fixed (commit b3611b8):

- **`retry: 2` reduced resilience instead of adding it** (`hooks/useUserTimezone.ts`): react-query's own default for queries is `3`, so the new option — added with the comment "a transient blip must not leave the pickers locked" — cut the attempt count from 4 to 3 and made a blip *more* likely to lock scheduling. Pinned at `retry: 3` with the reason stated, and the hook test's expected attempt count updated to match.
- **Two of the four guarded write surfaces had no test** (`ComposePost`, `ScheduledDMs`). `ComposePost` is the surface the reported post was scheduled from, and its guard is genuinely reachable — a session-token change re-keys the query, so `isResolved` can go false with a time already typed. Added `ComposePost.timezone.test.tsx` (picker locked + labelled `loading your timezone…` rather than the browser guess; a time typed while resolved is refused after the zone is lost, with no POST) and `ScheduledDMs.timezone.test.tsx` (its picker is *not* disabled, so the buttons are the only gate — asserted disabled with a complete draft and no POST, plus the positive control that 09:00 New York becomes `13:00Z` once the zone is known). `NewsletterQueue`'s guard is the same disabled-input + blank-value shape the ContentStudio editor test already covers, so it is not duplicated.

Checked and not findings:

- `Closes #774` is honest — the issue carries no later phase or unchecked box beyond "behavior fixed / unit tests / CI", and the PR body names the one thing deliberately out of scope (existing rows are not retro-corrected, and says why a blanket re-localize would be wrong).
- `to_naive_utc` and `_utc_iso` both round-trip correctly; the SPA never emits an offset-less datetime today, so `_warn_if_naive_schedule` really is a breadcrumb rather than a papered-over bug, and it stays quiet on the aware and absent cases.
- CLAUDE.md conventions: `log_warning` from `utilities.logger` with structured `user_id`/`post_id` context, no raw SQL, no new migration.

`poetry run pytest tests/unit/api/test_schedule_timezone_breadcrumb.py -q` → 9 passed locally. The SPA tests (existing and new) are verified by CI — this box is on Node 18 and the UI toolchain needs Node 20+.

gitchrisqueen:

Issue #792, opened 2026-07-29
fix(dms): Catch-up feature not sending any messages


gitchrisqueen:
## Why
The user reports that the auto catch‑up feature set to use the LinkedIn default message and send a certain number of messages daily is not sending any messages, not logging them, and not appearing in any queue or activity dashboard. They request verification and debugging with the admin account (user_id=1).

Reported by 1 distinct user(s) across 1 feedback item(s).

Auto-filed from in-app feedback #13 by the feedback→auto-work loop (`docs/launch-and-marketing-plan.md` §B.2/B.3). Classifier: bug/high, risk `none`, confidence 0.92.

## Scope
- Address the reported bug in the `dms` area of the product.
- Keep the change minimal and scoped to this report — no unrelated refactors.

## Files
Likely starting points (verify before editing; add/extend the unit tests):
- `src/cqc_lem/app/run_automation.py`
- `src/cqc_lem/utilities/ai/dm_nurture.py`
- `tests/unit/`

## Acceptance
- The behavior the reporter described is fixed/implemented.
- Unit tests cover the new/changed logic (≥80% patch coverage).
- All required CI gates pass.

gitchrisqueen:
PR #825 is up — it does **not** close this issue.

**What it fixes**
1. **The lane is now observable.** Every catch-up run of both phases emits ONE `catchup_run` PostHog event, including the runs that draft or send nothing: `status` (`disabled` / `throttled` / `session_failed` / `scrape_failed` / `no_moments` / `none_qualified` / `drafted`, plus `dispatched` / `capped` / `nothing_to_send` on the send drip) and the per-stage funnel `moments → classified → enabled_type → excluded / duplicate / below_bar → drafted`. Before this, all seven of those outcomes were the same silence, which is why "not sending, not logging, not in any queue" had no answer.
2. **A real message defect on the exact path reported.** `button[aria-label*='congrats']` matches LinkedIn's OWN "Say congrats to Jane" trigger, and an empty composer renders its placeholder — both cleared the 4-char floor, so the card's *chrome* became the congratulations. With `catchup_message_source='linkedin'` (the reported setting) a touch could be queued — and on `auto_approve`, sent — as literally `"Say congrats"`. Now rejected; the per-milestone one-liner is used instead.

**What's left here (why this stays open)**
[...]

PR #825, opened 2026-07-31
fix(dms): make every LinkedIn Catch-up run reportable (refs #792)

gitchrisqueen:
## Why

A user reported that the auto catch-up feature — configured to use LinkedIn's default message and send N messages a day — **isn't sending anything, isn't logging anything, and doesn't show up in any queue or activity dashboard**, and asked for a debugging pass on the admin account.

The lane was **write-only**. Every one of these produced exactly the same output — silence:

- the 429 breaker / automation pause held the beat, so no scan ever dispatched
- the browser session died before the feed loaded
- the scrape threw
- the feed rendered cards but none classified as a milestone (a selector or copy drift)
- moments classified but every one was excluded / already touched / below `CATCHUP_MIN_SCORE`
- touches drafted fine and sat `pending` behind approval, or `approved` behind a spent daily cap

There was no telemetry that could tell those apart, so the report was unanswerable.

## What changed

**1. Every catch-up run reports — including the ones that do nothing.**

`track_catchup_run` (`observability.py`) emits ONE `catchup_run` event per run of **both** phases, carrying:

- `phase` — `scan` (drafting) | `send` (the drip that dispatches) | `deliver` (the per-touch terminal outcome)
- `status` — `disabled` / `throttled` / `session_failed` / `scrape_failed` / `no_moments` / `none_qualified` / `drafted` on the scan; `dispatched` / `capped` / `inactive_users` / `nothing_to_send` on the drip; `sent` / `failed` / `capped` / `dm_capped` / `throttled` / `no_message` / `not_sendable` on delivery
- the **per-stage funnel**: `moments` → `classified` → `enabled_type` → `excluded` / `duplicate` / `below_bar` → `drafted`, plus `dispatched` / `capped` / `inactive` / `requeued`
- `auto_approve` and `message_source`, so an expectation mismatch (`pre_review` + a per-day cap reads as "it should just send") is visible too

**`dispatched` is not `sent`**, which is why the `deliver` phase exists. Every deferral in `send_catchup_touch` (the account-wide DM cap, the catch-up cap, a 429) puts the row back to `approved`, so the 20-minute drip re-dispatches the same touch on the next beat. A user whose `max_dms_per_day` is already spent by the other DM lanes produces a healthy, climbing `dispatched` count on a lane that has delivered nothing all day — the reported symptom exactly. Only the terminal report separates the two.

`moments > 0` with `classified == 0` is a selector/copy drift. `classified > 0` with `drafted == 0` is exclusion, dedup or the score bar — and the counts say which. This is the same posture as `company_page_invite_run` (#732), for the same reason: a series that only carries sends cannot distinguish "quiet today" from "silently broken".

The steady-state outcomes (`nothing_to_send`, `capped`, `dm_capped`, `throttled`, `disabled`, `not_sendable`) log **DEBUG** so an expected no-op never re-emits at ERROR under the recurrence escalation — the PostHog series is what carries them. Two pre-existing `log_warning`s on the same lane were doing exactly that and are now DEBUG too: the empty-feed warning fires on every quiet scan, and "user not active/connected" fires every 20 minutes per queued touch, so both crossed the escalation threshold and filed a grouped `$exception` for a no-op. Approved touches for a disconnected owner were also counted at no stage, so the drip reported `nothing_to_send` while a real queue sat behind a broken session; they now count as `inactive`.

**2. LinkedIn's button chrome is no longer mistaken for its draft.**

`_CATCHUP_SUGGESTED_TEXT_LOCATORS` includes `button[aria-label*='congrats']`, which matches LinkedIn's **own** "Say congrats to Jane" trigger, and an empty composer renders its placeholder. Both clear `_clean_suggested_message`'s 4-character floor, so the card's *chrome* became the congratulations — queued as `"Say congrats"`, and on `catchup_touch_mode='auto_approve'` **sent** as it. This is exactly the `catchup_message_source='linkedin'` path the reporter configured. `_clean_suggested_message` now rejects the affordance labels and placeholders, so the per-milestone one-liner is used instead — which is what its docstring already claimed it did.

## Testing

`tests/unit/app/test_catchup_touches.py` (+41) and `tests/unit/utilities/test_observability.py` (+3):

- a report on every scan exit path — drafted, empty feed, unclassifiable cards, duplicate, excluded, below-bar, disabled, throttled, dead session, scrape failure
- both beat dispatchers report when throttled, when nobody is enabled, when the cap swallowed the queue, when the queue is stuck behind a disconnected account, and when the queue is empty
- a report on every `send_catchup_touch` exit — sent, failed, both caps (kept distinguishable from each other), throttled, empty body, vanished row
- a telemetry outage never fails a run that otherwise worked
- 13 chrome labels rejected, 4 real drafts preserved, plus the end-to-end fallback to the plain congratulations

`poetry run pytest tests/unit -q` → **8513 passed** locally.

## Remaining on #792

This PR makes the failure **diagnosable** and fixes the message defect on the reporter's exact configuration, but it does **not** close the issue: #792 asks for verification and debugging against the admin account (`user_id=1`), which needs real credentials and a live Selenium session I can't run headless. After this ships, the `catchup_run` series answers it directly — filter `user_id = 1` and read `status` + the funnel counts on `phase = 'scan'` (was anything drafted?) and `phase = 'deliver'` (did anything actually go out, and if not, which gate stopped it?). I've left a comment on #792 with that lookup.

### DIFF
--- a/src/cqc_lem/app/run_automation.py
+++ b/src/cqc_lem/app/run_automation.py
@@ -88,7 +88,7 @@ from cqc_lem.utilities.linkedin_formatter import normalize_public_text
 from cqc_lem.utilities.logger import myprint, log_error, log_info, log_warning, log_debug
 from cqc_lem.utilities.observability import track_post_outcome, track_audience_snapshot, \
     track_comment_outcome, track_golden_hour_report, track_company_page_invite_run, \
-    attribute_llm_cost, llm_attribution, \
+    track_catchup_run, attribute_llm_cost, llm_attribution, \
     FEATURE_COMMENT, FEATURE_CONTENT, FEATURE_DM
 from cqc_lem.utilities.selenium_util import click_element_wait_retry, \
     get_element_wait_retry, get_elements_as_list_wait_stale, getText, close_tab, get_driver_wait_pair, quit_gracefully, \
@@ -5734,6 +5734,22 @@ CATCHUP_URL = "https://www.linkedin.com/mynetwork/catch-up/all/"
 # generic "Congrats!" is worse than saying nothing.
 CATCHUP_MIN_SCORE = int(os.getenv("CATCHUP_MIN_SCORE", "30"))
 
+# Why a catch-up run produced what it produced (issue #792). Every run reports one of these — the
+# quiet outcomes especially, because "the feed had no milestone today", "the 429 breaker never let
+# the scan start" and "the card selectors have drifted" are the SAME silence without them.
+CATCHUP_PHASE_SCAN = "scan"
+CATCHUP_PHASE_SEND = "send"
+CATCHUP_STATUS_DRAFTED = "drafted"
+CATCHUP_STATUS_DISABLED = "disabled"                # no milestone type enabled for this user
+CATCHUP_STATUS_THROTTLED = "throttled"              # 429 breaker / automation pause
+CATCHUP_STATUS_SESSION_FAILED = "session_failed"
+CATCHUP_STATUS_SCRAPE_FAILED = "scrape_failed"
+CATCHUP_STATUS_NO_MOMENTS = "no_moments"            # the feed rendered nothing we could read
+CATCHUP_STATUS_NONE_QUALIFIED = "none_qualified"    # moments read, none survived the funnel
+CATCHUP_STATUS_DISPATCHED = "dispatched"
+CATCHUP_STATUS_NOTHING_TO_SEND = "nothing_to_send"
+CATCHUP_STATUS_CAPPED = "capped"                    # approved touches exist, today's cap is spent
+
 # Ordered fallback chain for the catch-up cards. LinkedIn's SDUI ships hashed classes, so we prefer
 # data-view-name, then the semantic list, then any main-content list item that links to a profile.
 # NOTE: needs a supervised live-grounding pass on a real account before broad enablement (see #478).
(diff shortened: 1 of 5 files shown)


gitchrisqueen:

- **`dispatched` was never `sent`, and the terminal stage reported nothing.** `send_catchup_touch` emitted no telemetry on any of its exits, and every deferral there (the account-wide DM cap, the catch-up cap, a 429) puts the row back to `approved` — so the 20-minute drip re-dispatches the same touch on the next beat. A user whose `max_dms_per_day` is already spent by the other DM lanes produces a healthy, climbing `dispatched` count on a lane that has delivered nothing all day, which is the reported symptom this PR exists to answer. Added a third phase, `deliver`, carrying the per-touch outcome (`sent` / `failed` / `capped` / `dm_capped` / `throttled` / `no_message` / `not_sendable`) + `touch_id`; 6 new tests, including that the two caps stay distinguishable from each other.
- **The send drip lost the inactive-user cohort entirely.** Approved touches whose owner is no longer connected were counted at no funnel stage, so the report read `nothing_to_send` while a real queue sat behind a broken session. Now counted as `inactive` with its own `inactive_users` status; test added.
- **Two expected no-ops on this lane warned, so the recurrence escalation re-emits them at ERROR and files a grouped `$exception`.** "Catch-up feed produced no moments" fires on every quiet scan, and "user not active/connected" fires every 20 minutes per queued touch — both cross the 3-in-window threshold trivially. That is the exact defect the PR's own `_CATCHUP_QUIET_STATUSES` was added to avoid, so both are DEBUG now and the run report carries them durably. Existing test updated to assert the level.

Also checked and found clean: the `_CATCHUP_CHROME_RE` fullmatch does not eat real drafts (`Congrats to Jane on the new role!`, `Congratulations on the promotion — well earned.` all survive, and a rejected chip falls back to a strictly better per-milestone one-liner); the funnel counters are consistent with the exclusion order in the scan loop; `track_catchup_run` matches the `track_company_page_invite_run` shape; the fleet dispatchers correctly attribute to `system`; and the close is honest — the PR carries no closing keyword and states what remains on #792.

PR body and the #792 lookup comment updated to document the `deliver` phase. `poetry run pytest tests/unit -q` → 8513 passed.

gitchrisqueen:

PR #836, opened 2026-07-31
feat(feedback): admin panel for feedback triage (closes #793)

### LINKED ISSUE #793: feat(other): Implement admin panel for feedback triage
## Why
There should be an admin panel in LEM that allows designated admin users to manage feedback submissions. Only feedback from admin users should be auto-triaged into the GitHub development cycle, while other feedback should require admin review before acceptance. This feature would enhance the management of feedback within LEM and GitHub.

Auto-filed from in-app feedback #16 by the feedback→auto-work loop (`docs/launch-and-marketing-plan.md` §B.2/B.3). Classifier: feature/high, risk `none`, confidence 0.90.

## Scope
- Address the reported feature in the `other` area of the product.

## Files
- `tests/unit/`

## Acceptance
- All required CI gates pass.
- A human has answered the Decision Comment before this is merged.

gitchrisqueen:
## What & why

Feedback submitted in-app was auto-filed into the GitHub work cycle for **every** reporter. #793 asks for an admin gate: only admin users' feedback should be auto-triaged; everyone else's waits for a human decision in an in-app panel.

Owner answered the Decision Comment on the issue with `1A` — build it now.

**The gate**
- `users.is_admin` designates admins; `ADMIN_USER_EMAILS` (comma-separated, empty by default) is an additional allowlist.
- `process_new_feedback` / `auto_recluster_feedback` only act on admin reports. Everything else stays `new` until reviewed.
- Reclustering is gated too — attaching a report to an open cluster *accepts* it, so it needs the same check as filing.

**The panel**
- `GET /api/admin/feedback` (status/source filters, paging) and `POST /api/admin/feedback/{id}/review` with `approve` | `dismiss`. Approve runs the normal classifier/filer path; dismiss sets status `dismissed`. Both stamp `feedback.reviewed_by` / `reviewed_at`.
- Auth is a user-level admin role (`_require_user_admin` → 401/403), distinct from the operational `X-Admin-Secret` endpoints.
- SPA: `/admin/feedback` behind `AdminRoute`, nav entry shown only to admins, `is_admin` added to the session response.

### Two things worth a close look

1. **The gate filters in SQL, not in the loop.** Parked non-admin rows keep their `new` + NULL-cluster shape forever. A caller-side skip meant that once `limit` of them sat at the head of the FIFO, admin feedback would never be reached again — a permanent, silent stall of the auto-work loop. `get_unprocessed_feedback(admin_only=True)` does it in the query instead. `count_pending_admin_review()` reports the parked backlog on every pass so it can't grow unseen.
2. **Bootstrap.** `is_admin` defaults to 0, so without a seed *no* user is an admin: the filer would park every report and nobody could reach the panel to release them. The migration seeds the founding account (`UPDATE users SET is_admin = 1 ORDER BY id ASC LIMIT 1`, a no-op on an empty table) and `ADMIN_USER_EMAILS` is the no-SQL path for any other deploy. Both fail closed — empty grants nobody, and a DB error is never read as admin rights.

## Migration

`V20260729194956__add_feedback_admin_review.sql` — additive: `users.is_admin TINYINT(1) NOT NULL DEFAULT 0`, `feedback.reviewed_by` / `reviewed_at` + index, plus the one-row bootstrap `UPDATE` above. Nothing dropped or rewritten.

## Testing

- `poetry run pytest tests/unit -q` → **8574 passed**.
- New/updated: `tests/unit/api/test_feedback_admin_api.py` (403/401/422/404 paths, approve runs the filer + records the reviewer, dismiss, `is_admin` on the session), `tests/unit/utilities/test_db_feedback.py` (`is_user_admin` incl. allowlist + fail-closed, `admin_only` SQL shape, `count_pending_admin_review`), `tests/unit/utilities/test_feedback_issue_service.py` (both passes ask for `admin_only=True`; backlog is reported).
- `npx tsc --noEmit` clean. Vitest (incl. the new `AdminFeedbackPage.test.tsx`) runs in CI on Node 22 — local Node 18 can't start it.

Closes #793

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,6 +7,9 @@ CQC_LEM_CHECK_SCHEDULE_DELTA_MINUTES=5
 CQC_LEM_POST_TIME_DELTA_MINUTES=20
 # Secret key required in X-Admin-Secret header for admin endpoints
 ADMIN_SECRET=change_me_to_a_strong_random_secret
+# Comma-separated emails that may use the in-app feedback triage panel (issue #793), on top of the
+# users.is_admin column. Only admin feedback is auto-filed into GitHub; everything else waits here.
+ADMIN_USER_EMAILS=
 # Comma-separated bearer tokens accepted on /api routes (Authorization: Bearer ...).
 # Leave EMPTY for local/dev to disable the gate. Set on internet-facing deploys.
 API_ACCESS_TOKENS=
--- a/compose/local/database/migrations/V20260729194956__add_feedback_admin_review.sql
+++ b/compose/local/database/migrations/V20260729194956__add_feedback_admin_review.sql
@@ -1,0 +1,12 @@
+-- Admin feedback triage panel (issue #793): designate admin users and track who reviewed a feedback row.
+--
+--   users.is_admin        -> TINYINT(1): 1 = this user may use the feedback triage panel.
+--   feedback.reviewed_by  -> user id of the admin who approved/dismissed the row.
+--   feedback.reviewed_at  -> when the admin action happened.
+ALTER TABLE users
+    ADD COLUMN is_admin TINYINT(1) NOT NULL DEFAULT 0;
+
+ALTER TABLE feedback
+    ADD COLUMN reviewed_by INT NULL,
+    ADD COLUMN reviewed_at DATETIME NULL,
+    ADD INDEX idx_feedback_reviewed (reviewed_by, reviewed_at);
--- a/compose/local/database/migrations/V20260729194956__add_feedback_admin_review.sql
+++ b/compose/local/database/migrations/V20260729194956__add_feedback_admin_review.sql
@@ -10,3 +10,10 @@ ALTER TABLE feedback
     ADD COLUMN reviewed_by INT NULL,
     ADD COLUMN reviewed_at DATETIME NULL,
     ADD INDEX idx_feedback_reviewed (reviewed_by, reviewed_at);
+
+-- Bootstrap: with the column defaulting to 0, NO user is an admin, so the auto-filer would park
+-- every report and nobody could reach the panel to release them — the feedback->auto-work loop
+-- would go silently dead on deploy. Seed the founding account (lowest id) so the loop keeps
+-- running. Additional admins: UPDATE users SET is_admin = 1 WHERE email = '...', or the
+-- ADMIN_USER_EMAILS env allowlist. No-op on a fresh install with no users yet.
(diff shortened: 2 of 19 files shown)


gitchrisqueen:

- **Re-review of a settled row double-counted its own GitHub issue** (`api/main.py`): `POST /admin/feedback/{id}/review` ran the filer on any row it found. A filed row *is* its own open cluster (`get_open_feedback_clusters`: `s.cluster_id = s.id AND s.status IN ('clustered','issue_created')`), so a second approve re-classified it (LLM spend) and matched it to itself at similarity 1.0 — posting a false "+1 — another report" on the very issue it created and inflating `reporter_count`/`item_count`. Dismiss had the mirror bug: the row went `dismissed` while its issue stayed open. The panel's `status === 'new'` button check is not the guard — the list is cached 30s and the filing beat or a second admin can settle a row between render and click. Now 409 for `clustered`/`issue_created`/`resolved` or any row carrying an issue number; three tests cover it.
- **Failed approve/dismiss was invisible** (`AdminFeedbackPage.tsx`): `void handleAction(...)` awaited `mutateAsync` with nothing to catch the rejection — a GitHub outage, the new 409, or an expired session produced an unhandled rejection and no UI change, so the click read as a no-op. Switched to `mutate` + a rendered error, and the filer's actual outcome is now shown (approve can rate-limit/drop/error and leave the row exactly where it was — "approved" never meant an issue exists). Vitest added.
- **`get_feedback_list` dragged `f.embedding` out of MySQL** for a column the panel never renders: a 50-row page pulled 50 full vectors to discard them. Dropped from the SELECT.
- **`is_admin_reporter` ignored `ADMIN_USER_EMAILS`**: the auto-filer's join counts an allowlisted email as admin, so those reports *are* filed automatically — but the panel labelled them non-admin, i.e. as if still awaiting review. `get_feedback_list` now applies the same allowlist.
- **CodeQL PR gate failure**: removed the unused `_SECRET` global in `tests/unit/api/test_feedback_admin_api.py`.

Also verified and found clean: the SQL-side admin gate (starvation reasoning holds — the LEFT JOIN is param-bound, allowlist lowercased both sides), fail-closed `is_user_admin`, the migration's additive shape + one-row bootstrap, `AdminRoute`/`isAdmin` reset on logout, and the `Closes #793` claim (every acceptance criterion is covered, the Decision Comment was answered `1A`, no later phase left in prose).

Ran `tests/unit/api` + the feedback DB/service suites: 951 passed. `tsc --noEmit` and eslint clean; vitest needs CI's Node 22.

gitchrisqueen:

PR #837, opened 2026-07-31
fix(replies): confirm email forwarding from a forwarded email, not just our click (closes #813)

### LINKED ISSUE #813: fix(replies): Email forwarding not confirmed despite setup for comment replies
## Why
User has configured "Replying to comments on my posts" but still sees the warning "Email forwarding is not confirmed yet, so replies to comments on your posts will never fire. Finish the 3-step setup below." The issue prevents comment reply handling, including @mentions in emails.

Auto-filed from in-app feedback #22 by the feedback→auto-work loop (`docs/launch-and-marketing-plan.md` §B.2/B.3). Classifier: bug/high, risk `none`, confidence 0.92.

## Scope
- Address the reported bug in the `replies` area of the product.

## Files
- `tests/unit/`

## Acceptance
- All required CI gates pass.

gitchrisqueen:
## What

"Email forwarding is not confirmed yet, so replies to comments on your posts will never fire" stayed on screen for a user who **had** finished the setup.

`gmail_forward_confirmation.confirmed` was only ever set by our **server-side click** of Gmail's verify link. `_handle_gmail_forwarding_confirmation` already assumes that click often fails ("interstitial / datacenter IP") and emails the link + code to the user so they can finish it from their own signed-in browser — but nothing recorded that they did. The record stayed `confirmed: false` forever, so conflict **C6** kept telling a working setup it would never fire. The 7-day Redis TTL compounded it: even a *successful* auto-click silently un-confirmed itself a week later, and the Gmail filter is set up once and forwards indefinitely.

## Why this fix

Confirm on **evidence**, not on our own click. A LinkedIn notification arriving at the user's private `reply+<token>@parse.<domain>` address is end-to-end proof the forward rule is live — strictly better proof than clicking a link.

- `_process_reply_inbound` records the confirmation when LinkedIn mail lands, **before** the comment/reaction split: a forwarded *reaction* email proves the chain just as well, and the status chip is about the chain working, not about that email being actionable.
- New `is_linkedin_notification()` (`notification_email.py`) is the evidence test — broader than `is_comment_notification` but not blanket, so unrelated mail to the address can't stand in as proof.
- `_store_gmail_forward_confirmation()` centralises the write: a confirmed record has **no TTL** and is never downgraded by a later pending one (re-adding the address and failing the auto-click again must not undo forwarding we've watched work). Pending records keep the 7-day TTL.

## Also: the warning stopped lying about where the user is

A pending record means Gmail already asked us to verify — the address *was* added. Repeating "finish the 3-step setup" is what made a half-done setup read as a total failure, so **C6** and the `ReplySetupSteps` chip now have three states and name the one outstanding step ("Waiting for your first forwarded email") instead of sending the user back to step 1.

## Testing

- `tests/unit/api/test_main_general.py` — `TestForwardingConfirmedByDelivery`: forwarded comment email confirms; the confirmed record carries no `ex`; a forwarded reaction confirms but starts no sweep; unrelated mail is not evidence; an already-confirmed record isn't rewritten; a failed re-confirmation never downgrades a confirmed one; pending still gets its TTL.
- `tests/unit/utilities/linkedin/test_notification_email.py` — `TestIsLinkedInNotification`.
- `conflicts.test.ts` — C6's three messages; `ReplySetupSteps.test.tsx` (new) — chip states + the code fallback only while outstanding.
- `poetry run pytest tests/unit -q` → **8591 passed**. `tsc --noEmit` clean.

No migration, no schema change, no new env var.

Closes #813

### DIFF
--- a/src/cqc_lem/api/main.py
+++ b/src/cqc_lem/api/main.py
@@ -1377,31 +1377,35 @@ async def linkedin_verification_pin_inbound(request: Request) -> ResponseModel:
 
 
 _GMAIL_CONFIRM_KEY = "linkedin:gmail_forward_confirm:{user_id}"
-# A pending record is transient (the code fallback goes stale); a CONFIRMED one is not. The Gmail
-# filter is set up ONCE and forwards indefinitely, so expiring a proven-working record re-raised the
-# "forwarding is not confirmed" warning a week later on a setup that was fine (issue #813).
-_GMAIL_PENDING_TTL_SECONDS = 7 * 24 * 60 * 60
+# What goes stale is the CODE fallback, not the fact that Gmail asked us to verify the address. The
+# record itself is what the UI reads to tell "never started" apart from "waiting on the first
+# forwarded email", and neither the Gmail filter nor the user's progress through it expires — an
+# expiring record put a fine setup back on "Forwarding not confirmed yet" (issue #813). So no record
+# carries a TTL; only the code is dropped once it is too old to be worth offering.
+_GMAIL_CODE_FRESH_SECONDS = 7 * 24 * 60 * 60
 
 
-def _store_gmail_forward_confirmation(user_id: int, record: "dict") -> bool:
-    """Persist forwarding status; True when something was written. Confirmed records never expire and
-    are never downgraded by a later pending one — evidence that the chain worked does not stop being
-    evidence."""
+def _store_gmail_forward_confirmation(user_id: int, record: "dict", quiet: bool = False) -> bool:
+    """Persist forwarding status; True when something was written. Confirmed records are never
+    downgraded by a later pending one — evidence that the chain worked does not stop being evidence.
+    Pass quiet=True on per-email paths so one Redis outage can't turn into a warning flood."""
     try:
         from cqc_lem.utilities.linkedin.rate_limit import _redis_client
         client = _redis_client()
         if client is None:
             return False
-        if record.get("confirmed"):
-            client.set(_GMAIL_CONFIRM_KEY.format(user_id=user_id), json.dumps(record))
-            return True
-        if (get_gmail_forward_confirmation(user_id) or {}).get("confirmed"):
-            return False
-        client.set(_GMAIL_CONFIRM_KEY.format(user_id=user_id), json.dumps(record),
-                   ex=_GMAIL_PENDING_TTL_SECONDS)
+        if not record.get("confirmed"):
+            if (get_gmail_forward_confirmation(user_id) or {}).get("confirmed"):
+                return False
+            if record.get("code"):
+                record = {**record, "code_expires_at": int(time.time()) + _GMAIL_CODE_FRESH_SECONDS}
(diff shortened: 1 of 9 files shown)


gitchrisqueen:

Reviewed the diff, the surrounding inbound-mail path, and issue #813's acceptance criteria. The core call — confirm on *evidence* (a LinkedIn notification actually landing on the private `reply+<token>` address) rather than on our own datacenter-IP click — is right, and the `is_linkedin_notification` gate is appropriately narrow (the parse host is `parse.christopherqueenconsulting.com`, so no self-referential "linkedin" match from an echoed To: line). Two real defects at the seams, both fixed in `6a3dc1b`:

- **The pending record's 7-day TTL decayed the fix back into the bug.** The new three-state UI reads the *record* to tell "never started" from "waiting on the first forwarded email", but pending records still expired. The recommended Gmail filter is `From linkedin.com` + `Subject: commented OR replied`, so a user with a quiet week forwards nothing — and after 7 days a finished-by-hand setup dropped from "Waiting for your first forwarded email" back to "Forwarding not confirmed yet. Finish the 3-step setup below." That is exactly #813's symptom, one week later. What actually goes stale is the *code*, not the fact that Gmail asked us to verify: the record is now stored with no TTL and carries `code_expires_at`, and `get_gmail_forward_confirmation` drops only the `code` once it passes. UI and C6 copy unchanged (the code fallback is already gated on `confirmation.code`). Replaced `test_pending_confirmation_still_stored_with_a_ttl` with `test_pending_confirmation_outlives_the_code_it_carries`, plus `test_stale_code_is_dropped_but_the_record_survives` / `test_fresh_code_is_still_offered`.
- **A Redis outage now files a defect for working behaviour.** `_store_gmail_forward_confirmation` moved onto the per-inbound-email path, but kept `log_warning` in its except branch — and `_redis_client()` builds the handle lazily, so an unreachable Redis raises on `.set()`, once per forwarded email. Per `src/cqc_lem/utilities/CLAUDE.md`, a repeated `log_warning` re-emits at ERROR and files a grouped `$exception` → auto GitHub issue. The delivery path now passes `quiet=True` and logs DEBUG; the once-per-Gmail-confirmation auto-click path still warns. Covered by `test_redis_outage_on_the_delivery_path_does_not_warn`. (The existing guard comment in `_record_forwarding_confirmed_by_delivery` claimed this was already handled — it only covered the INFO log.)

Checked and NOT flagged: `Closes #813` is honest (both acceptance criteria met, no later phase in the issue); confirmed records are never downgraded by a later pending one, and `SET` without `EX` correctly clears a legacy TTL on upgrade; a forwarded reaction email confirms without triggering a sweep; the webhook stays token-gated and always 200s. One thing left deliberately alone: `is_comment_notification` still classifies "mentioned you" as a reaction, so an @mention-in-a-comment email confirms forwarding but starts no sweep — that predates this PR and changing it is a separate scope call.

`poetry run pytest tests/unit -q` → **8594 passed**. UI tests (`vitest`) not runnable in this worktree — deferred to CI.

gitchrisqueen:

Issue #816, opened 2026-07-31
fix(engagement): reaction flow broken by SDUI drift — 63 selector misses in 48h


gitchrisqueen:
## Why

The inline post-reaction flow is missing its anchors on nearly every attempt. PostHog Logs, last 48h:

| Message | Occurrences | Days |
|---|---|---|
| `Selector miss: Open reactions menu` | 30 | 3 |
| `Selector miss: Reaction state` | 17 | 2 |
| `Selector miss: Reaction state (post-click)` | 16 | 2 |

`react_to_post_inline` (`src/cqc_lem/app/run_automation.py:901`) keys on three aria-labels that the docstring says were "verified live": `button[aria-label^='Reaction button state']` (`:912`), `button[aria-label='Open reactions menu']` (`:932`), and the post-click state re-read (`:949`). Each is a **single locator with no fallback chain**, unlike `find_first`'s intended ordered-fallback design. LinkedIn's SDUI has drifted and all three now miss.

Consequence: reactions are not landing, and each miss costs `MAX_WAIT_RETRY` × ~5s of wall clock inside a Selenium session that holds a browser slot.

## Scope

Re-ground the three locators against the live DOM and give each an ordered fallback chain, per the `find_first` contract and the CLAUDE.md SDUI guidance (prefer `data-testid` / `aria-label`, never hashed class names).

Use `scripts/linkedin_live_validation.py` and/or the `selenium-lem` MCP browser tools to capture the current markup before changing selectors — do not guess.

Note the related mis-flagging bug being fixed separately: `react_to_post_inline` returns `False` both for "already reacted" (a benign no-op, `:914-915`) and for genuine failure, and the caller at `:1184` logs both as `Could not leave a reaction on post`.

## Files

- `src/cqc_lem/app/run_automation.py` (`react_to_post_inline`, and the second reaction path at `:3023-3035`)

## Acceptance criteria

- [ ] All three anchors re-grounded against live DOM with an ordered fallback chain each
- [ ] Live validation output pasted into the PR showing a reaction actually registering
- [ ] `Selector miss: Open reactions menu` / `Reaction state` volume drops to ~0 after deploy

gitchrisqueen:
## 🧑‍⚖️ Human decision needed — reply with option letters
[...]

PR #838, opened 2026-07-31
fix(engagement): center the comment composer before clicking it (closes #815)

### LINKED ISSUE #815: fix(engagement): feed commenting is failing — 'Inline comment post failed' 26x in 48h
## Why

`comment_on_feed_inline` is throwing on 26 posts in the last 48 hours (3 consecutive days). Surfaced by a PostHog Logs audit, not by an alert — the exception is swallowed into a WARNING at `src/cqc_lem/app/run_automation.py:897`, and `log_warning` can never produce a `$exception`, so it never reached error tracking and never filed an issue.

This is a core product function silently failing. Feed commenting is one of the two pillars in CLAUDE.md.

Evidence (PostHog `logs`, last 48h):
- `Inline comment post failed` — **26 occurrences across 3 days**, severity `warn`, with `exc=` attached.

## Scope

Find the real exception. `post_comment_inline` (`run_automation.py:~870-898`) wraps the whole compose+submit sequence in one `try`, so the WARNING hides which step throws: "Open comment composer" click, the `div[role='textbox']` lookup, `send_keys`, the `_SUBMIT_NEAR_COMPOSER_JS` execute_script, or `_composer_submitted` verification.

Likely candidates given LinkedIn's SDUI:
- `composer.send_keys(comment_text)` throwing on non-BMP emoji — CLAUDE.md documents `_strip_non_bmp()` for exactly this, check it is applied on this path.
- A stale-element/intercepted-click as the composer re-renders between `click()` and `send_keys`.
- `_SUBMIT_NEAR_COMPOSER_JS` returning falsy and the `Keys.CONTROL+RETURN` fallback also failing.

Do NOT just broaden the `except`. Narrow the try blocks so the failing step is identifiable, and log the step in the message so the new escalation path files a distinct issue per failure mode.

## Files

- `src/cqc_lem/app/run_automation.py` (`post_comment_inline`, `_composer_submitted`)
- `src/cqc_lem/utilities/selenium_util.py` (`find_first`, `click_first`)

## Acceptance criteria

- [x] The failing step is identified and named in the log message
- [x] Non-BMP emoji stripping verified on this path
- [x] Unit test covering the failure mode with a mocked driver
- [ ] `Inline comment post failed` volume drops in PostHog Logs after deploy _(post-deploy observation — verifiable only once the fix ships)_

gitchrisqueen:
## What & why

`comment_on_feed_inline` threw 26 times in 48h, swallowed into one `Inline comment post failed`
WARNING. The v0.113.0 recurrence-escalation captured the real exception (recorded on the issue):

```
ElementClickInterceptedException: element click intercepted:
  Element <div class="ql-editor" contenteditable="true" role="textbox" ...>
  is not clickable at point (890, 9).
  Other element would receive the click: <svg ...>
```

`y=9` is the tell — the composer was sitting at the very top of the viewport, under LinkedIn's
**sticky global nav**, so the nav's `<svg>` received the click. The composer was found fine; it was
never a selector or an emoji problem.

## Changes

- **`_focus_composer()`** — scroll the composer to the **middle** of the viewport (`scrollIntoView({block:'center'})`),
  then click. On `ElementClickInterceptedException` it re-centers and retries **once**; a *second*
  interception is allowed to raise. A JS `arguments[0].click()` would also bypass the nav, but would
  equally bypass a genuine modal/overlay — this keeps real overlay faults visible.
- Applied to the two sibling inline composers that carried the identical top-of-viewport click:
  `_reply_to_comment_inline` and `_reply_under_comment_inline`. (The reaction flow is #816 — not touched here.)
- **Step naming** — `post_comment_inline` now reports which step threw:
  `Inline comment post failed at focus composer` / `at type comment` / `at submit composer` /
  `at verify submit`, instead of one catch-all message. Step names are deliberately **quote- and
  digit-free**: `log_escalation.normalize_message` masks quoted strings and numbers, so a quoted step
  name would re-merge exactly the keys this split exists to separate — a test asserts all seven
  normalize to distinct dedup keys.
- **Non-BMP stripping verified** on this path (`_strip_non_bmp` was already applied) and now pinned by
  an assertion that only BMP text reaches `send_keys`.
- `CLAUDE.md` SDUI gotcha updated with the sticky-nav rule.

`selenium_util.find_first` / `click_first` needed no change — the lookup was never the failure.

## Testing

`tests/unit/app/test_comment_inline.py` — mocked-driver coverage for: centering before the click,
re-center-and-retry on the live `at point (890, 9)` interception, a second interception raising, an
unscrollable element (positioning is best-effort), emoji stripped before `send_keys`, and one test
per named step (focus / type / submit).

- `pytest tests/unit -q` → **8604 passed** locally.

## Note on the last acceptance box

`Inline comment post failed` volume dropping in PostHog Logs is a **post-deploy observation**, not
work left in the diff — the other three criteria are implemented and ticked on the issue. No later
phase is deferred.

Closes #815


gitchrisqueen:

- **Test regression the centering change introduced (`tests/unit/app/test_reply_inline.py`)**: `_focus_composer` adds one `execute_script` (the `scrollIntoView({block:'center'})`) to `_reply_to_comment_inline`, which shifted `test_returns_false_when_composer_still_full`'s two-entry `side_effect` list. The test still passed — but by accident: the third call exhausted the mock and `_composer_submitted`'s `except Exception: return False` swallowed the `StopIteration`, so the "no submit button" and "text not in the comment list" legs it was written to guard stopped being exercised (verified: 3 `execute_script` calls against 2 side effects). This is the same shift the PR correctly fixed in `test_comment_followups.py`, just missed in the sibling file. Fixed by restoring the sequence (`[None, False, False]`) and pinning `call_count == 3` so a future shift fails loudly rather than passing green. That guard matters here: `_composer_submitted` exists specifically to stop the old false-positive "comment posted" verdict.

Verified and found clean:
- The fix matches the root cause on the issue — `scrollIntoView({block:'center'})` then a real `.click()`, not a JS click, so a genuine modal/overlay still surfaces. One retry, second interception raises. `_scroll_into_center` is correctly best-effort (a positioning failure is left to the click to report).
- All three inline composer clicks are converted — `grep` shows no remaining bare `composer.click()` on a feed/reply composer. The group-post editor (`run_automation.py:2414`) is inside LinkedIn's share modal, not under the sticky nav, so leaving it is right.
- `_SUBMIT_NEAR_COMPOSER_JS` and `_composer_submitted` walk the DOM, not viewport coordinates, so re-centering can't perturb them; `_reply_under_comment_inline`'s nearest-composer proximity JS runs *before* the scroll, so the #478 "reply under the right comment" behaviour is unchanged.
- Step naming is sound and every name is reachable (`prepare text` fires on a `None` `comment_text`); `click_first` does not catch `ElementClickInterceptedException`, so an intercepted Comment button now correctly reports `at open composer`. The 7-distinct-dedup-keys test is the right guard given `normalize_message` masks quotes and digits.
- Honest close: acceptance criteria 1–3 are in the diff; criterion 4 is a post-deploy PostHog observation, not deferred work, and no later phase is declared. The reaction flow the issue asks about is tracked separately on #816 (open, `agent:ready`) — correctly out of scope here.
- `pytest tests/unit -q` → **8604 passed** locally after the fix.

gitchrisqueen:

- **CodeQL `py/empty-except` on `_scroll_into_center` (unresolved review thread, required gate).** The `except Exception: pass` that swallows a failed `scrollIntoView` carried its rationale only in the docstring, so Code Scanning opened alert 2629 on `run_automation.py:877` — a live thread on a required check (`CodeQL Security Analysis`) that the merge gate would have carried forward. Fixed in 49639df by naming the swallow inline (`pass  # a stale element or a rejected scroll is not a failure — the click below decides`), matching the file's existing best-effort pattern (`_reply_under_comment_inline`'s hover swallow). Thread resolved.

Everything else I went after held up:

- **The fix matches the captured exception.** `at point (890, 9)` is a hit-test rejection, so no click is dispatched to the nav — `scrollIntoView({block:'center'})` + retry is the right repair, and refusing the JS-click shortcut keeps a genuine modal/overlay visible instead of silently clicking through it. The rest of the chain is already interception-immune (`send_keys` targets the element directly; `_SUBMIT_NEAR_COMPOSER_JS` clicks in JS), so centering the composer closes the whole observed path.
- **Step naming is genuinely distinct.** Re-ran the `normalize_message` masks against all seven step strings — quote- and digit-free holds, 7 distinct dedup keys, no collision with the `<str>`/`<n>`/`<list>` maskers.
- **Both sibling call sites were re-verified**, and the three mocked `execute_script` side-effect lists that shifted (`test_reply_inline`, `test_comment_followups`) line up with the real call order — `test_returns_false_when_composer_still_full` pins the count so a future shift can't turn a swallowed `StopIteration` into a passing test.
- **Emoji stripping** is applied before `send_keys` on all three paths and is now asserted, not assumed.
- **Close is honest.** Acceptance criteria 1–3 are implemented and ticked; the fourth ("volume drops in PostHog Logs") is a post-deploy observation, not deferred scope — and a recurrence now files its own per-step issue through the v0.113.0 escalation path rather than staying invisible, so nothing needs a follow-up issue to stay tracked. No later phase is declared anywhere in #815.

Not changed, for the record: the `button[aria-label='Comment']` click that *opens* the composer goes through `click_first`, which does not treat `ElementClickInterceptedException` as a miss under `required=False`. The captured trace shows that click succeeding (the composer opened and was found), so hardening it here would be speculative — and it now reports as `Inline comment post failed at open composer` if it ever does fire. The reaction flow is #816.

`pytest tests/unit -q` → 8604 passed locally on the pushed head.

gitchrisqueen:
🧩 phase-guard — heads up: issue #815 still has **boxes: 1 unchecked** acceptance box(es). Merging anyway (unticked boxes are common here), but if any of that scope did NOT ship, file a follow-up issue with `agent:ready` and link it, or drop the `Closes #815` from this PR.

gitchrisqueen:

PR #839, opened 2026-07-31
docs(analytics): SPA posthog-js does transmit — the grid result was a proxy artifact (closes #834)

### LINKED ISSUE #834: verify(analytics): SPA posthog-js initializes but transmitted no events in the Selenium test browser
## What's confirmed working

After `UI_POSTHOG_KEY` shipped in v0.113.x, posthog-js **initializes in production for the first time**. Verified live against `https://lem.christopherqueenconsulting.com`:

```
posthog: loaded
token:  phc_wP8C…            (correct project)
api_host: us.i.posthog.com   (ingest host, correct)
disable_session_recording: false
session_recording.sampleRate: 0.1
session_recording.maskAllInputs: true
distinct_id / session_id: assigned
```

All PostHog assets fetch, including `posthog-recorder.js`, `surveys.js`, `exception-autocapture.js` and the remote `config.js`. Project side is correct too: `session_recording_opt_in: true`, `minimum_duration_milliseconds: 5000`, project `sample_rate: null` (100%, so it doesn't multiply with the SDK's 0.1).

Session recording **can be activated**: `posthog.startSessionRecording({sampling: true})` flipped `sessionRecording.status` from `disabled` → `sampled`. (`disabled` is what posthog-js reports for a session that simply lost the 10% sampling draw — expected, not a fault.) That is the same call `ensureSessionRecorded()` makes on `$exception` / feedback-widget open, so the forced-recording path works.

## What did NOT work

**The SDK transmitted nothing.** Over ~60s with repeated `posthog.capture(...)` calls: zero requests to any capture endpoint, and zero events in PostHog. `$pageview` is still absent from the project taxonomy.

Not a config problem — all of these were checked and are clean: `has_opted_out_capturing() = false`, `opt_out_capturing_by_default = false`, `before_send = undefined`, `disable_persistence = false`, `__loaded = true`, `document.visibilityState = visible`.

And **the endpoint works from that same browser**: a manual `fetch` POST of a plain-JSON event to `https://us.i.posthog.com/i/v0/e/` returned `200 {"status":"Ok"}` and the event **landed in PostHog** (`replay_wiring_manual`).

## The likely explanation — and why this needs a real browser to settle

The test ran in the **Selenium grid Chrome, whose egress is a proxy** (measured egress `2.25.68.66`, not the VPS). posthog-js sends with `compression: "gzip-js"` — a gzipped POST body. My successful manual probe sent **uncompressed** JSON.

So the most probable cause is the proxy mishandling or dropping the gzipped body, which would mean **a normal user's browser is fine** and this is a test-environment artifact. That is a hypothesis, not a conclusion — I could not confirm it, and I did not want to report "SPA analytics is broken" on the strength of a proxied headless browser.

## Scope

1. **Load the app in a normal browser** (no proxy) and check for `$pageview` / `$autocapture` in PostHog within a minute. That single check decides everything below.
2. If events DO flow → close this; optionally note the proxy caveat in `docs/posthog-advanced-surface.md` so the next person doesn't repeat the investigation.
3. If events do NOT flow → the send path is genuinely broken. Start by testing `compression: undefined` in `initAnalytics()` (`ui/src/utils/analytics.ts`), since that is the one difference between the failing SDK path and the succeeding manual POST.
4. Either way: nothing will be recorded for most sessions at `sampleRate: 0.1` — to confirm replay end-to-end, trigger an exception or open the feedback widget, which forces recording regardless of sampling.

## Files

- `src/cqc_lem/ui/src/utils/analytics.ts` (`initAnalytics`, `ensureSessionRecorded`)

## Acceptance criteria

- [x] `$pageview` appears in the project taxonomy after a real browser visit — verified 2026-07-31, desktop Chrome + Mobile Safari, unproxied
- [x] At least one session recording exists in PostHog — two, both forced by `ensureSessionRecorded()` on feedback-widget open
- [x] If a code change was needed, a note in `docs/session-replay.md` explaining why — n/a: no code change was needed. The proxy caveat is documented in `docs/posthog-advanced-surface.md` + `docs/session-replay.md` instead (scope item 2)

gitchrisqueen:
## What & why

#834 asked one deciding question: does the SPA's posthog-js actually transmit from a **real, unproxied** browser, or is the send path genuinely broken? The original investigation ran in LEM's Selenium grid Chrome, where the SDK initialized perfectly but sent nothing over ~60s of `capture()` calls — while a hand-rolled *uncompressed* `fetch` to the same `/i/v0/e/` endpoint, from that same browser, returned `200` and landed.

**It transmits.** Checked live against production (`lem.christopherqueenconsulting.com`, posthog-js 1.407.3, project 475262) on 2026-07-31:

| Event | Count (21d) | First → last |
|---|---|---|
| `$autocapture` | 58 | 19:34 → 21:48 UTC |
| `$pageview` | 19 | 19:34 → 21:48 UTC |
| `$web_vitals` | 7 | — |
| `feedback_opened` | 2 | — |
| `$pageleave` / `$identify` | 1 / 1 | — |

From **two independent unproxied clients** — desktop Chrome on a residential IPv6 address and Mobile Safari on a different IPv4 — not one browser that might be special. `$identify` resolves `distinct_id` to `1`, so browser and backend land on one PostHog person as designed.

**Two session recordings exist**, and they're the interesting part: each starts within ~150 ms of the `feedback_opened` event that triggered it (20:22:36.531 vs .671; 21:48:14.944 vs 21:48:15.001). That is `ensureSessionRecorded()` forcing a session the 10% sample would otherwise have dropped — the one replay rule config alone can't demonstrate.

So the grid result was a **test-environment artifact of the per-user residential proxy** (`utilities/proxy.py`) that Chrome egresses through, mishandling posthog-js's gzipped POST body. Not a defect.

## What changed

**No code change.** `compression` stays at the posthog-js default; `ui/src/utils/analytics.ts` is untouched. Docs only, so the next person doesn't repeat the investigation — and doesn't re-run the check in the browser that produces the false negative:

- `docs/posthog-advanced-surface.md` — new "Verifying the SPA surface: never in the Selenium grid Chrome" section: the false negative, what was ruled out, the real-browser result, and why a quiet project isn't proof either way (`capture_pageview: false` + build-time `VITE_POSTHOG_KEY`).
- `docs/session-replay.md` — same caveat on the "Verifying it" steps, plus the end-to-end confirmation of the forced-recording path.

## Testing notes

Verification is the deliverable here and it was done against live PostHog, not a test double — event counts, per-event browser/IP/`$session_id` metadata, and the recordings list are all quoted above. No new logic, so no unit tests; docs-only diff.

## Scope check

All three acceptance criteria are met and ticked on the issue. Criterion 3 is conditional ("*if* a code change was needed") and no code change was needed — the proxy caveat is documented instead, which is scope item 2's explicit instruction. No later phase is described in the issue, so nothing is left untracked.

Closes #834

### DIFF
--- a/docs/posthog-advanced-surface.md
+++ b/docs/posthog-advanced-surface.md
@@ -204,3 +204,37 @@ through OUR gate instead of around it.
 Set `POSTHOG_OPS_WEBHOOK_URL` in `/opt/lem/.env` to a Slack incoming-webhook URL (or any other
 `https://` endpoint) and run `scripts/posthog_ops_destination.py --apply` to make the realtime
 429-breaker ping live.
+
+---
+
+## Verifying the SPA surface: never in the Selenium grid Chrome (issue #834)
+
+**Do not test browser-side analytics in LEM's own Selenium Chrome.** Its egress is the per-user
+static residential proxy (`utilities/proxy.py`), and a proxied session drops the SDK's sends while
+leaving every piece of local config looking healthy — which reads, wrongly, as "SPA analytics is
+broken".
+
+That is exactly what happened once posthog-js first shipped to production with `UI_POSTHOG_KEY` in
+v0.113.x. In the grid browser the SDK initialized cleanly (correct token and `api_host`, `__loaded`,
+not opted out, no `before_send`, replay assets all fetched, `startSessionRecording()` flipping
+`disabled` → `sampled`) yet transmitted **nothing** over ~60s of `posthog.capture()` calls — while a
+hand-rolled `fetch` POST of plain JSON to the same `/i/v0/e/` endpoint, from that same browser,
+returned `200` and landed. The one difference between the two paths is the body: posthog-js sends
+`compression: "gzip-js"`, the manual probe did not.
+
+**Verified in real browsers on 2026-07-31: the SDK transmits fine.** Same production build
+(posthog-js 1.407.3, `https://lem.christopherqueenconsulting.com`), two independent unproxied
+clients — desktop Chrome and Mobile Safari — produced `$pageview`, `$autocapture`, `$web_vitals`,
+`$pageleave`, `$identify` and the named `feedback_opened` product event, all within seconds. So the
+grid result was a test-environment artifact of the proxy, **not** a defect: no code change was made,
+and `compression` stays at the posthog-js default.
+
+Two things follow for anyone re-checking this:
+
+- Verify from an ordinary browser (yours, or any non-proxied one). If you must use automation, drive
+  a Chrome whose egress is NOT the proxy.
+- A quiet project is not proof either way — with `capture_pageview: false` the router owns pageviews
+  (`ui/src/utils/analytics.ts`), and with no `VITE_POSTHOG_KEY` in the build the posthog-js chunk is
+  never even fetched. Confirm the build first, then look for events.
+
+The replay half of the same check is in `docs/session-replay.md` § Verifying it.
--- a/docs/posthog-advanced-surface.md
+++ b/docs/posthog-advanced-surface.md
@@ -209,10 +209,9 @@ Set `POSTHOG_OPS_WEBHOOK_URL` in `/opt/lem/.env` to a Slack incoming-webhook URL
 
 ## Verifying the SPA surface: never in the Selenium grid Chrome (issue #834)
 
-**Do not test browser-side analytics in LEM's own Selenium Chrome.** Its egress is the per-user
-static residential proxy (`utilities/proxy.py`), and a proxied session drops the SDK's sends while
-leaving every piece of local config looking healthy — which reads, wrongly, as "SPA analytics is
-broken".
+**Do not test browser-side analytics in LEM's own Selenium Chrome.** That browser drops the SDK's
+sends while leaving every piece of local config looking healthy — which reads, wrongly, as "SPA
+analytics is broken".
 
 That is exactly what happened once posthog-js first shipped to production with `UI_POSTHOG_KEY` in
 v0.113.x. In the grid browser the SDK initialized cleanly (correct token and `api_host`, `__loaded`,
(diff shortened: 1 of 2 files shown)


gitchrisqueen:

The deliverable here is a *claim* that a live verification happened, so I re-ran the verification independently against PostHog project 475262 rather than taking the PR body at face value. **Every number holds:** `$autocapture` 58 / `$pageview` 19 / `$web_vitals` 7 / `feedback_opened` 2 / `$pageleave` 1 / `$identify` 1, 19:34:05 → 21:48:35 UTC; three web sessions across two clients (Chrome/macOS on a residential IPv6, Mobile Safari/iOS on a different IPv4), `$lib_version` 1.407.3 on all three, `distinct_id = 1` on the identified session. Both recordings exist and start at 20:22:36.531 and 21:48:14.944 — 140 ms and 57 ms *before* their `feedback_opened` events, and 6 s / 13 s *after* their sessions' first events, which is the actual proof that `ensureSessionRecorded()` forced them rather than the 10% sample catching them. Close is honest: all three acceptance criteria are met and criterion 3 is genuinely conditional. CI is green.

What I fixed — both are the write-up asserting more than the evidence supports, which matters more than usual in a PR whose whole product is a documented conclusion:

- **Proxy promoted from suspect to cause**: the issue was explicit that gzip-vs-proxy is "a hypothesis, not a conclusion", but the docs stated it as settled ("was a test-environment artifact of the proxy"). The grid session also differs by the MV3 proxy-auth extension and headless/automation flags, and nobody re-ran it with the proxy off. Reworded to "an artifact of that browser's environment, **not** a defect in the SPA", plus a short paragraph naming the leading suspect, the unruled-out variables, and the one test that would settle it.
- **"Its egress is the per-user static residential proxy" is not unconditionally true**: `get_docker_driver()` applies whatever `resolve_proxy()` returns — explicit user proxy → regional → global default → **none** — and a session created without `user_id` egresses straight from the host (`selenium_util.py:119-160`). As written, someone whose grid session isn't proxied would conclude the caveat doesn't apply to them. The operational rule is now "not the grid, whatever its egress", and `docs/session-replay.md` no longer attributes the silence to the proxy either.

Docs-only, no logic touched, so no tests to add.

gitchrisqueen:

PR #840, opened 2026-07-31
fix(deploy): resilient worker converge + verify + early tag persist (closes #831)

### LINKED ISSUE #831: fix(deploy): compose recreate races and leaves the whole worker tier in Created — second occurrence
## Why

The v0.113.2 deploy dropped **every Celery worker** at the final `docker compose up -d --remove-orphans` step:

```
Container celery_flower  Recreated
Container celery_beat    Recreated
Container celery_worker  Error response from daemon: No such container: 5edffe0695...
Error response from daemon: No such container: 5edffe0695...
```

`deploy.sh` aborted there. The result: `web_api_blue`/`web_api_green` healthy on the new tag, and **all five celery workers + beat + flower sitting in `Created`** — the worker tier completely down, with no alert. `IMAGE_TAG` and `.last_good_tag` were never written because the script exits before `persist_image_tag`, so the box also disagreed with itself about what was deployed.

**This is the second occurrence.** 2026-07-28 had the same signature (all celery workers + Grid hub down, containers left in `Created`) and was attributed to a timeout killing compose mid-start. It is not a one-off, and the failure is silent — nothing pages when the worker tier stops.

Recovery both times was a manual `docker compose up -d --remove-orphans` re-run, which converged cleanly on the first try.

## Probable cause

Compose renames the old container (`<hash>_celery_worker`) before creating the replacement. If anything references the old ID between the rename and the create — a concurrent `docker exec` from the health/maintenance path, or compose's own bookkeeping — the daemon returns `No such container` and compose aborts **mid-tier**, leaving the already-renamed services stopped and the not-yet-processed ones untouched.

Note the deploy runs `maint`/health `docker exec`s against `celery_worker` around this window, which is a plausible source of the racing reference.

## Scope

- Make the worker recreate **resilient**: retry the converge once on `No such container` before failing, and/or recreate services explicitly rather than relying on one bulk `up -d`.
- Make it **loud**: `deploy.sh` should verify every expected service is `running` before declaring success, and fail (or roll back) if any is left in `Created`/`Exited`. Right now the script exits non-zero but nothing checks the resulting container states.
- Make it **atomic enough to resume**: `persist_image_tag` / `.last_good_tag` should reflect reality even on a partial deploy, or the next deploy starts from a wrong baseline.
- Consider whether `maint`'s `docker exec` into `celery_worker` should be skipped once the recreate has begun.

## Files

- `scripts/deploy.sh` (final converge step, `maint` helper, `persist_image_tag`)
- `docs/zero-downtime-deploys.md`

## Acceptance criteria

- [x] A recreate that hits `No such container` retries and converges instead of aborting
- [x] Deploy fails loudly if any expected service is not `running` at the end
- [x] A partial deploy leaves `IMAGE_TAG` / `.last_good_tag` consistent with what is actually running
- [x] Regression test asserting deploy.sh verifies final container state

## Note

The blue/green web cutover was NOT affected — that had already completed and the site stayed up on the new tag throughout. This is purely the worker tier.

gitchrisqueen:
## What\n\nHardens `scripts/deploy.sh` against the Docker compose recreate race that left the whole worker tier in `Created` on the v0.113.2 deploy (issue #831).\n\n## Changes\n\n- `converge_stack()`: retries the final `docker compose up -d --remove-orphans` once when the output contains "No such container".\n- `verify_stack_running()`: enumerates expected compose services and fails loudly if any container is stuck in `Created`, `Exited`, or `Dead`, or if an expected service has no running container.\n- Persists `IMAGE_TAG` / `.last_good_tag` immediately after the blue/green edge flip, so a later partial deploy still records the tag that is actually running.\n- Wraps the top-level deploy code in `main()` so the script can be sourced for unit tests.\n- Adds `tests/unit/test_deploy.py` regression tests that source `deploy.sh` and exercise the new functions with a fake `docker-compose`.\n- Updates `docs/zero-downtime-deploys.md` with the new resilience behavior.\n\n## Testing\n\n- `poetry run pytest tests/unit/test_deploy.py -q`: 6 passed.\n- `poetry run pytest tests/unit/test_deploy.py tests/unit/app/test_worker_shutdown.py -q`: 31 passed.\n\nCloses #831\n\n🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/zero-downtime-deploys.md
+++ b/docs/zero-downtime-deploys.md
@@ -45,14 +45,25 @@ That aborts the converge mid-tier and leaves workers in `Created` — a silent w
 
 `scripts/deploy.sh` now:
 
-- **Retries once** if the converge output contains `No such container`.
-- **Verifies the final state** by enumerating expected compose services and checking that
-  none are stuck in `Created` / `Exited` / `Dead`. If verification fails, the deploy exits
-  with a loud error instead of leaving the worker tier down silently.
+- **Retries once** if the converge output contains `No such container` (`converge_stack`).
+  Every attempt's compose output is echoed to the deploy log, success or failure — those
+  per-container lines are the only evidence of what the converge did, and they are what
+  identified this race both times it happened.
+- **Verifies the final state** (`verify_stack_running`): `compose config --services` is the
+  expectation, and any of those services sitting in `Created` / `Exited` / `Dead` / `Paused` /
+  `Restarting`, or absent from `compose ps`, fails the deploy loudly instead of leaving the
+  worker tier down silently. The expectation is deliberately the **profile-filtered** service
+  list: `compose ps` labels containers by PROJECT, not by profile, so the standalone
+  `selenium-chrome` the Grid overlay parks for rollback sits `Exited` indefinitely and must
+  never fail a deploy it is not part of. `flyway` is excluded too — it is a `run --rm` one-shot.
 - **Persists `IMAGE_TAG` / `.last_good_tag` immediately after the edge flip**, while the
   serving tier is live on the new tag. A later worker-tier failure is a partial deploy, but
   the box's recorded baseline matches what is actually running so the next deploy diffs
-  against the correct starting point.
+  against the correct starting point — and a manual `compose up -d` recovery uses the right tag.
+
+The legacy full-rollback path recreates the **same** worker tier, so it goes through
+`converge_stack` too — a rollback is the worst place to hit the race, since something has
+already gone wrong by then.
 
 Rollback still exists at every step; the only deploy with a brief blip was the first cutover
 (recreating `web_app` from a FastAPI container into the nginx edge). `scripts/rollback.sh` remains
--- a/docs/zero-downtime-deploys.md
+++ b/docs/zero-downtime-deploys.md
@@ -36,6 +36,24 @@ Cloudflare Tunnel (dashboard ingress: http://web_app:8000  — UNCHANGED)
    `/opt/lem/.active_color`.
 4. `up -d --remove-orphans` converges workers/beat and the now-standby color onto the new tag.
 
+### Worker-tier resilience (issue #831)
+
+The final `up -d --remove-orphans` can hit a Docker race: compose renames the old
+container before creating the replacement, and a concurrent `docker exec` (or compose's
+own bookkeeping) can reference the old ID in that window, producing `No such container`.
+That aborts the converge mid-tier and leaves workers in `Created` — a silent worker outage.
+
+`scripts/deploy.sh` now:
+
+- **Retries once** if the converge output contains `No such container`.
+- **Verifies the final state** by enumerating expected compose services and checking that
+  none are stuck in `Created` / `Exited` / `Dead`. If verification fails, the deploy exits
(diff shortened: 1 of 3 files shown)


gitchrisqueen:

Reviewed as the gate for #840 (issue #831). Verified the two load-bearing Docker CLI assumptions against compose source rather than trusting them: `config --services` **is** profile-filtered ([#8676](https://github.com/docker/compose/issues/8676)), and `ps --format` **does** accept arbitrary Go templates with `.Service`/`.State` (v2 `cmd/compose/ps.go` → `cliformatter.Context`). Both hold. What did not:

- **`verify_stack_running` would have failed every Grid deploy**: it compared against the *profile-filtered* `config --services` list but scanned container states across the whole project. `compose ps` labels containers by PROJECT, not by profile — so the standalone `selenium-chrome` that `docker-compose.grid.yml` *deliberately* parks behind `profiles: ["standalone"]` for instant rollback sits `Exited` indefinitely and would have tripped the Exited check on every run. State checks are now intersected with the expected service list (one `ps -a` read instead of two), and also catch `paused`/`restarting`/`removing`. Regression test added (`profile_leftover` mode).
- **`{{.Service}}\t{{.State}}` was a silent-failure risk**: it depended on the CLI expanding `\t` inside a Go template. If it did not, awk sees one field, `$2` is empty, and the Created check never fires — i.e. the exact bug this PR exists to catch, silently disabled, with the mocked test unable to see it (the fake emitted real tabs). Changed to a space separator: neither a compose service name nor a container state can contain a space, so this no longer hinges on CLI escape handling.
- **`converge_stack` deleted the evidence**: it redirected compose output to a temp file and `rm`-ed it on success — and on a *retried* attempt too. Those per-container `Recreated` / `Error response from daemon: No such container` lines are precisely what diagnosed this race both times it happened; the deploy log lost them. Every attempt’s output is now echoed, success or failure.
- **Rollback path still had the bare converge**: line ~322’s legacy full-rollback called `${COMPOSE} up -d --remove-orphans` directly. It recreates the *same* worker tier and runs when something has already gone wrong. Routed through `converge_stack` (mirrors the `drain_workers` lesson already recorded in that file’s comments).
- **Test gap on acceptance criterion 4**: the tests exercised the two functions in isolation, so nothing would fail if the `converge_stack` / `verify_stack_running` calls were dropped from `main()` — the regression the criterion asks about. Added `TestDeployScriptWiring` (verify is called and aborts, no bare bulk converge remains, tag persist precedes the worker converge).

Checked and **not** findings: `main()` wrapping keeps `set -e` live (called unconditionally, not in a condition context) and propagates its exit code; the source-guard works for the CI invocation form; `flyway` is a `run --rm` one-shot and is correctly excluded; replicated `selenium-node-chrome` collapses under `sort -u`; the issue’s “consider skipping `maint`’s `docker exec` once the recreate has begun” is moot — every `maint`/`stop` call completes inside `drain_workers`, before `converge_stack` starts, so there is no concurrent exec left to skip.

Close is honest: all four acceptance criteria on #831 are met and I ticked the boxes in the issue body. No later phase is declared.

`tests/unit/test_deploy.py tests/unit/app/test_worker_shutdown.py`: 36 passed.

gitchrisqueen:

PR #848, opened 2026-08-01
fix(security): don't let a failed cookie write delete the user's LinkedIn password (refs #745)

gitchrisqueen:
Adversarial review of PR #807 (issue #745, secrets at rest). The review tick fired after #807 had already merged, so the two findings land here instead.

Both are in the seam between the cookie write and the new cookie-only migration path.

## 1. A failed cookie write reported success — and then deleted the password

`_store_cookie_rows` swallows `mysql.connector.Error` per row (pre-existing, and fine when the only consequence was a missing cookie). #807 made that consequence destructive: `store_cookies` returned `None`, `store_linkedin_li_at` returned `True` regardless, and `POST /user/linkedin-cookie` treats that `True` as "the cookie that replaces the password is safely stored" before calling `clear_user_linkedin_password`.

So any per-row driver error (FK violation from a concurrent delete, deadlock, connection blip) produced: no `li_at` stored, `200 "LinkedIn session saved"`, **and the user's only remaining engagement login deleted**. The PR body explicitly promises the opposite — "Only drop it once the cookie that replaces it is safely stored".

`store_cookies` now returns `False` when any row failed, and `store_linkedin_li_at` propagates it, so the endpoint raises 500 and never reaches the drop.

## 2. An unbindable cookie wrote plaintext `li_at` the backfill can't see

`encrypt_secret` returns the value **unchanged** when `user_id is None` (documented — there is no AAD to bind to). `store_cookies` gets its `user_id` from `get_user_id(user_email)`, which returns `None` on any DB error, not just an unknown email. So a transient failure immediately after a successful login wrote every cookie — `li_at` included — as **plaintext**.

That row is also unreadable forever (`get_cookies` JOINs `users`), and `encrypt_secrets_at_rest` scans `cookies WHERE user_id IS NOT NULL`. Net effect: a live LinkedIn session sitting in the clear in every `mysqldump`, invisible to `plaintext_remaining` — the number `docs/secrets-at-rest.md` names as the gate for flipping `ENCRYPTION_REQUIRED`. The operator would read "0 still unprotected" and fail-close over the top of it.

- `_store_cookie_rows` now refuses a write it cannot bind (logs an error, reports the rows as failed — which also feeds finding 1's guard). Nothing is lost: such a row could never be read back.
- `encrypt_secrets_at_rest` counts pre-existing orphan rows into `plaintext_remaining` as a new `orphaned` stat, and names the remedy (`DELETE FROM cookies WHERE user_id IS NULL`) — deletion, not encryption, since they can't be encrypted.

## Testing

`8723 passed` (`poetry run pytest tests/unit -q`). Three new/rewritten tests in `tests/unit/utilities/test_db_secret_encryption.py`:

- `test_unbindable_cookie_is_refused_not_written_in_the_clear` — replaces `test_orphan_cookie_without_a_user_is_still_stored`, which asserted the old (plaintext-writing) behaviour.
- `test_failed_write_is_reported_so_the_password_is_never_dropped`
- `test_orphaned_plaintext_cookies_keep_the_gate_off_zero`

`docs/secrets-at-rest.md` gains the orphan-row section; the beat task's warning names the orphan count.

Remaining on #745: 2b and 2c, unchanged by this PR.

### DIFF
--- a/docs/secrets-at-rest.md
+++ b/docs/secrets-at-rest.md
@@ -70,10 +70,23 @@ overwritten, because a corrected key might still recover it.
 Watch the number it logs:
 
 ```
-Secret encryption backfill: 12 rewritten, 0 failed, 0 still unprotected
+Secret encryption backfill: 12 rewritten, 0 failed, 0 orphaned, 0 still unprotected
 ```
 
-`plaintext_remaining > 0` is logged as a WARNING. Run it on demand with:
+`plaintext_remaining > 0` is logged as a WARNING.
+
+**Orphaned rows** are `cookies` rows with no `user_id`. They cannot be encrypted (there is no user
+to bind the AAD to) and cannot be read back (`get_cookies` JOINs `users`) — a dead plaintext
+session. They count toward `plaintext_remaining` so the fail-closed gate can never read 0 while one
+exists, and the remedy is deletion, not encryption:
+
+```sql
+DELETE FROM cookies WHERE user_id IS NULL;
+```
+
+New ones can no longer be created: `_store_cookie_rows` refuses a write it cannot bind.
+
+Run the pass on demand with:
 
 ```bash
 docker exec celery_worker python -c \
--- a/src/cqc_lem/app/run_scheduler.py
+++ b/src/cqc_lem/app/run_scheduler.py
@@ -1505,7 +1505,8 @@ def auto_encrypt_secrets_at_rest(self):
         return "Encryption disabled (no LEM_SECRET_KEY) — nothing done"
     if stats.get("plaintext_remaining"):
         log_warning(f"{stats['plaintext_remaining']} secret(s) still unencrypted at rest "
-                    f"({stats.get('failed', 0)} failed to re-encrypt)",
+                    f"({stats.get('failed', 0)} failed to re-encrypt, "
+                    f"{stats.get('orphaned', 0)} orphaned cookie row(s) to delete)",
                     task_name="auto_encrypt_secrets_at_rest")
     return (f"Re-encrypted {stats.get('rewritten', 0)} secret(s); "
             f"{stats.get('plaintext_remaining', 0)} still unprotected")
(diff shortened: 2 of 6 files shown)


gitchrisqueen:

Reviewed the diff against `api/main.py`'s `/user/linkedin-cookie` handler, `helper.py`'s login path, `crypto.encrypt_secret`, and the `cookies` schema (V2/V4/V29). Both claimed fixes hold up: `store_linkedin_li_at` now returns `False` on any swallowed per-row error so the endpoint 500s before `clear_user_linkedin_password`, and the orphan rows counted into `plaintext_remaining` really are plaintext-only (the FK is `ON DELETE CASCADE`, so an encrypted row can never become an orphan). `Closes` is honest — the PR claims `refs #745` and names 2b/2c as remaining. CI is green on every required check; 8725 unit tests pass locally with my change.

- **The other writer of `li_at` still announced success on a lost write** (`utilities/linkedin/helper.py`): this PR made `store_cookies`' return value load-bearing, but the Selenium login path — the *other* place the cookie that replaces the password is written — discarded it at both call sites and printed `"Cookies stored to DB!"` unconditionally. So the exact two failures the PR closes in the API path stayed silent here: an unbindable write (new `user_id is None` refusal) or a swallowed `mysql.connector.Error` wrote nothing, and the log said it was saved. The consequence is delayed rather than immediate — this run is fine, the next finds no `li_at`, falls back to a password login and trips LinkedIn's new-device challenge, which is the thing #745 exists to avoid. Both sites now go through `_persist_session_cookies`, which `log_error`s the failure and only claims the store when it landed. Two tests in `tests/unit/utilities/linkedin/test_helper.py::TestPersistSessionCookies`.

Considered and not raised: `store_cookies([])` returning `True` (unreachable from `store_linkedin_li_at` — the endpoint validates `len(li_at) >= 20`); the daily orphan alarm firing from both `db.py` (ERROR) and the beat task (WARNING) for one standing condition — noise, and the beat warning predates this PR.

gitchrisqueen:

PR #851, opened 2026-08-01
feat(observability): group a post's whole generation chain into one $ai_trace (closes #746)

### LINKED ISSUE #746: PH2 — LLM analytics: $ai_trace/$ai_span pipeline traces — Phase 2 (dropped follow-up of #647)
## Why this exists

#647 (PH2) shipped the LiteLLM→PostHog native callback in PR #679 and was closed — but the issue's scope had a second half that never landed:

> - Stretch (separate PR ok): trace structure — wrap multi-call pipelines (post generation: research→draft→review→humanize) in a `$ai_trace`/`$ai_span` hierarchy via posthog-python AI helpers so a post's full generation chain reads as one trace.

The shipped doc records the gap and points at an issue that is now closed — `docs/llm-analytics.md` §"Not in scope here":

> `$ai_trace` / `$ai_span` hierarchies — wrapping a post's full generation chain (research → draft → review → humanize) into one readable trace. Every event already carries a `$ai_trace_id`, but it is per-call; grouping them is app-side work via the `posthog-python` AI helpers. **Tracked as the stretch half of #647.**

With #647 closed, nothing tracks it. Verified on `origin/main`: no `$ai_span`/trace grouping exists in `src/`.

**Why it matters:** today every LLM call is an isolated `$ai_generation`. The per-post question the analytics exist to answer — *what did one post actually cost, end to end, across research → draft → review → humanize?* — cannot be asked without grouping. It is a cost/quality-attribution gap, not a user-facing bug, hence low priority.

## Scope
- Give each multi-call pipeline one trace id and emit spans under it via the `posthog-python` AI helpers, threaded through the ONE client (`utilities/ai/client.py`) alongside the existing `llm_attribution()` scope — do **not** add a parallel attribution mechanism at call sites.
- Cover at minimum the post-generation chain (research → draft → review → humanize); newsletters and comments follow the same shared core, so they should inherit it for free.
- Keep the LiteLLM-side `$ai_generation` stream untouched — this is app-side grouping only.
- Update `docs/llm-analytics.md` §"Not in scope here" when it lands.

## Acceptance
- [x] One post's full generation chain reads as a single trace in PostHog LLM analytics, with per-step spans.
- [x] Trace/span ids are set in the shared client, not at individual call sites; a call that forgets them still works (fail-soft).
- [x] Unit tests assert the trace id is stable across a pipeline and distinct between pipelines.
- [x] `docs/llm-analytics.md` no longer lists this as out of scope.

Original issue: #647 (stretch half). Shipped half: PR #679.

gitchrisqueen:
## Why

#647's stretch half never landed and vanished when #647 closed: a post is **six** model calls
(research → draft → refine → humanize → authenticity → review), each landing in PostHog as its own
isolated `$ai_generation` with its own proxy-minted trace id. The per-post question the LLM
analytics exist to answer — *what did THIS post cost, end to end?* — could not be asked.

Grouping is app-side work, because only the app knows where a pipeline starts.

## What

**`utilities/observability.py`** (the ONE place events are emitted) gains `llm_trace()` /
`llm_span()` and their decorator forms, emitting `$ai_trace` / `$ai_span`:

- `@llm_pipeline` marks a function that IS one pipeline and **supersedes `attribute_llm_cost`** on
  it — it opens the trace *and* the same `llm_attribution()` scope, reading the user id off the
  call's own `user_id` argument. Three roots: `create_text_post`, `generate_newsletter_edition`,
  `generate_ai_response`.
- `@llm_step` goes on the **step function**, never at a call site — newsletters and comments draw
  draft/research/humanize/authenticity from the same shared content core as posts, so decorating the
  core is what gives every pipeline a legible trace without touching one caller. Spans: `research`,
  `draft`, `refine`, `hook`, `humanize`, `authenticity`, `review`.

**`utilities/ai/client.py`** puts the ids on the wire. Two invariants worth knowing:

- Tracing rides in its **own** hook, not inside `_attach_attribution` — that one bails out whenever
  the caller supplied its own metadata, and `_call_llm` always does, so folding them together would
  have silently excluded nearly every generation LEM makes.
- **Two wires**, because LiteLLM's PostHog logger sources them from two places: the trace id must be
  the `x-litellm-trace-id` *header* (it becomes the proxy's `litellm_trace_id` → `$ai_trace_id`;
  sending it as metadata does nothing, the logger overwrites it from its own standard logging
  payload), while the parent span rides in `metadata.parent_run_id` → `$ai_parent_id`.

Behavioural edges, all covered by tests: a trace opened inside an open trace becomes a **span**
(`create_text_post` recurses for post-type fallbacks, and two half-traces of one post answer
nobody); a span outside a pipeline is a **no-op**; a capture failure is swallowed at DEBUG; and
`LLM_TRACING_ENABLED=false` mints no ids at all.

The proxy-side `$ai_generation` stream is untouched — no extra event, no changed property, no
LiteLLM config change.

## Testing

- `tests/unit/utilities/test_llm_traces.py` (32 tests) — trace id stable across a pipeline and
  distinct between pipelines, nesting/parenting, kill switch, error paths, fail-soft, and a
  parametrized guard that every pipeline root and shared-core step is still decorated (strip one and
  the chain silently splinters with nothing to say a step went missing).
- `tests/unit/utilities/ai/test_client_tracing.py` (10 tests) — builds **real** requests through the
  OpenAI SDK, so an SDK upgrade that moves the injection point fails CI rather than quietly dropping
  every trace. Covers header + `parent_run_id`, caller-set metadata, embeddings, raw provider models,
  untraced calls, and each hook surviving the other throwing.
- Full suite: `8820 passed`.

Docs: `docs/llm-analytics.md` §"Not in scope here" replaced with the pipeline-traces posture;
`CLAUDE.md` map entry and `.env.example` updated.

Closes #746

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -584,6 +584,11 @@ POSTHOG_HOST=https://us.i.posthog.com
 # Both are also passed to the litellm container (as POSTHOG_API_KEY / POSTHOG_API_URL), where the
 # native posthog callback emits one $ai_generation per LLM call — issue #647, docs/llm-analytics.md.
 # Prompts/completions are redacted before they reach it (litellm_settings.turn_off_message_logging).
+# Pipeline traces (issue #746): group the 5-6 calls one post/edition/comment takes into ONE
+# $ai_trace with per-step $ai_span children, so "what did this post cost end to end?" is answerable.
+# App-side grouping only — the $ai_generation stream itself is untouched. false = no trace ids are
+# minted and the proxy events are exactly what they were before tracing shipped.
+LLM_TRACING_ENABLED=true
 # Minimum log level forwarded to PostHog (DEBUG/INFO/WARNING/ERROR/CRITICAL). Default: ERROR.
 # WARNING is the useful floor once escalation is on (below): the warning body is the CONTEXT for the
 # escalated $exception. DEBUG floods — it put 2,185 info rows into PostHog Logs in 48h.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -164,11 +164,18 @@ Track events via `utilities/observability.py` (`track_llm_call` / `track_task` /
 Each subsection below is a one-paragraph invariant; rationale, contracts, sample code, and edge
 cases live in the pointed-to doc. Plan with this section, drill in with the doc.
 
-### LLM analytics (issue #647)
+### LLM analytics (issue #647, traces #746)
 Two streams, never summed: `llm_call` (app, cost ESTIMATE — every money question reads this) vs
 `$ai_generation`/`$ai_embedding` (proxy-native, post-fallback, provider-priced — latency/error/
 volume). Attribution lives in `utilities/ai/client.py` (the ONE client); `_attach_routing_metadata`
-must stay. Full posture: `docs/llm-analytics.md`.
+must stay. **Pipeline traces** group a post/edition/comment's 5–6 calls into ONE `$ai_trace` with
+per-step `$ai_span`s: `@llm_pipeline` on the three roots (`create_text_post`,
+`generate_newsletter_edition`, `generate_ai_response` — it SUPERSEDES `attribute_llm_cost`),
+`@llm_step` on the SHARED-core step functions (never at a call site, so newsletters/comments inherit
+it). The client sends the ids two ways because LiteLLM reads them from two places — trace id in the
+`x-litellm-trace-id` HEADER (metadata would be overwritten), parent span in `metadata.parent_run_id`.
+Nested trace = span; span with no trace = no-op; `LLM_TRACING_ENABLED=false` mints nothing.
+Full posture: `docs/llm-analytics.md`.
 
 ### Error tracking (issue #648)
 Logs (`logger.py` → PostHog Logs) carry message+context; `$exception` is the grouped/fingerprinted
--- a/docs/llm-analytics.md
+++ b/docs/llm-analytics.md
@@ -156,5 +156,12 @@ Two invariants worth knowing before you add either:
 plus LEM's `feature` / `user_id` and `$ai_is_error` / `$ai_error` when the step raised. distinct_id
 follows the same rule as every other event here: the user id, or the `"system"` sentinel.
 
+**The root span IS the trace.** `llm_trace()` uses the trace id as its own span id, so a top-level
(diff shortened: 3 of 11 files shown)


gitchrisqueen:

- **The trace would have rendered empty** (`observability.llm_trace`): the root span was minted as its own uuid, so every top-level `$ai_span` — and the `parent_run_id` of any generation made directly under the root — pointed at an id no node in the trace carries. PostHog builds a trace's child list with `toString($ai_parent_id) = toString($ai_trace_id)` and totals `total_latency` over that same set (`posthog/hogql_queries/ai/traces_query_runner.py`; the public `$ai_span` property table says `$ai_parent_id` is "`trace_id` or another `span_id`"). Net effect: every event carries the right `$ai_trace_id`, the trace opens with **no children and zero latency** — exactly the question #746 exists to answer. Fixed by making the root span *be* the trace (`span_id = trace_id`), locked in on both wires (`test_a_top_level_span_parents_onto_the_trace_id_itself`, `test_a_call_directly_under_the_trace_parents_onto_the_trace_itself`) and written into `docs/llm-analytics.md` — the failure mode is silent, the events still ship, they just never nest.

Verified, not findings:
- The two-wire claim holds against LiteLLM's own logger: `$ai_trace_id` ← `standard_logging_object["trace_id"]` (i.e. the `x-litellm-trace-id` header), `$ai_parent_id` ← `metadata.get("parent_run_id")` (`litellm/integrations/posthog.py`). `$ai_span_name` is a valid trace name — PostHog reads `ifNull($ai_span_name, $ai_trace_name)`.
- `llm_attribution(user_id=None)` never clears an outer value, so the two roots without a `user_id` parameter (`generate_newsletter_edition`, and callers that pass none to `generate_ai_response`) keep the caller's attribution; the forced `feature=` matches what those call sites already set (`FEATURE_NEWSLETTER` / `FEATURE_COMMENT`).
- No threading in the content path, so the contextvar carries; `_attach_trace` is idempotent across SDK retries; `LLM_TRACING_ENABLED` defaults on and `env_file: .env` reaches every app service.
- Close is honest: all four acceptance criteria of #746 are covered by the diff, no later phase left in prose.

`55 passed` on the trace/attribution suites, `1337 passed` across `tests/unit/utilities/test_observability.py` + `tests/unit/utilities/ai`.

gitchrisqueen:

PR #852, opened 2026-08-01
fix(feedback): stop a GitHub DNS blip from filing itself as a defect (closes #767)

### LINKED ISSUE #767: fix(errors): ConnectionError: HTTPSConnectionPool(host='api.github.com', port=443): Max retries exceeded with url:  (3x)
## Why
PostHog error tracking grouped this exception into an issue that is still active: **ConnectionError**

> HTTPSConnectionPool(host='api.github.com', port=443): Max retries exceeded with url: /repos/christopherqueenconsulting/linkedin_engagement_manager/issues/727 (Caused by NameResolutionError("HTTPSConnection(host='api.github.com', port=443): Failed to resol

- Occurrences (last 24h): **3** across 1 distinct actor(s)
- First seen: `2026-07-28T08:00:05.694000Z` · last seen: `2026-07-28T08:00:05.792000Z`
- Celery task: `cqc_lem.app.run_scheduler.auto_file_feedback_issues`
- SDK: `posthog-python`

[Open the issue in PostHog](https://us.posthog.com/project/475262/error_tracking/019fa7bc-f4a9-7131-a3df-af9e618851dc) — it has the stack trace, the grouped occurrences and the affected people.

## Scope

## Files
- `tests/unit/`

## Acceptance
- All required CI gates pass.

Auto-filed from PostHog Error Tracking. Dedup marker (do not remove): `posthog-issue-019fa7bc-f4a9-7131-a3df-af9e618851dc`

gitchrisqueen:
## What & why

PostHog grouped a live `ConnectionError` into an active issue: `HTTPSConnectionPool(host='api.github.com') ... NameResolutionError` — 3 occurrences inside 100ms during the `auto_file_feedback_issues` beat, on `GET /issues/727` (the #718 repair sweep).

`github_request` treated **every** failed call as `log_error(..., exc=e)`. GitHub is reached best-effort here — filing is retried on the next pass and the feedback row is never lost — so a DNS/connect blip is the network, not a defect in this loop. Logging it at ERROR captured a `$exception` per lost request, and the daily error→issue cron filed the outage as a bug.

Now, per `utilities/CLAUDE.md` ("once is a warning, repeatedly is a defect"):

- **Transport failures are retried in place** — the OSError family (DNS, connect, reset, timeout; every `requests` transport exception subclasses `RequestException` → `IOError`), 3 attempts with 0.5s linear backoff. A blip that recovers is reported nowhere.
- **Exhausted retries log ONE `log_warning`**, so `log_escalation` makes the call: one blip stays a warning, GitHub unreachable across three passes is re-emitted at ERROR + `$exception` and legitimately becomes an issue.
- **Anything that is not the transport is untouched** — still `log_error` immediately, no retries. HTTP ≥300 responses (403/422 — the deterministic #598/#718 token problems) still log at ERROR exactly as before.
- **A 60s process-local cool-off short-circuits the rest of the pass.** Without it, one outage costs three warnings inside a single pass — which *is* the escalation threshold, so the outage would still file itself.
- **The repair sweep defers with one warning** when the transport is down, instead of stopping with the #718 `FEEDBACK_GITHUB_TOKEN needs Issues: Read and write` error — which, during a network outage, is a lie.

Shared helper, so the shipped-fix notifier (#502) and the capacity alerter get the same posture.

## Testing

- New `TestTransportFailures`: DNS failure retried then warned (never errored), a recovered blip reported nowhere, a non-transport exception still an immediate error with no retries, one outage = one warning + short-circuited pass, success clears the cool-off.
- New repair-sweep test: an unreachable transport defers after ONE read with a warning and never emits the token-permission error.
- Autouse fixture resets the process-global cool-off between tests.
- `poetry run pytest tests/unit -q` → **8826 passed**.

Closes #767

### DIFF
--- a/src/cqc_lem/utilities/feedback/issue_service.py
+++ b/src/cqc_lem/utilities/feedback/issue_service.py
@@ -45,6 +45,7 @@ Env:
 import json
 import os
 import re
+import time
 from typing import Optional
 
 from cqc_lem.utilities.ai.content_framework import as_vector, cosine_similarity, text_similarity
@@ -81,6 +82,21 @@ RATE_LIMIT_WINDOW_HOURS = 24
 MAX_CLUSTER_CANDIDATES = 100
 MAX_TITLE_CHARS = 120
 GITHUB_TIMEOUT_SECONDS = 20
+
+# Issue #767: a DNS/connect blip is the NETWORK, not a defect in this loop — GitHub is reached
+# best-effort and the row is left for the next pass either way. It is retried in place, then
+# reported as a WARNING so `log_escalation` decides: one blip stays a warning, GitHub unreachable
+# for three passes running becomes the ERROR/$exception it deserves.
+GITHUB_RETRY_ATTEMPTS = 3
+GITHUB_RETRY_BACKOFF_SECONDS = 0.5
+# Once the transport is down, every other call in the same pass fails the same way. Short-circuiting
+# them keeps ONE outage to ONE warning instead of one per request — three of which is exactly the
+# escalation threshold, i.e. an outage would file itself as a defect.
+GITHUB_UNREACHABLE_COOLDOWN_SECONDS = 60.0
+
+# Monotonic deadline for the cool-off above. Process-local on purpose: it exists to collapse the
+# retries of ONE pass, not to coordinate workers.
+_unreachable_until = 0.0
 
 # Issue #718: the label/assignee/comment sub-resource endpoints 403'd for weeks while `POST /issues`
 # kept succeeding, so every auto-filed issue landed label-less. Named in every failure log so the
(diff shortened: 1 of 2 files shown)


gitchrisqueen:

Reviewed the diff against #767's acceptance (exception downgraded to a warning, unit-tested, scoped) plus the surrounding `issue_service` seams. The close is honest — #767 has no later phase and every acceptance box is covered. Three real defects in the new retry path, all fixed on this branch (`pytest tests/unit -q` → 8830 passed):

- **Retrying non-idempotent writes could file duplicate issues/comments.** The retry loop applied to every method. A `POST /issues` that failed *after* GitHub accepted it (read timeout, connection reset mid-response) is indistinguishable from one that never landed, so a blip could open up to 3 copies of the same auto-filed issue — the exact thing this module's dedup exists to prevent. Fixed: `GITHUB_IDEMPOTENT_METHODS = ("GET", "HEAD")`; writes get one attempt and are still classified as transport (warning + cool-off), which fully covers #767 since the reported incident was `GET /issues/727`.
- **`isinstance(exc, OSError)` classified our own bugs as "the network".** `requests.exceptions.MissingSchema` / `InvalidURL` / `TooManyRedirects` / `InvalidHeader` all subclass `RequestException` → `IOError`, so a malformed repo/URL config would be retried 3×, reported as "GitHub unreachable", **and** trip the 60s cool-off that silently skips the rest of the pass — hiding the one fault worth alerting on. Fixed: `NON_TRANSPORT_REQUEST_ERRORS` named out by class name (keeps `requests` a lazy import).
- **The #718 token hint still lied during an outage — on the write paths.** The PR fixed this for the repair sweep's reads, but `create_github_issue`'s label/assignee attach and `repair_filed_issue`'s label attach still logged `FEEDBACK_GITHUB_TOKEN needs Issues: Read and write` at ERROR when the transport died *after* a successful create. That's reachable during the exact #767 outage, and it's still an ERROR → `$exception` → an auto-filed defect with the wrong diagnosis. Fixed: `_log_lost_write()` warns while `github_unreachable()` stands, and errors with the hint otherwise.

Verified and left alone: the escalation path does work (beat runs every 30 min, threshold 3-in-24h, and the repair-deferral message is digit-free so its dedup key is stable); the cool-off correctly prevents the capacity alerter from opening a duplicate issue when its lookup GET is lost; and the API `admin/feedback/{id}/review` path's worst-case latency is unchanged for writes now that they aren't replayed.

Tests added: write-never-replayed, `MissingSchema`-is-ours-not-the-network, and both lost-write paths warning instead of blaming the token.

gitchrisqueen:

PR #854, opened 2026-08-01
feat(auth): renew LinkedIn tokens daily and show the expiry countdown (closes #600)

### LINKED ISSUE #600: feat(auth): Auto-refresh LinkedIn token and show expiration countdown
## Why
User sees the SPA warning “Your LinkedIn token is expiring soon. Please reconnect to keep posting.” but did not receive an email. They request automatic token refresh, longer token lifespan, and UI indication of remaining time before expiration.

Auto-filed from in-app feedback #2 by the feedback→auto-work loop (`docs/launch-and-marketing-plan.md` §B.2/B.3). Classifier: enhancement/medium, risk `none`, confidence 0.92.

## Scope
- Address the reported enhancement in the `auth` area of the product.

## Files
- `src/cqc_lem/api/main.py`
- `src/cqc_lem/utilities/linkedin/token_refresh.py`
- `tests/unit/`

## Acceptance
- All required CI gates pass.

gitchrisqueen:
## What & why

The reporter saw *"Your LinkedIn token is expiring soon. Please reconnect to keep posting."* in the SPA, got **no email**, and asked for automatic refresh, a longer lifespan, and a visible countdown. All three were real gaps:

- **Nothing ever refreshed a token except an SPA page view.** `attempt_token_refresh` only ran inside `GET /user/token_status`, so a user who didn't open the account page simply lapsed.
- **No email existed** for token expiry at all — the SPA banner was the only notice.
- **No countdown.** The connection card renders *nothing* while the token is healthy, and the warning banner never said how long was left.

## Changes

**One decision core.** `resolve_token_status(user_id, auto_refresh=True)` in `utilities/linkedin/token_refresh.py` is now the ONE place a user's token state is decided. `GET /user/token_status` is a one-liner over it and the new beat uses the same call, so the countdown a user sees and the countdown that emails them cannot disagree. It adds `days_remaining` and `can_auto_refresh` to the payload (existing keys unchanged).

**Rolling refresh.** New daily beat `refresh-linkedin-tokens` → `auto_refresh_linkedin_tokens` at **08:30 UTC**, deliberately *before* the 09:00 missing-session pass so a token renewed here never produces a reconnect email half an hour later. It renews every connected subscriber inside `EXPIRY_WARNING_DAYS` (30) and emails the ones it could not renew.

On "longer token lifespan": LinkedIn fixes the access token at 60 days and gives no way to request more, so a rolling refresh *is* the only thing that makes the connection outlive it. The UI and email copy now say that plainly instead of implying we could ask for longer.

**The missing email.** `send_linkedin_token_expiring_email` + `notify_linkedin_token_expiring`, throttled per user through Redis (`LINKEDIN_TOKEN_EMAIL_THROTTLE_DAYS`, default 7) so the daily beat can't spam. It **fails open** — the bug being fixed was a warning with no email behind it, so a Redis outage degrades to at most one email a day, never back to silence.

**Countdown in the SPA** (`LinkedInLoginCard.tsx`): the warning banner now names the days and the date; a healthy connection renders a compact `Connected to LinkedIn · 44 days left — expires Sep 14, 2026` strip where it previously rendered nothing.

**Details worth a look**
- `get_linkedin_token_user_ids()` is new rather than reusing `get_active_user_ids()`: the latter requires an *unexpired* token, i.e. it filters out exactly the users the renewal pass most needs to reach.
- `refresh_token_usable()` was factored out of `attempt_token_refresh` so a provably-lapsed refresh token no longer counts as "refresh attempted".
- `days_remaining` is `None`, never `0`, when the expiry can't be read — the SPA renders "Expiration date unknown" rather than an alarming "0 days".
- "No refresh token" is LinkedIn's ordinary case (refresh tokens go to approved apps only), so it logs **DEBUG** — a `log_warning` there would escalate working behaviour into a filed defect under the recurrence rule.
- `myprint` in `token_refresh.py` replaced with the structured logger.

## Testing

- `tests/unit/utilities/linkedin/test_token_refresh.py` — `days_until_expiry` (floors, never negative, `None` ≠ 0), `refresh_token_usable`, and 7 `resolve_token_status` cases (disconnected, healthy-not-refreshed, refreshed-and-restated, failed refresh, no refresh token, `auto_refresh=False`).
- `tests/unit/app/test_token_refresh_beat.py` — refreshed user not emailed, unrefreshable user emailed with the countdown, expired user emailed, healthy user untouched, one user's exception doesn't stop the pass, empty roster no-op, and the beat's 08:30 slot.
- `tests/unit/utilities/test_token_expiry_notify.py` — throttle claims/blocks, configurable window, `0` disables, Redis outage and no-Redis both still email, no-email user, swallowed exception, plus email copy (singular/plural, already-expired subject, unknown-expiry avoids a number).
- `tests/unit/api/test_api_coverage_misc.py` token-status tests rewritten against real token rows instead of patching the pure helpers, plus a new no-refresh-token case.
- Full local run: **8864 passed**. SPA `tsc --noEmit` clean.

## Live validation needed

`risk:live-linkedin` — the refresh POST to `https://www.linkedin.com/oauth/v2/accessToken` can only be proven against a real account with a real refresh token. Everything else is exercised headless.

Closes #600


gitchrisqueen:
Held (`needs-human`, risk: live-linkedin). The refresh call to `https://www.linkedin.com/oauth/v2/accessToken` can only be proven against a real account with a real refresh token — and two of the choices below decide how loud this feature is for users we *can't* refresh.

### 1. When should the reconnect email start firing?
The beat emails a user once their token is inside `EXPIRY_WARNING_DAYS` **and** the renewal attempt didn't fix it. That constant is **30 days** today — half the life of a 60-day token — and it's also the threshold for the SPA banner.
- **A. Leave it at 30 days, throttled to one email per 7 days** — a user who can't be auto-refreshed gets ~4 emails across the last month of the token. ✅ *recommended*
- **B. Keep the 30-day SPA banner but only email inside 14 days** — quieter inbox, ~2 emails, less lead time to act.
- **C. Email only inside 7 days** — one email, right before it breaks; risks landing while the user is away.

### 2. Should the healthy-state countdown strip be visible on the account page?
Today a healthy connection renders **nothing** — that's why the reporter had no way to see time remaining until the warning fired. This PR adds a compact `Connected to LinkedIn · 44 days left — expires Sep 14, 2026` row with a small "Reconnect" link.
- **A. Ship the strip as-is** — the countdown is always visible, which is literally what was asked for. ✅ *recommended*
- **B. Show the countdown only inside the last 30 days** — account page stays emptier, but reverts to the "no indication" state the report was about.
- **C. Drop the strip; keep only the improved warning banner** — smallest UI change, doesn't satisfy the "UI indication of remaining time" ask.

**My recommendation: `1A 2A`.** The report was "warned in-app, told nothing by email, couldn't see how long I had" — 30 days of lead time with a 7-day throttle and an always-visible countdown answers all three, and the throttle keeps it to roughly four emails in the worst case.

*Note on the third ask (longer token lifespan): LinkedIn fixes the access token at 60 days and offers no way to request more, so it isn't a dial I can turn. The rolling daily refresh in this PR is what actually makes the connection outlive it, and the UI/email copy now says that plainly instead of implying otherwise.*

gitchrisqueen:
Ok

gitchrisqueen:
Addressed your review: read your `ok` as **1A + 2A** and applied both.

**1A — email window stays at 30 days, throttled to one per 7 days.** Verified against the code rather than the PR prose: `EXPIRY_WARNING_DAYS = 30` (`utilities/linkedin/token_refresh.py`) is the single threshold behind *both* the SPA banner and the beat's email, and `LINKEDIN_TOKEN_EMAIL_THROTTLE_DAYS` defaults to `7` in `utilities/notifications.py` and `.env.example`. Worst case is ~4 emails across the token's final month.

**2A — the healthy countdown strip ships as-is.** `showHealthyCountdown` is gated only on "connected and not in the warning window", so `Connected to LinkedIn · 44 days left — expires Sep 14, 2026` renders whenever the connection is healthy, not just near expiry.

Both were already the shipped behaviour, so **no behaviour changed**. The one commit (`a879d86`) records the decision at the two constants it governs — same convention as the `issue #745, decision 2A` marker already in `LinkedInLoginCard.tsx` — so nobody re-litigates 30-vs-14 days or re-hides the strip behind the warning window later.

**What I did NOT change**, so it isn't read as silently done:
- The 60-day token lifespan. As noted on the decision comment, LinkedIn fixes it and offers no way to request longer; the daily rolling refresh is what actually makes the connection outlive it, and the UI/email copy says that rather than implying we could ask for more.
- The `CLAUDE.md ≤ 40,000 chars` check, currently red. It is **not** one of the required checks, and it is not this PR's doing: `CLAUDE.md` was already 41,269 chars at this branch's merge base, over the 40k cap before a word of #600 was added. `main` has since trimmed it to 30,666, so the post-merge file lands around 31.6k and the check goes green on its own — trimming it here would only collide with that trim.

Tests re-run locally on the touched areas: 153 passed (`test_token_refresh.py`, `test_token_refresh_beat.py`, `test_token_expiry_notify.py`, `test_api_coverage_misc.py`).

Still needs the `risk:live-linkedin` validation called out in the PR body: the refresh POST to `oauth/v2/accessToken` can only be proven against a real account holding a real refresh token.

gitchrisqueen:

- **Throttle slot burned on a send that never happened** (`notifications.py:notify_linkedin_token_expiring`): the Redis key was claimed with `SET NX EX=7d` *before* the email was attempted and never released. A user with no email on file, an SMTP failure, or any exception in the `try` block therefore got **no notice for a full week** — precisely the "expiry warning with no email behind it" failure #600 exists to fix, and the opposite of the sibling `notify_linkedin_session`, which only records its marker after a successful send. Fixed: the claim still happens up front (two workers must not both email the same user), but a send that did not happen releases the key so the next daily pass retries. Tests added for released-on-failure, released-on-exception, released-on-no-email, kept-on-success, and a delete that itself throws.
- **Raw ISO-8601 timestamp in a customer email** (`run_scheduler.py` → `email.py`): the beat passes `status["token_expiry_date"]`, which `resolve_token_status` produces as `expiry.isoformat()`, so the sent email read `It expires in 4 days (on 2026-09-14T08:30:12+00:00)`. The existing email test masked this by passing a pre-formatted `"Sep 1, 2026"`. Fixed in `send_linkedin_token_expiring_email` via a `_human_date()` normalizer (ISO → `Sep 14, 2026`, anything unparseable passed through), so every caller is covered rather than just this one. Tests added for both branches.

Also checked, not blocking: the `Closes #600` claim is honest (the issue's acceptance is behaviour + tests + CI, all met, no later phase in the body or comments); `get_linkedin_token_user_ids` correctly widens `get_active_user_ids` only where it must, though note it also drops the `last_login_inactivate_delay` clause, so a paying-but-dormant subscriber whose token has lapsed will receive the reconnect notice every 7 days indefinitely — consistent with how `notify_linkedin_session` already nags active subscribers, so left as-is; the `resolve_token_status` seam, `days_until_expiry` flooring/`None`-vs-`0`, `refresh_token_usable`, the DEBUG level on the expected "no refresh token" no-op, and the SPA countdown/`showHealthyCountdown` gating all hold up.

Full `tests/unit` suite green after the fixes: **8869 passed**.

gitchrisqueen:

PR #856, opened 2026-08-01
feat(groups): make group posting a visible, per-group choice (closes #769)

### LINKED ISSUE #769: chore(content-generation): Clarify group engagement settings for posts
## Why
The Who I engage setting in the LinkedIn Groups card states that LEM engages in joined groups by default, but it is unclear how posts are shared with these groups. Users cannot see which groups a post will go to or select/deselect groups for their posts. More configurations are needed to clarify if posts will be duplicated, reshared, or posted in the groups.

Auto-filed from in-app feedback #7 by the feedback→auto-work loop (`docs/launch-and-marketing-plan.md` §B.2/B.3). Classifier: cleanup/medium, risk `none`, confidence 0.80.

## Scope
- Address the reported cleanup in the `content-generation` area of the product.

## Files
- `src/cqc_lem/app/run_content_plan.py`
- `src/cqc_lem/utilities/ai/ai_helper.py`
- `src/cqc_lem/utilities/ai/content_framework.py`
- `tests/unit/`

## Acceptance
- All required CI gates pass.

gitchrisqueen:
## What & why

The LinkedIn Groups card told users LEM "engages in your joined groups" behind a single per-group
switch. That one switch covered two very different actions, and the one the reporter cared about was
invisible:

- **Are my posts duplicated / reshared / cross-posted into groups?** Nowhere in the product did it say.
  (They never were — a group post is generated fresh by `generate_group_post`.)
- **Which group does a group post go to?** `auto_group_posts` took `get_enabled_group_ids(uid)[0]` —
  whichever enabled row MySQL happened to return first. Unstated, unselectable, and it never rotated,
  so the same group got every weekly post forever.

### Changes

| | |
|---|---|
| **Migration** `V20260801023643__add_user_group_post_settings.sql` | Adds `user_groups.post_enabled` + `last_posted_at`. **Additive**, plus one backfill `UPDATE user_groups SET post_enabled = enabled` — the two flags are independent from here on, so an existing row that was already switched off must not start receiving posts because a new column defaulted to on. |
| **`db.get_next_group_for_post`** | The ONE place "which group is next" is decided: least-recently-posted post-enabled group, never-posted-there sorts first. |
| **`db.record_group_post`** | Stamps `last_posted_at`. Called only after the post actually shipped, so a failed run leaves that group next in line rather than silently skipping its turn. |
| **`db.set_groups_enabled`** | Accepts `{gid: {"enabled", "post_enabled"}}` **and** the old `{gid: bool}` (a stale SPA bundle still sends bools). Only keys present are written, so a partial payload never resets the other flag. |
| **`GET /user/groups`** | Marks `is_next_post` on the row. Deliberately not a new wrapper object — `detail` stays a plain array so an older cached bundle still renders it (issue #743). |
| **`auto_group_posts` / `auto_post_to_group`** | Dispatches the rotated group and passes `group_name` through, so the post is actually written for that group's members. |
| **`GroupsCard.tsx`** | Two switches per group (**Comment** / **Post**), a "Next group post: <name>" line + row badge, an explicit "none — LEM will not post in any group" state, and copy that says plainly: one original post, one group, about weekly, rotating; scheduled feed posts are never duplicated, reshared or cross-posted. |

Weekly volume is unchanged — still ONE group post per user per week. Commenting behaviour is untouched
(`auto_comment_in_groups` still reads `enabled`).

## Testing

- `tests/unit/app/test_groups.py` — rotation query shape (post_enabled alone gates posting; never-posted
  first), `record_group_post` stamping, bare-bool vs dict payloads, empty-state no-op write, dispatcher
  targets the rotated group, and a failed post leaves the group next in line.
- `tests/unit/api/test_api_coverage_misc.py` — `is_next_post` marking (and the none-opted-in case), the
  payload stays a plain array, `PUT` accepts both body shapes.
- `tests/unit/utilities/test_db_coverage_{avatar_groups,errors}.py` — flag/date coercion + error paths.
- `src/cqc_lem/ui/src/pages/account/GroupsCard.test.tsx` (new) — the duplicate/reshare copy, the
  next-group line, the badge dropping when that group's Post is switched off before saving, the
  nothing-opted-in state, separate save payloads, and a payload with no `post_enabled` reading as on.
- `poetry run pytest tests/unit -q` → **8846 passed** locally. Vitest runs in CI (`ui-build.yml`);
  this box is on Node 18 so it couldn't run here.

Closes #769

### DIFF
--- a/compose/local/database/migrations/V20260801023643__add_user_group_post_settings.sql
+++ b/compose/local/database/migrations/V20260801023643__add_user_group_post_settings.sql
@@ -1,0 +1,25 @@
+-- Per-group POSTING control, separate from per-group engagement (issue #769).
+--
+-- V41 gave `user_groups.enabled` a single meaning for two very different actions: value-add
+-- COMMENTS on other people's group posts (daily) and LEM publishing its OWN post into a group
+-- (weekly). A user who is happy to be commented-for in twelve groups has no way to say "but only
+-- post in these two", and the weekly post silently went to whichever enabled row the database
+-- happened to return first — invisible in the SPA and never rotating.
+--
+-- `post_enabled` splits the two so the Groups card can show (and control) them separately, and
+-- `last_posted_at` makes "which group is next" a stated, least-recently-posted rotation instead of
+-- an arbitrary row order. NULL last_posted_at = never posted there, which sorts FIRST so a newly
+-- joined group gets its turn before a group that already had one.
+--
+-- The two flags are independent from here on (commenting reads `enabled`, posting reads
+-- `post_enabled`), so existing rows are BACKFILLED from `enabled` rather than taking the column
+-- default: a group the user had already switched off must not start receiving posts because a new
+-- column defaulted to on. New rows keep DEFAULT 1, matching `enabled`'s opted-in-by-default posture.
+--
+-- Weekly volume is unchanged — still ONE group post per user per week; it just rotates now, and
+-- which group is next is visible in the SPA.
+ALTER TABLE user_groups
+    ADD COLUMN post_enabled   TINYINT(1) NOT NULL DEFAULT 1 AFTER enabled,
+    ADD COLUMN last_posted_at DATETIME NULL DEFAULT NULL AFTER last_synced_at;
+
+UPDATE user_groups SET post_enabled = enabled;
--- a/docs/SETTINGS_IA_RESEARCH.md
+++ b/docs/SETTINGS_IA_RESEARCH.md
@@ -115,7 +115,11 @@ Read-only fields the GET adds for the UI: `reply_inbound_address`, `gmail_forwar
   collaboration, profile viewer, nurture, plus six catch-up milestones. Each has step 0 + N
   follow-ups with `delay_hours`. Blank = built-in default.
 - **Lead magnet** — enabled, trigger keyword, DM message.
-- **Groups** — per joined group on/off; all on by default.
+- **Groups** — per joined group, TWO independent switches (issue #769): *Comment* (daily value-add
+  comments on other members' posts, out of the daily comment cap) and *Post* (the weekly ORIGINAL
+  group post — never a copy or reshare of a scheduled feed post, one group per week, rotating to
+  whichever post-enabled group has gone longest without one). Both on by default; the card names the
+  group the next post lands in.
 - **Billing** — plan/tier/trial, Stripe portal, video-credit and avatar-credit packs.
 
 ---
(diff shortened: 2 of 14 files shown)


gitchrisqueen:
🔎 Claude adversarial review — FIXED 2 findings (+1 filed as follow-up)

Reviewed the full diff plus the seams it touches (`upsert_user_group` / `get_enabled_group_ids` / the group beats / V41's schema / the SPA save-section wiring), then fixed what would have shipped a wrong answer to the exact question this issue is about — *which group does my post go to?*

- **The card kept naming the pre-save group after a successful save** (`GroupsCard.tsx`): `onSuccess` invalidated `['user-groups']`, but the `groupsInit` guard blocks the effect from ever re-initialising, so `is_next_post` in local state stayed at its mount value. Save a change that moves the rotation and the badge + "Next group post: X" line confidently named the old group until a full page reload. Fixed: the save now `refetch()`es and adopts the server's re-resolved marking (falling back to local state if the refetch fails).
- **Switching Post ON for a never-posted group left the old claim standing** (`GroupsCard.tsx`): a NULL `last_posted_at` sorts FIRST, so that group *is* the new next one — but the card only dropped its claim when the marked group itself was switched off. It now stands down to "picked from the groups below when you save" for any unsaved change to which groups take posts, instead of naming the wrong group.
- Tests: two new cases in `GroupsCard.test.tsx` (claim drops when another group is switched on; the server's re-resolved rotation is adopted after save). Vitest runs in CI — this box is Node 18. Python untouched; re-ran `tests/unit/app/test_groups.py`, `tests/unit/api/test_api_coverage_misc.py`, `tests/unit/utilities/test_db_coverage_{avatar_groups,errors}.py` → 267 passed.

Filed rather than fixed here (pre-existing, needs its own migration): **#858** — a group LEM *cannot* post in (admin-only groups render no share box) never gets `last_posted_at` stamped, so it stays "next" every week forever and starves every other post-enabled group. Correct call by this PR to stamp only on success for transient failures; the permanent-failure case needs a separate run-timestamp column.

Checked and clean: `Closes #769` is honest (every acceptance criterion is in the diff, no later phase declared in the issue); migration is additive with the `post_enabled = enabled` backfill that stops a switched-off group from starting to receive posts; the dynamic `SET` clause in `set_groups_enabled` is built only from a fixed column tuple and every write is scoped by `user_id`; the bare-bool payload path really does leave `post_enabled` alone; `enabled` still gates commenting only; `generate_group_post` already accepted `group_name`; `GET /user/groups` `detail` stays a plain array for stale bundles (#743).

gitchrisqueen:

PR #859, opened 2026-08-01
fix(analytics): explain and widen the subset of posts the dashboard measures (closes #809)

### LINKED ISSUE #809: fix(analytics): Analytics dashboard shows only subset of posts
## Why
Engagement Analytics displays "x of 11 posts" while the account total shows 30 posts. The rest of the analytics are missing, and previously it showed a different subset (e.g., x of 12 posts).

Auto-filed from in-app feedback #21 by the feedback→auto-work loop (`docs/launch-and-marketing-plan.md` §B.2/B.3). Classifier: bug/medium, risk `none`, confidence 0.94.

## Scope
- Address the reported bug in the `analytics` area of the product.

## Files
- `tests/unit/`

## Acceptance
- All required CI gates pass.

gitchrisqueen:
## Why

Reported via in-app feedback: Engagement Analytics showed **"x of 11 posts"** while the account tile said **30 posts**, "the rest of the analytics are missing", and the number had previously been 12 — so it looked like the dashboard was losing posts.

It was two separate problems.

**1. Three unrelated denominators shared one screen, unlabelled.**

| Number | What it actually counts |
|---|---|
| 30 (`/dashboard/stats/`) | posted, **all time** |
| 11 (content mix) | non-rejected posts scheduled in the last 90 days, **classified only** — drafts and scheduled posts included |
| "Posts measured" | posted posts in the last 90 days **that have a captured `post_stats` row** |

Nothing on the page said which was which, and the "12 → 11" drift was just posts ageing out of the rolling window.

**2. The measured subset was genuinely too small — permanently.**

`auto_scrape_post_stats` walks `get_recent_posted_post_ids`' **21-day** window, but the dashboard reads **90 days**. A post whose capture was missed while it was fresh (automation paused, no permalink logged yet, a 429) could never be measured afterwards — it aged out of the capture window and stayed invisible forever.

## What changed

**Coverage is reported.** `GET /user/engagement-analytics` returns a `coverage` block — `posted_total`, `posted_in_window`, `measured`, `awaiting_capture` — backed by one new read, `db.get_post_coverage_counts`. `measured` is taken from the rows the endpoint actually read rather than re-counted in SQL, so the coverage line can never contradict `sample_size`.

**The SPA says it out loud.** `Dashboard.tsx` renders one line under the header ("Measuring 1 of 11 post(s) published in the last 90 days (30 posted all time). 10 still need their stats read from LinkedIn…"). The empty state names those same numbers instead of a bare `(currently 0)`, and the content-mix footnote now states its denominator (planned posts, drafts and scheduled included) so it stops reading as a count of published posts. The coverage sentence is built as a string, not inline JSX — a newline between text and an expression container becomes a space, which would have rendered `post s` and ` .` mid-sentence.

**Older posts become measurable.** Each sweep tops itself up with up to `POST_STATS_BACKFILL_MAX` (default 5) posts from the last `POST_STATS_BACKFILL_DAYS` (default 90) that have **no** `post_stats` row at all (`db.get_uncaptured_posted_post_ids`), deduped against the recent set. The cap keeps the extra page loads predictable inside the same Chrome session; `POST_STATS_BACKFILL_MAX=0` disables it. A typo'd env value falls back to the defaults rather than taking the sweep down. Both documented in `.env.example`.

## Testing

- `tests/unit/utilities/test_analytics_coverage.py` (new) — both db reads: SQL shape, that only the second count is windowed, the `LEFT JOIN … IS NULL` selection, limit clamping, missing-row and `mysql.connector.Error` fallbacks.
- `tests/unit/app/test_sweep_fixes.py` — `TestScrapeBackfillsNeverCapturedPosts`: older uncaptured posts are appended, a post already in the recent sweep is not scraped twice, a backfill-only account still opens a session, window/cap are env-tunable, a cap of 0 disables it, unparseable env falls back.
- `tests/unit/api/test_api_coverage_misc.py` — coverage reconciles with the account total and matches `sample_size`; `awaiting_capture` never goes negative.
- Full suite green locally: `8878 passed`. UI `tsc -b` and `eslint` clean.

No migration, no schema change, no new external call.

Closes #809

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -195,6 +195,14 @@ SUPPRESSION_PAUSE_SECONDS=7776000
 # The comment-demotion half of the verdict reads a rolling week, matching #628's own scoring window
 # — a month-old episode that has since been remediated must not trip a 90-day pause today.
 SUPPRESSION_COMMENT_DAYS=7
+# Post-stats backfill (issue #809). The stats sweep walks the last three weeks, but the analytics
+# dashboard reads 90 days — a post whose capture was missed while it was fresh (automation paused, no
+# permalink logged yet, a 429) stayed unmeasurable forever, so the dashboard rendered a shrinking
+# SUBSET of the account's posts. Each sweep now also picks up POST_STATS_BACKFILL_MAX posts from the
+# last POST_STATS_BACKFILL_DAYS that have no stat row at all. Keep the cap small: every extra id is
+# two more page loads inside the same Chrome session. Set the cap to 0 to disable the backfill.
+POST_STATS_BACKFILL_DAYS=90
+POST_STATS_BACKFILL_MAX=5
 # Content-quality telemetry (issue #630). Every quality gate LEM has is a one-time verdict; this is
 # the TREND. A nightly pass scores everything shipped in CONTENT_QUALITY_WINDOW_DAYS (posts, feed
 # comments, newsletter editions) — the D1 slop score, self-similarity vs that surface's own recent
--- a/src/cqc_lem/api/main.py
+++ b/src/cqc_lem/api/main.py
@@ -60,7 +60,7 @@ from cqc_lem.utilities.db import (
     get_pending_newsletter_editions,
     get_latest_edition_scheduled_for, update_newsletter_edition, get_newsletter_edition,
     get_user_groups, set_groups_enabled, get_next_group_for_post,
-    get_post_engagement_rows, get_post_performance_rows,
+    get_post_engagement_rows, get_post_performance_rows, get_post_coverage_counts,
     get_content_mix_counts, get_comment_outcomes,
     get_follower_stats, get_daily_action_counts,
     get_lead_magnet_settings, update_lead_magnet_settings,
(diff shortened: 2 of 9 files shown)


gitchrisqueen:

- **Backfilled posts recorded fabricated zeros on an unreadable page** (`run_automation.auto_scrape_post_stats`): `_post_social_counts` returns an all-zero dict when the card can't be read and `container is None` yields `{}`, so a post whose page never rendered (auth wall, 429, dead permalink) was written as a real `record_post_stats(0, 0, …)` row. For a post inside the 21-day window a later sweep overwrites it; a **backfilled** post leaves `get_uncaptured_posted_post_ids` the moment ANY row exists, so a single bad read published a permanent zero-engagement row into the exact panel this PR is fixing — and into the `build_engagement_trend` series `auto_suppression_tripwire` (#629) reads. One 429 mid-sweep would have retired 5 old posts as "measured, 0 engagement". Fixed: skip `record_post_stats` when the read produced no signals at all (`log_debug`, expected no-op per the escalation rule). A readable page always yields the full zero-filled dict, so a genuinely zero-engagement post is still measured — covered by a test each way.
- **Backfill could starve itself on permalink-less posts** (`db.get_uncaptured_posted_post_ids`): the query offered posts the sweep can only `continue` past (`get_post_url_from_log_for_user` → None — one of the causes the PR body itself names). A post leaves that set only by GAINING a stat row, so a handful of permalink-less posts at the head of the 90-day window would occupy every slot of `POST_STATS_BACKFILL_MAX` on every run, forever, and no capturable post behind them would ever be reached — the backfill would report as working while the dashboard stayed at "1 of 11". Fixed: the query now requires a logged `post` / `success` row with a non-empty `post_url`.

Also checked, no finding: coverage windows on the same `p.scheduled_time` + `status='posted'` as `get_post_performance_rows`, so `measured`/`posted_in_window` genuinely reconcile; `measured` reads the rows the endpoint actually returned; `awaiting_capture` floors at 0; SQL fully parameterized; `PostStatus`/`LogActionType`/`LogResultType` enums used; env parsing falls back rather than taking the sweep down; `Closes #809` is honest (acceptance criteria are the generic auto-filed set, no later phase declared).

Tests: full unit suite green locally — `8881 passed`.

gitchrisqueen:

PR #861, opened 2026-08-01
fix(engagement): re-ground the feed sort control and record the sort each scan ran against (closes #817)

### LINKED ISSUE #817: fix(engagement): 'Selector miss: Feed sort control' — recency scoring silently runs against the algorithmic feed
## Why

`Selector miss: Feed sort control` — **15 occurrences across 3 days** (PostHog Logs, 48h window).

`_switch_feed_to_recent()` (`src/cqc_lem/app/run_automation.py:1085`) is documented as a best-effort "silent no-op" on failure. That was a reasonable call when it was written, but it is no longer harmless: issue #622 made the feed scorer **recency-dominant** (`_score_feed_post` = recency + relevance + reciprocity + activity, with `_post_age_minutes`). If the "Sort by → Recent" control is never found, the whole scoring matrix runs against LinkedIn's default algorithmic feed — so the recency term scores posts that the feed already reordered by engagement, and the #622 behaviour quietly degrades to roughly what it replaced.

It is called twice per run (`navigate_to_feed` + `comment_on_feed_inline`, per the comment at `:1092-1093`), and each miss burns `MAX_WAIT_RETRY` × ~5s first.

## Scope

- Re-ground the sort-control locator against the live DOM with an ordered fallback chain.
- Decide and implement what a miss should MEAN now: at minimum it must be visible (it will be, via the new recurrence escalation), and the run should record that scoring ran unsorted so the golden-hour/feed telemetry isn't read as if recency sorting was active.

## Files

- `src/cqc_lem/app/run_automation.py` (`_switch_feed_to_recent`, `_score_feed_post`)

## Acceptance criteria

- [ ] Sort control re-grounded with a fallback chain, validated live
- [ ] A miss is recorded in the run's telemetry, not just logged
- [ ] `Selector miss: Feed sort control` drops to ~0 after deploy

gitchrisqueen:
## What & why

`Selector miss: Feed sort control` fired **15 times across 3 days** and meant nothing to the run that hit it: `_switch_feed_to_recent` was documented as a best-effort *silent* no-op. That was fine when it was written. Since **#622** made `_score_feed_post` recency-**dominant**, it isn't: with no sort control the whole scoring matrix ranks a candidate pool LinkedIn has already reordered by engagement, so the recency term scores an algorithmic feed and #622 quietly degrades to roughly what it replaced — with nothing in the telemetry saying so.

### 1. Re-grounded locator chain
Ordered, most-stable anchor first, mirroring the #628 comment-sort idiom in the same file: `aria-label` → `data-testid` → visible `'Sort by'` text → a popup trigger whose whole label **is** the current sort (the `Sort by` prefix is dropped on narrow layouts) → `role=button`. Same treatment for the `Recent` menu option (`button` / `li` / `menuitem` / `menuitemradio` / `option`). Every comparison is case-folded through the existing `translate()` helper (LinkedIn renders `Recent`, and a case-sensitive literal silently never fires); no class names are keyed on.

`max_try=1` — the old miss burned `MAX_WAIT_RETRY` × ~5s **twice per run** to report the same thing. The `WebDriverWait` poll still gives the chain its full grounding window.

### 2. A miss now MEANS something
`_switch_feed_to_recent` returns the run's sort state instead of `None`:

| state | meaning |
|---|---|
| `recent` | the control **confirmed** Recent after the flip — the only value that means sorted |
| `top` | trigger resolved, the `Recent` option did not |
| `missing` | no trigger resolved at all |
| `unknown` | flipped, but the resulting sort was unreadable |
| `n/a` | a surface that never had the control (group feed) |

`recent` is only returned when the control reports it **afterwards** — an unverified flip recorded as sorted tells the same lie the silent no-op told. `top` vs `missing` matters because they have different fixes, and the log line alone cannot tell them apart.

**Group feeds:** `comment_on_feed_inline` is also the group-commenting engine (`auto_comment_in_groups`), and a group page never had a home-feed sort control — so a miss there was a warning filed against working behaviour, and part of the 15. It is now `n/a` at DEBUG, per `utilities/CLAUDE.md`.

### 3. Recorded in telemetry, not just logged
- `feed_sort` on the feed funnel (Redis, already surfaced on `GET /user/settings`)
- a new **`feed_scan`** PostHog event via `track_feed_scan` — emitted for **every** scan including the ones that comment on nothing, carrying the funnel + `feed_sort` as a **string** (boolean props match nothing in alert tiles)
- `sort <state>` on the `Engagement scan` log line
- a caveat line on the Account → targeting reach funnel, so a user reading "examined 40 posts" isn't reading it as 40 fresh ones

The `Selector miss: Feed sort control` warning itself is deliberately **kept** — the issue asks for the miss to stay visible, and #648's recurrence escalation is what surfaced it in the first place.

### 4. Live grounding
`scripts/linkedin_live_validation.py --feed-sort` reports whether the trigger resolves, what it reads before/after the flip, and **every visible control label on the screen** — captured after the flip attempt, so when the trigger resolved and the option didn't, the still-open dropdown is in the capture. A bare "not found" is not re-groundable; live labels are what the next chain gets written from.

```
sudo docker exec -i celery_worker_selenium python - --user-id 1 --feed-sort \
  < scripts/linkedin_live_validation.py
```

## Testing

- `tests/unit/app/test_feed_sort.py` — 23 tests: each sort state, the unverified flip never reading as sorted, aria-label-only labels, the group-feed no-op emitting no warning, fail-fast, `_is_home_feed` parametrized, locator-chain invariants (case-folded, never class-keyed), and the funnel/`feed_scan` recording a miss.
- `tests/unit/utilities/test_observability.py` — `track_feed_scan` emits for every scan, `feed_sort` reaches the event as a string, absent = `None` (never `recent`).
- `tests/unit/test_linkedin_live_validation.py` — the probe's verdict points at the right half of the control; `menu_item_labels` skips hidden/duplicate rows; `--feed-sort` alone is a valid invocation.
- `TargetingSection.test.tsx` — the caveat shows only when the scan genuinely wasn't sorted (silent for `recent`, `n/a`, and funnels written before this shipped).
- Full local suite: `8907 passed`.

## Scope note

Acceptance criterion 1 says "validated live" and criterion 3 is a post-deploy observation — neither is doable headless, which is why this is `risk:live-linkedin` and parked for you. The probe above is the mechanism for the first; `feed_scan` is the mechanism for the third. See the Decision Comment.

`_score_feed_post` is deliberately unchanged: per-card age comes from `_post_age_minutes`, so the scoring itself stays correct when unsorted — what degrades is the candidate POOL. Whether an unsorted run should still comment is question 2 below.

**Remaining on #817 — this PR deliberately does NOT close it.** Two of the three acceptance
criteria cannot be satisfied from here:

1. *"Sort control re-grounded with a fallback chain, **validated live**"* — the chain is still
   ungrounded. You chose `1A` (run the probe, then merge) and the `--feed-sort` JSON has never been
   posted on either thread, so merging with a closing keyword would file criterion 1 as met on a
   chain no live feed has ever seen.
3. *"`Selector miss: Feed sort control` drops to ~0 after deploy"* — a post-deploy observation by
   construction. `feed_scan.feed_sort` is the mechanism; the reading takes a day of data.

Criterion 2 (a miss is recorded in telemetry, not just logged) IS met here. Merging this ships the
instrumentation that grounds the other two — the issue stays open until a live reading confirms
`feed_sort=recent`, which is option `1C` from the Decision Comment. Say the word and I'll re-ground
the chain from your probe JSON on a follow-up PR.


gitchrisqueen:
Held (`needs-human`, risk: live-linkedin). The new sort-control locator chain was written from the SDUI idioms already grounded in this repo, but it has **not** been run against a live feed — and acceptance criterion 1 on #817 says "validated live". There's also one product call this PR deliberately did not make on its own.

### 1. Ground the chain live before merging, or merge and let telemetry ground it?
- **A. Run the probe first, merge after** — one read-only command in the Selenium worker; if the chain misses, paste `visible_controls` from the JSON here and I re-ground it in this PR before it ever ships. ✅ *recommended*
  ```
  sudo docker exec -i celery_worker_selenium python - --user-id 1 --feed-sort \
    < scripts/linkedin_live_validation.py
  ```
- **B. Merge now, ground from the `feed_scan` event** — the first day of `feed_sort` values answers the same question, one release later; if it reads `missing`, that's another PR.
- **C. Merge now and drop `Closes #817`** — leave the issue open until a live reading confirms `feed_sort=recent`, so the record can't close on an unvalidated chain.

### 2. What should a run DO when it discovers the feed is unsorted?
- **A. Comment anyway, record the state** — what this PR implements. Per-card age still comes from `_post_age_minutes`, so scoring stays correct; only the candidate pool is worse, and a partially-fresh pool beats commenting on nothing. ✅ *recommended*
- **B. Skip the feed pass entirely for that run** — never comment against an algorithmic pool. Cleanest signal, but on a bad selector day it silently zeroes feed commenting for every user — the exact failure shape of the #372 doom loop.
- **C. Comment anyway, but tighten the recency gate** — halve `max_post_age_hours` when unsorted, so an algorithmic pool can't feed us day-old posts. Safer than B, but it changes a user's saved preference behind their back.

**My recommendation: `1A 2A`.** 1A costs one read-only command and is the only thing that can tell `missing` (selectors rotated) from `top` (the menu rotated) *before* this ships; 2A keeps the fix observational — #817 asked for the miss to be visible and recorded, and turning it into a hard stop is a bigger behaviour change than the evidence supports yet.

gitchrisqueen:
Ok

gitchrisqueen:
Applied your `Ok` — **1A + 2A**. One of them needed a code change, and it wasn't the one you'd expect.

### 2A — comment anyway, record the state
No change: that is what the PR already does. Recording it as your explicit call rather than my default — an unsorted run still comments, per-card age still comes from `_post_age_minutes`, and only the candidate POOL is degraded. The state rides onto the funnel, `feed_scan` and the Account caveat so the run is never *read* as recency-sorted when it wasn't.

### 1A — run the probe first… which it could not have done
**The command in my own Decision Comment would have failed.** `probe_feed_sort` imported `_FEED_SORT_LOCATORS` / `_feed_sort_state` / `_switch_feed_to_recent` from `cqc_lem.app.run_automation` — but the probe is piped into `celery_worker_selenium`, whose `cqc_lem` is the code baked into the **deployed image**. That image is built off `main`, and `main` has none of those three symbols (`_switch_feed_to_recent` exists there, but returns `None`). So the pass would have died on `ImportError` before opening the feed.

That is backwards for the only grounding pass that matters: a **pre-merge** run cannot import the chain it is grounding, because merging is what ships it. Fixed in 6bf35fe:

- the probe **carries the chain** and takes the flip steps itself, preferring the running image's chain when it has one;
- the report says which it read — `"chain_source": "image"` vs `"script"`, and the verdict appends *"chain carried by this script — the running image predates #817"*, so a pre-merge reading can never be filed as a post-deploy one;
- `TestFeedSortChainCopy` asserts the carried copy is **identical** to `_FEED_SORT_LOCATORS` / `_FEED_RECENT_OPTION_LOCATORS` (and the state strings to `FEED_SORT_*`). A copy that drifts would ground a chain nothing ships — worse than not probing. Verified it fails on a one-line divergence.
- `option_found` is now reported separately. The production return has to collapse "the menu never opened" and "it opened without a Recent row" into `top`; those get re-ground from different halves of the capture.
- a session failure mid-flip reports `flip_error` instead of reading as a selector miss and sending the next pass after the wrong thing.

**The command is unchanged and now works from this branch's checkout:**

```
sudo docker exec -i celery_worker_selenium python - --user-id 1 --feed-sort \
  < scripts/linkedin_live_validation.py
```

Read-only: it navigates the feed and flips the sort control, exactly as the production reader does. Paste the JSON here — `verdict` + `visible_controls` are the whole payload — and I re-ground the chain and hand it to merge.

Tests: `8915 passed` locally (43 in `test_linkedin_live_validation.py`, +5 new probe cases and the drift lock).

---

## 🧑‍⚖️ Still parked — only because 1A is yours to run
I can't run it: `docker` and live LinkedIn sessions are both outside what I'm allowed to do headless. **Nothing more is needed from you than pasting that JSON** — this block only exists in case you'd rather not run it after all.

### 1. Still ground it live before merge?
- **A. Yes — I'll run the probe, paste the JSON** — you already chose this; I re-ground if it misses, then merge.  ✅ *recommended*
- **B. Merge now, ground from `feed_scan` instead** — the first day of `feed_sort` values says the same thing one release later; a `missing` reading becomes another PR.
- **C. Merge now and drop `Closes #817`** — issue stays open until a live reading confirms `feed_sort=recent`.

**My recommendation: `1A`** — unchanged, and now actually executable. It's the only thing that separates `missing` (the trigger's selectors rotated) from `top` (the menu did) *before* this ships, and the capture is what the next chain gets written from.

Reply with the probe JSON, or `1B` / `1C`.

gitchrisqueen:
Rebased onto `main` and resolved. Now `MERGEABLE`, CI re-running.

The conflict was **semantic, not mechanical** — worth a look before merge:

`main` gained **#872** (`_feed_sort_control_expected` + `warn_on_miss` / `max_try` plumbing) to stop group-feed misses warning. This PR supersedes that helper with `_is_home_feed` + the ordered locator chain. Both solve the same concern by different mechanisms, so I kept #872's three guarantees under this PR's mechanism rather than reverting either side:

| #872 guarantee | How #817 keeps it |
|---|---|
| group feed must not warn | surface rejected **before** the lookup — no miss to warn about, and skips the lookup's wall clock (stronger than `warn_on_miss=False`) |
| home feed must still warn | lookup runs; `find_first`'s `warn_on_miss` defaults `True` |
| unreadable URL must not warn | **was a real regression** — `_is_home_feed` raised out into the outer handler, which logs `Feed recent-sort failed` at WARNING, filing the same defect by the other door #872 existed to prevent. Now returns `False`. Caught by #872's own `test_unreadable_url_does_not_warn`. |

Also kept **`_X_SHORT_TEXT`** from `main`, which a naive "take theirs" would have dropped: this PR hoists the shared case-fold constants above the comment-sort block, but `_X_SHORT_TEXT` is used by two comment-sort locators — losing it is a `NameError` at import.

The three #872 tests now assert the superseding mechanism instead of the old kwargs.

**Verification:** full unit suite green locally — **9067 passed**.

Handing back for merge — flagging the mechanism swap explicitly since it changes how #872's guard is implemented, not just where.

gitchrisqueen:

Reviewed as the gate (no Copilot on this one). Fixes pushed as `9951cd1`; full unit suite green locally (**9071 passed**).

- **`_feed_sort_state` resolved an ambiguous label to `recent`, skipping the flip entirely** (`run_automation.py`): it returned the first token it matched, so a trigger that spells its options into the accessible name — `"Sort by, currently Top, options Top and Recent"` — read as already-sorted. That is both halves of the lie this PR exists to stop in one label: the early return never clicks the control, and the run is recorded `feed_sort=recent` while the feed is still on Top. A label naming BOTH sorts is now unreadable (``), so the flip runs and `recent` can only come from a control that says so unambiguously. Mirrored in the probe’s `control_sort_state` so a live pass cannot report a control healthy on the exact reading production reads as `unknown`.
- **The Account caveat asserted something untrue for 2 of the 3 states it fires on** (`TargetingSection.tsx`): it said LinkedIn’s control *"wasn’t available on this scan"* and that the posts *"came from your algorithmic feed"*. Only true of `missing` — `top` means the control WAS there and wouldn’t flip, and `unknown` means the flip may well have worked and just couldn’t be read back. Reworded to what holds for all three ("couldn’t confirm … may have come from").
- **The probe could miss the menu it exists to capture** (`linkedin_live_validation.py`): `menu_item_labels` queried the menu roles and `li` in ONE comma-joined selector, and `find_elements` returns **document order** — a feed page’s global nav, left rail and post action rows are all `li` and come before an overlay dropdown, so the 40-label cap could be spent on page furniture. Roles are now enumerated first, `li` only fills the remainder. This is the capture your `1A` pass exists to produce.
- **`Closes #817` was not honest** (PR body): criterion 1 says *validated live* and you chose `1A` (probe first, merge after) — the `--feed-sort` JSON has never been posted, and criterion 3 is post-deploy by construction. The last tick handed this to merge anyway, which would have closed the issue on a chain no live feed has ever seen. Dropped the closing keyword, replaced it with `Remaining on #817: …` (option `1C` from your own Decision Comment), and left the detail + both ways to close it on the issue: https://github.com/christopherqueenconsulting/linkedin_engagement_manager/issues/817#issuecomment-5153731558. Criterion 2 is met and merging ships the `feed_scan` telemetry that makes the other two answerable.

Checked and found sound: the `_is_home_feed` / #872 reconciliation (group + unreadable-URL misses stay DEBUG, home-feed rot still warns, `_X_SHORT_TEXT` preserved), every locator in both chains parses as valid XPath 1.0 and is case-folded with no class/id keying, `max_try=1` against the 15s `WAIT_DEFAULT_TIMEOUT`, `track_feed_scan` matching the `track_catchup_run` capture idiom (string `feed_sort`, `distinct_id`, no boolean props), and the funnel reaching the SPA unmapped through `GET /user/settings`.

+4 tests: both ambiguous-label paths (before and after the flip), the both-sorts probe reading, and menu rows surviving 60 list items.

gitchrisqueen:

PR #862, opened 2026-08-01
test(observability): make error-capture middleware probes outrank the SPA catch-all (closes #820)

### LINKED ISSUE #820: fix(observability): error-capture middleware tests fail on main — FastAPI 500 → $exception path is unverified
## Why

Two tests in `tests/unit/api/test_error_capture_middleware.py` fail on a clean `main` (verified by stashing all local changes):

```
FAILED TestUnhandledExceptionCapture::test_unhandled_route_error_is_captured_with_route_context
    assert 200 == 500   # /__probe_boom returned 200
FAILED TestUnhandledExceptionCapture::test_http_exception_is_not_captured
    assert 200 == 404   # /__probe_http_error returned 200
```

Both probe routes return **200** instead of raising, so the assertions about `capture_exception` never actually run. The middleware's unhandled-exception branch — the surface `docs/error-tracking.md` lists as the source of `$exception` for FastAPI routes — therefore has **no working test coverage**.

This matters more than a red test: a PostHog audit found **zero `$exception` events in 48h** and only 8 in 10 days. Server-side exception capture is either not firing or firing far less than expected, and this is precisely the test that would have caught it.

## Scope

- Find why `/__probe_boom` and `/__probe_http_error` return 200. Most likely the probe routes are registered after a catch-all (the SPA fallback route in `api/main.py`), so the catch-all wins and serves 200 before the probe handler is reached.
- Fix the fixture/route registration so the probes actually raise.
- Then confirm the assertions still hold — i.e. that `capture_exception` IS called with `route`/`method`/`source` on a 500, and is NOT called for an `HTTPException`.
- If the middleware itself turns out to be broken (not just the test), that is the real bug and the fix belongs there.

## Files

- `tests/unit/api/test_error_capture_middleware.py`
- `src/cqc_lem/api/main.py` (`observability_middleware`, route registration order)

## Acceptance criteria

- [x] Both tests pass for the right reason (probe routes genuinely 500 / 404)
- [x] A real unhandled route error is confirmed to emit `$exception` with route context
- [x] `HTTPException` still does not emit one

gitchrisqueen:
## What

`tests/unit/api/test_error_capture_middleware.py` registered its two probe routes on the real `app` **after** main.py's SPA fallback (`@app.get("/{full_path:path}")`, registered last, and only when `src/cqc_lem/ui/dist` exists). Starlette matches routes in registration order, so `/__probe_boom` and `/__probe_http_error` both resolved to the catch-all and returned **200 index.html**. Every assertion about the middleware then ran against the wrong response — the FastAPI `500 → $exception` surface that `docs/error-tracking.md` names had **no working coverage** anywhere the UI had been built (locally, and in any job that builds the SPA).

Reproduced exactly as the issue describes: with `ui/dist` present, `assert 200 == 500` and `assert 200 == 404`; with it absent, the file passed only because the catch-all wasn't registered at all.

## Fix

Test-only. The middleware turned out to be **correct** — see below.

- Probe routes are inserted at the **front** of `app.router.routes`, so they win against any catch-all.
- The fixture registers an equivalent SPA-style catch-all **unconditionally**, so the shadowing hazard is exercised whether or not `ui/dist` was built — the file can no longer pass for the wrong reason in one environment and fail in the other.
- The shared `app` singleton's route table is **restored on teardown** (it previously leaked the probe routes into the rest of the session).
- `track_api_call` is patched in `cqc_lem.api.main`'s own namespace rather than on `observability` — main.py binds it at import time, so the old patch was a no-op whenever another test module had already imported main.

## Confirming the middleware actually emits `$exception`

The existing tests patch `cqc_lem.api.main.capture_exception`, which proves the middleware *calls* something, not that an `$exception` is emitted. Two new end-to-end tests (`TestExceptionReachesPostHog`) run the **real** `capture_exception` with a stand-in `posthog` that reports itself enabled (the test env is keyless, so `posthog.disabled` would short-circuit it):

- a real unhandled `RuntimeError` in a route reaches `posthog.capture_exception` with `distinct_id="system"` and properties `route=/__probe_boom`, `method=GET`, `source=fastapi.middleware`;
- an `HTTPException` emits nothing.

So the FastAPI 500 path is healthy; the low `$exception` volume in the PostHog audit is not caused by this surface (the `log_warning` escalation gap that did suppress events was fixed separately in v0.113.0).

## Acceptance criteria

- [x] Both tests pass for the right reason (probes genuinely 500 / 404) — plus a guard test asserting an unmatched path still resolves to the catch-all with a 200, so a future regression in probe ordering fails loudly instead of silently passing
- [x] A real unhandled route error is confirmed to emit `$exception` with route context
- [x] `HTTPException` still does not emit one

## Testing

- `tests/unit/api/test_error_capture_middleware.py` — 6 passed, run **both** with and without a `src/cqc_lem/ui/dist` present
- `tests/unit` full suite — 8884 passed

Closes #820

### DIFF
--- a/tests/unit/api/test_error_capture_middleware.py
+++ b/tests/unit/api/test_error_capture_middleware.py
@@ -99,8 +99,11 @@ class TestUnhandledExceptionCapture:
 
     def test_healthy_request_is_not_captured(self, client):
         with patch("cqc_lem.api.main.capture_exception") as mock_capture:
-            client.get("/health")
+            response = client.get("/health")
 
+        # Assert the body, not just a 200: the catch-all also answers 200, so "no capture" on a
+        # shadowed /health would be the same vacuous pass the probes above were fixed out of.
+        assert response.json() == {"status": "healthy"}
         mock_capture.assert_not_called()
 
 
--- a/tests/unit/api/test_error_capture_middleware.py
+++ b/tests/unit/api/test_error_capture_middleware.py
@@ -1,7 +1,7 @@
 """Unit tests for the FastAPI unhandled-exception capture in the observability middleware
 (issue #648)."""
 
-from unittest.mock import patch
+from unittest.mock import MagicMock, patch
 
 import pytest
 from fastapi import HTTPException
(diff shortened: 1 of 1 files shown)


gitchrisqueen:

Verified the core claim empirically rather than taking it on faith: reproduced the original bug (probes 200 with `ui/dist` present), confirmed the fix passes **both** with and without a built `ui/dist`, and confirmed the fix actually *bites* — deleting the three route-reordering lines puts all four probe assertions back to red, so this can't silently regress to the shadowed state.

- **Vacuous assertion in `test_healthy_request_is_not_captured`**: it only asserted `capture_exception` was not called, which a catch-all-shadowed `/health` satisfies for the wrong reason — exactly the failure mode this PR exists to eliminate, and the fixture now registers a catch-all *unconditionally*, so the exposure is new in environments where `ui/dist` was absent. Fixed: it now asserts the response body is the real health payload, so the request must have reached the real handler.

Checked and found clean:
- **Middleware order** — `@app.middleware("http")` prepends, so `api_token_middleware` is outer and `observability_middleware` sits directly inside `ExceptionMiddleware`; an unhandled route error genuinely reaches its `except`, and `HTTPException` genuinely does not. The PR's "middleware is correct, the test was lying" conclusion holds.
- **Route-table mutation on the shared `app` singleton** — probes inserted at index 0 are all literal paths, so nothing existing is shadowed; teardown restores `original_routes`, and 749 `tests/unit/api` tests + the full 8884-test suite pass, so no leak into other modules. The only global that could outlive the module is the cached `app.openapi_schema` (probes default to `include_in_schema=True`), but the one test that fetches `/openapi.json` lives in a different module, and module-scoped teardown runs before/after it in either order.
- **`patch("cqc_lem.api.main.track_api_call")`** — correct namespace; main.py binds it at import, so the previous `observability`-level patch was a no-op once any other module had imported main first.
- **`TestExceptionReachesPostHog`** — `disabled = False` on the stand-in is required (the keyless test env would short-circuit `capture_exception`), and the assertions land on the real `posthog.capture_exception` kwargs.
- **Honest close** — all three acceptance criteria in #820 are covered by the diff, and the issue declares no later phase, so `Closes #820` is legitimate.

Tests: `tests/unit/api/test_error_capture_middleware.py` 6 passed (run with and without `src/cqc_lem/ui/dist`); `tests/unit` 8884 passed.

gitchrisqueen:

PR #814, opened 2026-07-31
feat(affiliate): LEM-generated promotional content with FTC disclosure (closes #770)

gitchrisqueen:
## What & why

#750 shipped (B)'s consent record, `apply_disclosure()`, `disclosure_report()` and the
`post_to_linkedin` gate — and deliberately no writer, so `promo_content_allowed()` had nothing
upstream of it. This is that writer.

`utilities/marketing/affiliate_content.py` is the ONE place promotional copy about LEM is produced.
Four refusals define it:

| | |
|---|---|
| **Not a second consent path** | `promo_content_allowed()` is still the only authorisation. `AFFILIATE_PROMO_CONTENT_ENABLED` can only turn the writer **off** — `True` with nobody consented writes nothing. An unreadable enrollment row reads as *no*. |
| **Not an extra post** | An affiliate post **CLAIMS** one of the 70/20/10 governor's promo slots (#618) instead of running beside it, so the 10% ceiling holds *by construction* rather than via a second counter that could drift. Only `content_mix == 'promo'`, and only 1 in `AFFILIATE_PROMO_EVERY_N_PROMO_SLOTS` of those (default 3 ⇒ **≤1 post in 30**). Unclaimed slots write the author's own case study exactly as before. |
| **Not an author of facts** | The writer's allow-list is `PRODUCT_FACTS` and nothing else — no result, metric, hours-saved or testimonial, because LEM cannot verify any claim about what the tool did for *this* user and an invented one is a false endorsement. The #625 slop lint regenerates then **blocks**. |
| **Not silently published** | `apply_disclosure(..., tagged=True)` stamps the disclosure at generation; the new `affiliate_promo` quality gate holds the post **PENDING** for the author's explicit approval; `disclosure_report()` in `post_to_linkedin` stays the backstop. Three different guarantees, not the same one three times. |

**Pacing** rides `human_pacing.ACTION_AFFILIATE_PROMO` (own daily budget, rest days inherited) —
deliberately outside `ENVELOPE_ACTIONS`, since posting is API-driven and was never gated by the
engagement envelope. Pacing fails open; consent and the promo slot are the hard limits.

**Events**: `affiliate_promo_generated` / `_published` / `_blocked`. `_blocked` (generation could not
produce compliant copy) is **never summed** with `affiliate_disclosure_blocked` (publish gate caught
undisclosed content) — one piece of content can fire both. Declining to write (no consent, not this
slot, paced out) is deliberately not an event.

## Surfaces shipped

[...]

PR #863, opened 2026-08-01
fix(content): keep URLs byte-identical through sanitize_for_linkedin (closes #823)

### LINKED ISSUE #823: sanitize_for_linkedin() eats underscores inside URLs — any link with two `_` params publishes broken
## Why

`sanitize_for_linkedin()` (`src/cqc_lem/utilities/linkedin_formatter.py`) strips markdown italics with

```python
re.sub(r"(?<!_)_(?!_)(.+?)(?<!_)_(?!_)", r"\1", text, flags=re.DOTALL)
```

It has no URL protection, and `_word_` is exactly the shape of a UTM'd link. Any URL in a post body
carrying **two or more** underscore-bearing query params gets the underscores between them removed:

```
https://app.example.com/signup?utm_source=referral&utm_medium=referral&utm_campaign=x&ref=12
                            -> ?utmsource=referral&utmmedium=referral&utm_campaign=x&ref=12
```

The same happens to the markdown-bold pass (`**…**`) for `*`-bearing URLs, and to any path/slug with
two underscores (`/newsletter/my_weekly_note`).

Called on finished post bodies at `run_content_plan.py:920` and `run_content_plan.py:1696`, so this
reaches published posts: a user-supplied newsletter URL or lead-magnet link with two underscores
publishes dead, and the click is silently lost.

Found while reviewing #814 (issue #770). That PR carries a **narrow** fix — the (B) writer repairs
its own referral link in `affiliate_content._restore_referral_link()`, because the referral URL
always has both `utm_source` and `utm_medium` and is the one link whose entire purpose is
attribution. The general sanitizer is untouched there deliberately (scope), which is what this
issue is for.

## Scope

- Make `sanitize_for_linkedin()` URL-safe: mask `https?://…` spans before the markdown transforms
  and restore them after (the `[text](url)` conversion still works against a masked token).
- Once the sanitizer is safe, `affiliate_content._restore_referral_link()` becomes redundant —
  remove it and its test, or keep it as a belt-and-braces assertion, but do not leave two
  half-repairs.
- Audit the other markdown regexes in the same function for the same blind spot (`**`, `` ` ``,
  `^[-*_]{3,}$`).

## Acceptance

- [x] A post body containing `https://host/x?utm_source=a&utm_medium=b` survives `sanitize_for_linkedin()` byte-identical.
- [x] A URL path with two underscores (`/newsletter/my_weekly_note`) survives byte-identical.
- [x] Markdown link/bold/italic/code stripping outside URLs is unchanged (existing tests still pass).
- [x] Unit tests in `tests/unit/utilities/test_linkedin_formatter.py` cover both URL shapes.

## Related

- #814 / #770 — where this was found, and the narrow referral-link repair
- #392 — first-comment link split (the other place a post's link has to stay intact)
- #658 — marketing attribution (`utm_source`/`utm_medium` are what this breaks)

gitchrisqueen:
## What & why

`sanitize_for_linkedin()` ran its markdown transforms over the whole post body, with no URL
protection. A URL is markdown-shaped by accident — `?utm_source=a&utm_medium=b` is exactly the
`_word_` shape the italic pass strips — so **any link carrying two underscore-bearing query params
published broken**:

```
https://app.example.com/signup?utm_source=referral&utm_medium=referral&utm_campaign=x
                            -> ?utmsource=referral&utmmedium=referral&utm_campaign=x
```

The sanitizer is called on finished post bodies (`run_content_plan.py:920`, `:1696`), so this
reached published posts: a user's newsletter URL or lead-magnet link died and the click was silently
lost. The same blind spot hit the bold pass (`**`), inline code, the horizontal-rule rule, and any
path with two underscores (`/newsletter/my_weekly_note`).

## The fix

Mask every `https?://` span before the markdown transforms and restore it verbatim after — one
change that covers every regex in the function rather than patching them one at a time.

- The mask token starts with a control char, and `normalize_public_text()` (which runs first) has
  already stripped any control char the input itself carried — so no line-start rule (header,
  bullet, numbered list) can ever match a token.
- The URL character class stops at `)`, `]`, quotes, backtick and `*` — the characters markdown
  wraps a URL in. So `[text](url)` still converts (it matches against the masked token) and a
  bolded URL still loses its markers.

With the sanitizer safe, `affiliate_content._restore_referral_link()` — the narrow repair added in
#814 for the one URL whose whole job is attribution — is redundant and **removed**, per the issue's
"do not leave two half-repairs". Its end-to-end guard test stays and now passes on the general fix.

## Acceptance

- [x] `https://host/x?utm_source=a&utm_medium=b` survives byte-identical
- [x] A path with two underscores (`/newsletter/my_weekly_note`) survives byte-identical
- [x] Markdown link/bold/italic/code stripping outside URLs unchanged (all existing tests pass)
- [x] Unit tests in `tests/unit/utilities/test_linkedin_formatter.py` cover both shapes

## Testing

New `TestSanitizeLeavesUrlsIntact` covers: UTM'd URLs, double-underscore paths, `__` in a path
(the bold-underscore pass), fragments, a URL mid-sentence with trailing punctuation, `**url**`,
`[text](url)` conversion, markdown *outside* a URL on the same line still stripped, multiple URLs
each restored to their own span, and a URL alone on a line not read as a horizontal rule.

Full unit suite green locally: `8895 passed`.

Closes #823

### DIFF
--- a/src/cqc_lem/utilities/linkedin_formatter.py
+++ b/src/cqc_lem/utilities/linkedin_formatter.py
@@ -136,10 +136,17 @@ def enforce_post_readability(text: str, max_chars: int = 2200, target_paragraph_
 # A URL is markdown-shaped by accident: `?utm_source=a&utm_medium=b` is exactly the `_word_` the
 # italic pass eats, and `**`/backtick/`---` can fall inside one just as easily. Mask every http(s)
 # span before the markdown transforms and restore it verbatim after, so a link publishes
-# byte-identical to the one the user configured (issue #823). The class stops at the characters
-# markdown wraps a URL in — `)`, `]`, quotes, backtick, `*` — so `[text](url)` still converts (it
-# matches the masked token) and a bolded URL still loses its markers.
-_URL_SPAN_RE = re.compile(r"https?://[^\s<>\"'`*)\]]+")
+# byte-identical to the one the user configured (issue #823).
+#
+# Two character classes, because a markdown marker means different things inside a URL and at its
+# edge. `*` and `_` are legal URL characters, so the span must be allowed to CONTAIN them (stopping
+# at the first `*` exposes the rest of the URL to the italic pass: `?q=a*b*c` -> `?q=abc`). But a
+# span may not END on one, or emphasis wrapped around a URL gets swallowed: `_url_` would absorb the
+# closing `_`, leaving the opener unpaired and publishing a link with a trailing underscore. Trailing
+# sentence punctuation is excluded for the same reason — it belongs to the prose, not the link.
+# `)`, `]`, quotes and backtick are excluded outright: those are what markdown wraps a URL IN, so
+# `[text](url)` still converts (it matches against the masked token) and `` `url` `` still unwraps.
+_URL_SPAN_RE = re.compile(r"https?://[^\s<>\"'`)\]]*[^\s<>\"'`)\]*_.,;:!?]")
 # Starts with a control char, so no line-start rule (header, bullet, numbered list) can match it.
 # normalize_public_text has already dropped any control char the input itself carried.
 _URL_TOKEN = "\x00lemurl{}\x00"
--- a/src/cqc_lem/utilities/linkedin_formatter.py
+++ b/src/cqc_lem/utilities/linkedin_formatter.py
@@ -133,17 +133,47 @@ def enforce_post_readability(text: str, max_chars: int = 2200, target_paragraph_
     return text
 
 
+# A URL is markdown-shaped by accident: `?utm_source=a&utm_medium=b` is exactly the `_word_` the
+# italic pass eats, and `**`/backtick/`---` can fall inside one just as easily. Mask every http(s)
+# span before the markdown transforms and restore it verbatim after, so a link publishes
+# byte-identical to the one the user configured (issue #823). The class stops at the characters
+# markdown wraps a URL in — `)`, `]`, quotes, backtick, `*` — so `[text](url)` still converts (it
+# matches the masked token) and a bolded URL still loses its markers.
+_URL_SPAN_RE = re.compile(r"https?://[^\s<>\"'`*)\]]+")
+# Starts with a control char, so no line-start rule (header, bullet, numbered list) can match it.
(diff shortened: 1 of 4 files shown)


gitchrisqueen:

Approach is right (mask URL spans around the markdown passes, restore verbatim) and every acceptance criterion in #823 is covered, with no later phase left untracked. Two real defects in the mask's character class — each one re-opening the broken-link failure the PR exists to close:

- **`*`-bearing URLs were still mangled** (the shape the issue's Why section explicitly names). `*` was excluded from the span outright, so the mask stopped at the first one and left the remainder of a legal URL exposed to the italic pass: `https://host/find?q=a*b*c` published as `?q=abc`. Fixed: the span may now CONTAIN `*`.
- **Emphasis wrapped around a URL swallowed its own closing marker.** `_` was allowed with no end guard, so `_https://host/a_` masked `https://host/a_`, leaving the opener unpaired — the post published `_https://host/a_` and the auto-linked href gained a trailing underscore. Same for `__url__`. This is a regression: `main` stripped both cleanly. Fixed: the span may not END on `*`/`_` (nor on trailing sentence punctuation, which belongs to the prose, not the link).

One rule covers both — `https?://[^\s<>"'`)\]]*[^\s<>"'`)\]*_.,;:!?]`. `**url**` still unwraps, `[text](url)` still converts, `` `url` `` still unbackticks, and a URL that genuinely ends in `_` is still returned byte-identical.

Also verified, no change needed: no markdown transform can split or delete a mask token (the token carries no `_`/`*`/digit-boundary collision), and a `\x00lemurl0\x00` token planted in the *input* is defanged before masking because `normalize_public_text` drops control chars first — so no injected token can steal a URL and no NUL can reach LinkedIn. Added a regression test asserting that.

Tests: 4 new cases in `TestSanitizeLeavesUrlsIntact` (emphasis unwrapping parametrized over `*`/`**`/`_`/`__`, `*`-bearing URL, genuine trailing `_`, no-token-leak). Full unit suite green locally: `8902 passed`.

Removing `_restore_referral_link` is sanctioned by the issue ("do not leave two half-repairs") and its end-to-end guard test still passes against the general fix.

gitchrisqueen:

PR #843, opened 2026-07-31
feat(models): refresh the Ollama Cloud roster — deepseek-v4-flash + gemma4:31b (closes #717)

gitchrisqueen:
## What

Refreshes the Ollama Cloud roster in `.litellm/config.yaml` (issue #717).

| tier | before | after |
|---|---|---|
| `lem-medium` | gpt-oss:120b, minimax-m2.7, gpt-4o-mini | **gpt-oss:120b, deepseek-v4-flash:cloud, gemma4:31b**, gpt-4o-mini |
| `lem-complex` | qwen3.5:397b, gpt-4o, claude-sonnet-4-6 | **qwen3.5:397b, deepseek-v4-flash:cloud**, gpt-4o, claude-sonnet-4-6 |

- **deepseek-v4-flash:cloud** — 284B MoE / 13B active, 1M context, native structured output, fast
  no-thinking mode, **MEDIUM** usage level. Joins both serving tiers.
- **gemma4:31b** — **Low/Medium** usage level, 256K context. Joins `lem-medium`.
- **minimax-m2.7 removed.** It is not retired, it is a version behind — and the weekly model-health
  cron re-filed that family gap on every run (that is where #717 came from). Following the family to
  `minimax-m3` moves the tier from a **MEDIUM to a HIGH** usage level, so it is an evaluation, never
  an auto-swap. Rather than pay a HIGH tier for the tier that is *not* the quality anchor,
  `lem-medium` leaves the family: v4-flash (MEDIUM) + gemma4 (Low/Medium) cover the same work.

## Quota-burn delta (the owner's caveat on #717)

Everything added is **at or below** the usage level of what it joins, so the change is **flat or
down** on Ollama quota:

| model | usage level | note |
|---|---|---|
| deepseek-v4-flash | MEDIUM | same class as gpt-oss:120b / the departed minimax-m2.7 |
| gemma4:31b | Low/Medium | cheapest deployment in `lem-medium` |
| minimax-m2.7 *(removed)* | MEDIUM | — |
| minimax-m3 *(NOT adopted)* | **HIGH** | a quota increase; gated on the benchmark, see below |
| glm-5.2 *(NOT adopted here)* | HIGH | already serves `lem-agent-tier1` only |

## Why neither new model is promoted to the head of its group

`routing_strategy: latency-based-routing` means order inside a group is not a priority list — every
deployment competes and the fastest wins, so both new models are live traffic from the first restart.
[...]

Issue #842, opened 2026-07-31
Ollama Cloud roster refresh — Phase 2: benchmark minimax-m3/glm-5.2 for lem-complex + settle the gpt-oss:120b demotion (follow-up of #717)


gitchrisqueen:
Follow-up of #717. The roster refresh (config PR) landed the two models that could be adopted on
spec — **deepseek-v4-flash:cloud** (lem-medium + lem-complex) and **gemma4:31b** (lem-medium). Two
questions from #717 and the owner's 2026-07-27 direction on it remain, and both need a **live run**
(Ollama Cloud key + `scripts/benchmark_models.py` on the box), which the config PR could not do
headless.

## Remaining scope (quoted from #717 + its owner comment)

> evaluate demoting gpt-oss:120b if v4-flash + gemma4 cover the tier

> also evaluate **minimax-m3:cloud** (and optionally glm-5.2:cloud) as the quality option for
> lem-complex — we are on minimax-m2.7 while m3 has been out, which is exactly the staleness we're
> fixing. Caveat: m2.7 is a MEDIUM usage-level model, m3 is HIGH — include the quota-burn delta in
> the PR notes and keep deepseek-v4-flash as the cost-first primary if m3's spend isn't justified by
> the long-form quality spot-check.

> a small quality spot-check (generate one comment + one long-form draft per candidate, eyeball via
> the review gate)

`minimax-m2.7` was removed from lem-medium in #717 rather than bumped to m3, so the *medium* tier's
half of that staleness is already resolved — what is left is the **lem-complex quality option**.

## What to do

1. Run the #721 harness against the live roster, champions read from the shipped config
   (`lem-medium` = `gpt-oss:120b`, `lem-complex` = `qwen3.5:397b`):
   `poetry run python scripts/benchmark_models.py --models deepseek-v4-flash:cloud,gemma4:31b,minimax-m3:cloud,glm-5.2:cloud`
2. Record each candidate's **usage level** beside its scores. minimax-m3 and glm-5.2 are HIGH where
   everything currently serving lem-medium/lem-complex is MEDIUM or below — a `recommend` verdict on
   either is a quota increase, not a free upgrade, and the report must say so in numbers.
3. Apply only what the gate emits as `recommend`, as an ordinary config PR (a promotion = moving a
   deployment to the head of its group, which also moves that tier's benchmark champion).
[...]

Issue #721, opened 2026-07-27
Model-tier evaluation harness: benchmark cron-detected Ollama candidates against tier expectations (PostHog evals) with repo-committed reports


gitchrisqueen:
## Context

The weekly model-health cron (`scripts/weekly_model_check.sh` + `scripts/model_health_check.py`, Sun 09:00 UTC) currently answers only "is the configured model alive?". #716 extends it with advance-retirement scanning, new-model discovery, and family major-version upgrade detection; #717 is the first roster-refresh decided largely on spec sheets. That is the gap: **when the cron detects a NEW or UPGRADED Ollama Cloud candidate, we adopt (or file swap PRs) on speculation, not measurement.**

Requirement: encode the expectations of each LiteLLM tier as evaluations, run every cron-detected candidate (plus the current champions as baseline) through them, and **commit a report to the repo each run** so model standing is tracked over time.

Tier expectations to encode (from `.litellm/config.yaml` aliases + `ai_helper.py` call sites):

| Tier | Contract |
|---|---|
| `lem-simple` | Short outputs ≤300 chars: refine, brief summarize, comma lists; obeys `max_tokens`, no preamble |
| `lem-medium` | LinkedIn comments (must satisfy the comment quality contract), post refinement, blog summaries |
| `lem-complex` | Long-form thought-leadership / personal-story / industry-news posts; blueprint + slop-lint clean |
| `lem-router` | Complexity classification (the `LEMComplexityRouter` contract in `.litellm/complexity_router.py`) |

Related: #716, #717, #647 (LiteLLM→PostHog native `$ai_generation` callback — CLOSED/shipped, but this harness must NOT hard-depend on the prod callback: it emits its own tagged events).

## PostHog evaluations — viability verification (2026-07-27, docs + API schema)

Verified against [PostHog Evaluations docs](https://posthog.com/docs/ai-evals) and the live LLM-analytics API surface ([API reference](https://posthog.com/docs/api/llm-analytics)). Findings:

- **Types**: `llm_judge` (LLM scores each generation against a rubric prompt → boolean pass/fail + reasoning), `hog` (deterministic code, free, ms-latency), `sentiment` (n/a here).
[...]

PR #864, opened 2026-08-01
feat(models): carry each candidate's Ollama Cloud usage level through the benchmark (#842)

gitchrisqueen:
Enabling half of #842. The measurement itself needs the box (Ollama Cloud key + a litellm restart), so this PR ships the part that can be built headless: the harness's ability to state a swap's **quota cost** beside its quality score.

## Why

The #721 gate scores quality and says nothing about price. On Ollama Cloud those are separate questions — metering is by the model's **usage level**, so promoting a HIGH model over a MEDIUM one raises quota burn on every call that tier serves. #842 asks the `minimax-m3` / `glm-5.2` question in exactly those terms:

> a `recommend` verdict on either is a quota increase, not a free upgrade, and the report must say so in numbers

Before this, `scripts/benchmark_models.py` had no notion of a usage level at all, so acceptance criterion 2 could only have been met by hand-writing the numbers into the PR body — which is how they go stale.

## What changed

- `collect_usage_levels()` reuses `model_health_check.fetch_usage_level` (injected, so the pure half stays network-free) for every measured model — **champions included**, because a delta needs both sides.
- Every scorecard gains a `Usage` column; every gate verdict and every swap recommendation carries a `usage_delta` (`up` / `flat` / `down` / `unknown`), which also rides on the JSON `--recommendations-out` writes for the cron.
- An increase renders as a ⚠️ **quota increase** line with its step count, and the Swap-recommendations section adds a "decide the extra burn deliberately" note.
- New `--usage-levels model=<low|medium|high|'extra high'>` and `--no-usage-levels`.
- Reported, **never gated** — "is the extra quota worth it" is a human call, not a threshold.

### `unknown` is deliberately not `flat`

A level that could not be read renders with the same warning an increase gets. The failure this prevents is a HIGH model being adopted as if it were free because nobody could see its level.

Unknown is expected, not a bug: ollama.com publishes the Usage stat only on **cloud-only** model pages. Verified live against ollama.com while writing this:

| Model | Usage stat on its page |
|---|---|
| `minimax-m3` | **High (3)** |
| `glm-5.2` | **High (3)** |
| `deepseek-v4-flash` | **Medium (2)** |
| `gpt-oss:120b` | none published (also pullable locally) |
| `qwen3.5:397b` | none published |
| `gemma4:31b` | none published |

So both #842 candidates for `lem-complex` are **one usage level above** the `deepseek-v4-flash` that #843 just landed there — confirmed from the source, not from a spec sheet. The incumbents' levels are a human input; `--usage-levels` is how they enter the run.

## Testing

- 20 new unit tests in `tests/unit/test_benchmark_models.py` (`TestUsageLevels` + 3 CLI tests): override precedence, no-fetch/no-network, a failing fetch degrading to unknown, up/flat/down/unknown deltas, the gate reporting without blocking, recommendations carrying the delta, both report renderings, and both sides being fetched.
- One test asserts `USAGE_LEVELS_BY_LABEL` equals `model_health_check._USAGE_LEVELS`, so an override and a scrape can never mean different things by "medium".
- `poetry run pytest tests/unit -q` → **8919 passed**.
- Rendered smoke run: `--dry-run --models minimax-m3 --tiers lem-complex --usage-levels "qwen3.5:397b=medium,minimax-m3=high"` produces
  `⚠️ **quota increase** — High (3) vs champion qwen3.5:397b Medium (2) (+1 usage level).`

## Owner decision applied — `1B 2A 3A`

**`2A` is now code, not prose.** A swap that raises the Ollama Cloud usage level is adoptable only
on `lem-complex`, and only on a **strict** judge-rate win — a tie does not buy +1 usage level.
`QUOTA_INCREASE_TIERS` + `quota_policy()` encode exactly that, so every `recommend` verdict and
every swap recommendation carries `{decision: adopt|hold, reason}`, rendered under the verdict, per
recommendation, and as a `**N held** by that policy` line naming them. It stays **reported, never
gated**: the quality verdict is unchanged and `hold` means "this one goes to the owner", not "this
one is blocked". `unknown` holds too — for the same reason it never renders as `flat`.

Rendered on a live dry run (`--models minimax-m3 --tiers lem-complex`, canned outputs tie at 100%):

```
- 💳 usage level — ⚠️ **quota increase** — High (3) vs champion `qwen3.5:397b` Medium (2) (+1 usage level). …
- 🧾 standing spend policy — **hold**: ties or trails the champion on judge rate (100% vs 100%) — a tie is not worth the extra quota.
…
**1 held** by that policy: `minimax-m3` on lem-complex — take these to the owner rather than shipping them.
```

**`1B` — the run happens unattended.** On a run nobody is watching there is no one to type
`--usage-levels`, and an unsupplied level renders `unknown`, which now HOLDS a swap: a run without
the incumbents' levels can recommend but can never conclude. So `--usage-levels` defaults to
`$BENCHMARK_USAGE_LEVELS` (an explicit flag still wins), `.env.example` carries it, and the docs
table names the four vars that separate "paste me the report" from a run that happens on its own.

The runner side is wired too (outside this diff, on the box — reported for the record, trivially
revertible from the `.bak.benchmark-env.*` copies beside each file):

- `agent-pipeline/lib/run_lane.sh` — sourcing `secrets.env` makes those shell variables, not
  environment ones, so a `claude -p` child never saw the key. It now exports exactly four:
  `OLLAMA_CLOUD_URL`, `OLLAMA_CLOUD_API_KEY`, `BENCHMARK_ENABLED`, `BENCHMARK_USAGE_LEVELS`.
- `agent-pipeline/config.env` — `BENCHMARK_ENABLED=true` and
  `BENCHMARK_USAGE_LEVELS=gpt-oss:120b=medium` (what #843's own config comment asserts).
  `qwen3.5:397b` and `gemma4:31b` are still unfilled and will render `unknown` → their swaps hold.
  Verified end to end: a child process now reads all four.

**`3A` needs no diff.** Champions are read from the shipped config head, so `gpt-oss:120b` stays
`lem-medium`'s champion and the demotion is decided on the measurement, as the acceptance criterion
asks.

## Testing

- 10 new tests in `TestQuotaPolicy` (adopt on flat/down, adopt on a strict `lem-complex` win, hold
  on a tie, hold on any other tier however good the candidate, hold with no judge rate, hold on an
  unread level, gate + recommendation both carrying it, a missing policy defaulting to `hold` not
  `adopt`, and both report renderings) + 2 CLI tests for the env default and flag precedence.
- `pytest tests/unit/test_benchmark_models.py -q` → **143 passed**.

## Remaining on #842

This PR still does **not** close the issue — no diff satisfies its five acceptance criteria. What
changed is who can do them:

- [ ] Benchmark run for all four candidates against the live champions; scorecard linked. — now a
      **tick's** job, not a hand-run: the runner's env carries the key, the flag and the levels.
- [ ] Quota/usage-level delta stated for `minimax-m3` / `glm-5.2` **with a keep-or-adopt call**. —
      the harness states the delta AND applies `2A` to it; what remains is a real measurement to
      apply it to.
- [ ] `gpt-oss:120b` demotion decided on the measurement. — `3A`: it stays champion until the run
      says otherwise.
- [ ] Any resulting swap shipped as a config PR — adopting only what the policy marks `adopt`.
- [ ] All tiers smoke-tested green after the litellm restart. — **still needs the box**; nothing
      here can do it.

The Decision Comment on this PR is answered (`1B 2A 3A`) and all three are applied above.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1024,6 +1024,11 @@ BENCHMARK_MAX_CANDIDATES=3
 # Seconds to wait for PostHog's async evaluation verdicts before a case is reported as
 # judge:timeout. A missing verdict is NEVER counted as a pass.
 BENCHMARK_JUDGE_TIMEOUT_SECONDS=90
+# Ollama Cloud usage levels (quota class) for the models whose ollama.com page publishes none —
+# every model that is also pullable locally. Same syntax as --usage-levels, which overrides it.
+# An unsupplied level renders `unknown`, and the report treats unknown as a quota RISK rather than
+# as flat, so an unattended run (issue #842) carries them here instead of on the command line.
+BENCHMARK_USAGE_LEVELS=gpt-oss:120b=medium
 # Judge scoring uses PostHog Evaluations via POSTHOG_PERSONAL_API_KEY / POSTHOG_PROJECT_ID (see
 # the PostHog section above) plus POSTHOG_API_KEY for emitting the tagged $ai_generation events.
 # PostHog's judge needs its own provider key configured IN PostHog (Ollama Cloud is not a judge
--- a/docs/model-benchmarks/README.md
+++ b/docs/model-benchmarks/README.md
@@ -82,7 +82,25 @@ on **cloud-only** model pages, so models that are also pullable locally (`gpt-os
 ```
 
 `--no-usage-levels` skips the fetch entirely (one page request per measured model); a fetch that
-fails leaves that model `unknown` rather than failing the run.
+fails leaves that model `unknown` rather than failing the run. On an **unattended** run there is
+nobody to type that flag, so the same string is read from `BENCHMARK_USAGE_LEVELS` (an explicit
+flag wins) — see *Unattended runs* below.
+
+### Standing spend policy — when a quota increase may be taken (#842)
+
+The delta says what a swap costs. The policy says who gets to accept that cost, and the owner
+settled it once (#842 decision `2A`) so a run does not park for the same question every time:
+
+> A usage-level **increase** is adoptable only on `lem-complex` — long-form is the one tier where
+> quality *is* the product — and only on a **strict** judge-rate win. A tie is not worth +1 usage
+> level on every call that tier serves.
+
+`QUOTA_INCREASE_TIERS` in `scripts/benchmark_models.py` is that rule, and every `recommend` verdict
+carries a `quota_policy` of `adopt` / `hold` with the reason, rendered under the verdict and again
+per recommendation. `hold` is **not** a gate — the quality verdict is unchanged and the swap is
+still recommended; it means the swap goes to the owner rather than into a config PR. `unknown`
+holds too, for the same reason it never renders as `flat`: a level nobody read cannot be checked
+against a rule written about increases.
 
 A recommendation is **not** a change. `.litellm/model_upgrades.yaml` is the RETIREMENT map and the
 reactive half of the model-health check auto-swaps whatever lands in it, so adopting a benchmark
@@ -108,6 +126,25 @@ poetry run python scripts/benchmark_models.py --render /tmp/bm.json --out-dir do
 `scripts/weekly_model_check.sh` invokes it automatically for the candidates the #716 catalog scan
 found, and opens the report as a PR. A benchmark failure alerts but never blocks the retirement-swap
 safety path.
+
+### Unattended runs (#842)
+
+A real run needs a key, so it used to need a person. Four env vars are the whole difference between
+"paste me the report" and a run that happens on its own:
+
+| Variable | Why it has to be in the environment |
+|---|---|
+| `BENCHMARK_ENABLED=true` | Without it `--run` prints "nothing to do" and exits 0 — a silent no-op, not an error. |
+| `OLLAMA_CLOUD_URL` | `https://ollama.com/v1`, the direct API that serves the bare model ids (`deepseek-v4-flash`, not `deepseek-v4-flash:cloud`). |
+| `OLLAMA_CLOUD_API_KEY` | The metered credential. Never in the repo, never in a report. |
+| `BENCHMARK_USAGE_LEVELS` | The incumbents' levels, which ollama.com does not publish. Unsupplied is `unknown`, and `unknown` holds a swap — so a run without it can recommend but can never conclude. |
+
+`POSTHOG_PERSONAL_API_KEY` + `POSTHOG_API_KEY` are optional: both present uses PostHog Evaluations
+as the judge, otherwise the run falls back to the in-runner judge and says so in the report.
+
+The report is the deliverable — `--results-out` for the JSON a later `--render` replays, and
+`--recommendations-out` for the swap list a follow-up config PR reads. Anything the standing spend
+policy marks `hold` is named in the report and belongs to the owner, not to the config PR.
 
 ## Leaderboard
 
--- a/docs/model-benchmarks/README.md
+++ b/docs/model-benchmarks/README.md
@@ -56,6 +56,34 @@ reported but go no further:
 | `no-baseline` | No champion measured for that tier — nothing to compare against. |
 | `reject` | Failed at least one expectation. |
 
+### Usage level — what a swap COSTS (issue #842)
+
+The gate scores quality. It says nothing about price, and on Ollama Cloud those are separate
+questions: metering is by the model's **usage level** (Low / Medium / High / Extra high), so
+promoting a High model over a Medium one raises quota burn on every call that tier serves. That is a
+spend decision, not a free upgrade.
+
+So the harness carries the level **beside** the scores and never gates on it:
+
+- every scorecard has a `Usage` column, champion and candidate alike;
+- every gate verdict carries a `usage_delta` (`up` / `flat` / `down` / `unknown`), rendered under
+  the expectations, and it rides on the JSON that `--recommendations-out` writes;
+- a recommendation that raises the level renders **⚠️ quota increase** with the step count, and the
+  Swap-recommendations section adds a "decide the extra quota burn deliberately" note.
+
+`unknown` is **not** `flat`. A level that could not be read renders with the same warning an
+increase gets — the failure this prevents is a High model being adopted as if it were free because
+nobody could see its level. Unknown is common and expected: ollama.com publishes the Usage stat only
+on **cloud-only** model pages, so models that are also pullable locally (`gpt-oss:120b`,
+`qwen3.5:397b`, `gemma4:31b`) have no level to scrape. Supply theirs from the cloud listing:
+
+```bash
+--usage-levels gpt-oss:120b=medium,qwen3.5:397b=medium
+```
+
+`--no-usage-levels` skips the fetch entirely (one page request per measured model); a fetch that
+fails leaves that model `unknown` rather than failing the run.
+
 A recommendation is **not** a change. `.litellm/model_upgrades.yaml` is the RETIREMENT map and the
 reactive half of the model-health check auto-swaps whatever lands in it, so adopting a benchmark
 winner is a deliberate edit to `.litellm/config.yaml` (or a #717-style PR). The report renders the
--- a/scripts/benchmark_models.py
+++ b/scripts/benchmark_models.py
@@ -103,6 +103,11 @@ LEADERBOARD_COLUMNS = ("Date", "Run", "Tier", "Model", "Role", "Deterministic", 
 LEADERBOARD_HEADER = "| " + " | ".join(LEADERBOARD_COLUMNS) + " |"
 LEADERBOARD_DIVIDER = "|" + "---|" * len(LEADERBOARD_COLUMNS)
 
+SCORECARD_COLUMNS = ("Tier", "Model", "Role", "Usage", "Cases", "Deterministic", "Judge", "Judged",
+                     "Timeouts", "p50", "p90")
+SCORECARD_HEADER = "| " + " | ".join(SCORECARD_COLUMNS) + " |"
+SCORECARD_DIVIDER = "|" + "---|" * len(SCORECARD_COLUMNS)
+
 # Openers that mean the model narrated instead of answering. `lem-simple` outputs are pasted
 # straight into LinkedIn copy, so a preamble is a correctness failure, not a style nit.
 _PREAMBLE_RE = re.compile(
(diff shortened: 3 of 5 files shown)


gitchrisqueen:

Held (`needs-human`, risk: product-decision). This PR only builds the *instrument*; #842's actual deliverable is a **live benchmark run** that needs the Ollama Cloud key and a litellm restart on the box, plus two spend calls that are yours. Verified live from ollama.com while writing this: **`minimax-m3` = High (3)**, **`glm-5.2` = High (3)**, **`deepseek-v4-flash` = Medium (2)** — so both `lem-complex` quality candidates are **one usage level above** what #843 just landed there.

### 1. How does the live run happen?
- **A. You run it on the box and paste the report; a follow-up tick ships any config swap** — one command, you keep the key; the pipeline does the rest. ✅ *recommended*

  ```bash
  cd /opt/lem && poetry run python scripts/benchmark_models.py --run \
    --models deepseek-v4-flash,gemma4:31b,minimax-m3,glm-5.2 \
    --usage-levels gpt-oss:120b=medium,qwen3.5:397b=<level>,gemma4:31b=<level> \
    --results-out /tmp/bm842.json --out-dir /tmp/bm842-report
  ```
  (`gpt-oss:120b=medium` is what #843's own config comment asserts. `qwen3.5:397b` and `gemma4:31b` publish no Usage stat — fill them from your Ollama cloud listing, or drop them and they render `unknown`, which the report treats as a quota risk rather than as flat.)
- **B. Put `OLLAMA_CLOUD_API_KEY` + `BENCHMARK_ENABLED` in the agent-pipeline's env so a later tick runs it unattended** — no more round-trips on this class of issue, but a benchmark key lives outside `/opt/lem/.env`.
- **C. Skip the m3/glm-5.2 evaluation entirely; keep the #843 roster as-is** — zero spend, and `lem-complex`'s quality anchor stays `qwen3.5:397b` with `deepseek-v4-flash` taking most of the traffic on latency.

### 2. If `minimax-m3` or `glm-5.2` earns `recommend`, is a HIGH usage level adoptable?
Answering this now makes the run actionable in one pass instead of two.
- **A. Adoptable only for `lem-complex`, and only if it beats `qwen3.5:397b` on the judge rate — not on a tie** — long-form is the one tier where quality is the product; a tie is not worth +1 usage level. ✅ *recommended*
- **B. Adoptable on any tier that earns it** — follow the measurement wherever it goes; accept the quota burn.
- **C. HIGH is out of budget on principle** — cap the whole stack at MEDIUM; a HIGH `recommend` is reported and ignored. (Note: `lem-agent-tier1`/`tier2-alt` already run `glm-5.2:cloud`/`minimax-m3:cloud` on the agent lane, so this would be a content-tier-only rule.)

### 3. `gpt-oss:120b` — the #717 demotion question
It is currently the **champion** of `lem-medium` (first Ollama deployment in the group), i.e. the model every future `lem-medium` candidate is measured against. Under latency-based routing, position is not priority — demoting it changes the benchmark baseline, not the live traffic split.
- **A. Keep it as champion until the run says otherwise; decide on the measurement** — this is what the acceptance criterion asks for, and Q1 supplies the data. ✅ *recommended*
- **B. Demote now — put `deepseek-v4-flash` at the head of `lem-medium`** — makes the newer MEDIUM model the baseline immediately, but sets a baseline nothing has measured.
- **C. Remove `gpt-oss:120b` from `lem-medium` entirely** — drops a working deployment on a spec-sheet argument; the tier would fall back to `gemma4:31b`/`gpt-4o-mini` under load.

**My recommendation: `1A 2A 3A`.** Option 1A costs you one command and keeps the key where it is; 2A pre-commits the only quota increase actually worth paying for (long-form quality is the one place a HIGH model earns its meter) so the run's outcome is directly actionable; 3A honours the acceptance criterion literally — the demotion is supposed to be settled *on the measurement*, and 1A produces it.

Merging **this** PR is independent of all three: it only teaches the harness to report the level, and it leaves #842 open (`Remaining on #842` in the PR body). If you want it merged ahead of the decision, say so and I'll hand it to the runner.

gitchrisqueen:
1B 2A 3A

gitchrisqueen:
Addressed your review — `1B 2A 3A` applied. Commit be514df.

**2A → code, not prose.** A swap that raises the Ollama Cloud usage level is adoptable only on `lem-complex`, and only on a **strict** judge-rate win; a tie does not buy +1 usage level. That is `QUOTA_INCREASE_TIERS` + `quota_policy()` in `scripts/benchmark_models.py`, so every `recommend` verdict and every swap recommendation now carries `{decision: adopt|hold, reason}` — rendered under the verdict, per recommendation, and as a `**N held** by that policy` line naming them. It stays **reported, never gated**: the quality verdict is untouched, and `hold` means "this one goes to you", not "this one is blocked". `unknown` holds as well, for the same reason it never renders as `flat` — a level nobody read can't be checked against a rule about increases.

Live dry run (`--models minimax-m3 --tiers lem-complex`; the canned outputs tie at 100%, which is exactly the case 2A is about):

```
- 💳 usage level — ⚠️ **quota increase** — High (3) vs champion `qwen3.5:397b` Medium (2) (+1 usage level). …
- 🧾 standing spend policy — **hold**: ties or trails the champion on judge rate (100% vs 100%) — a tie is not worth the extra quota.
**1 held** by that policy: `minimax-m3` on lem-complex — take these to the owner rather than shipping them.
```

**1B → the run can now happen with nobody watching.** Two halves:

*In this diff* — `--usage-levels` defaults to `$BENCHMARK_USAGE_LEVELS` (an explicit flag still wins), it's in `.env.example`, and the docs name the four vars that separate "paste me the report" from an unattended run. This matters more than it looks: an unsupplied level is `unknown`, and `unknown` now HOLDS a swap, so a run without the incumbents' levels can recommend but can never conclude.

*On the box, outside this diff* — I made the runner actually pass them through, since sourcing `secrets.env` produces shell variables, not environment ones, so a `claude -p` child never saw the key you'd already put there:
- `agent-pipeline/lib/run_lane.sh` now exports exactly four vars — `OLLAMA_CLOUD_URL`, `OLLAMA_CLOUD_API_KEY`, `BENCHMARK_ENABLED`, `BENCHMARK_USAGE_LEVELS`. Nothing else in `secrets.env` is exported.
- `agent-pipeline/config.env` gains `BENCHMARK_ENABLED=true` and `BENCHMARK_USAGE_LEVELS=gpt-oss:120b=medium` (what #843's own config comment asserts).

Both files were backed up in place as `*.bak.benchmark-env.<timestamp>` and edited atomically, and I verified a child process reads all four. **Flagging deliberately:** `qwen3.5:397b` and `gemma4:31b` are still unfilled, so their levels render `unknown` and any swap involving them will HOLD rather than adopt. If you can read those two off your Ollama cloud listing, adding them to that one `config.env` line is what makes the next run conclusive on `lem-medium` too. I didn't guess them.

**3A needs no diff** — champions are read from the shipped config head, so `gpt-oss:120b` stays `lem-medium`'s champion and the demotion is decided on the measurement, which is what the acceptance criterion asks for.

Tests: 10 new in `TestQuotaPolicy` + 2 CLI tests for the env default and flag precedence; `tests/unit/test_benchmark_models.py` → **143 passed**. PR body and #842 updated with what's left.

The issue stays open — the live run, the resulting config PR and the post-restart smoke test are still ahead, and the smoke test is the one part nothing headless can do.

gitchrisqueen:
Also cleared the CodeQL note on the scorecard header (2f66663): it flagged the two adjacent string literals inside the report's list as an implicit concatenation — a fair read of a line that only wrapped to fit 100 cols. The header and its divider are now derived from a `SCORECARD_COLUMNS` tuple, the way `LEADERBOARD_*` already was, so they can't drift apart either. Report output is byte-identical; 143 tests still pass.

gitchrisqueen:

- **An off-scale usage level was `unknown` in the table but a real number in the decision path** (`scripts/benchmark_models.py`) — `usage_name` clamped `level` to 1–4, `usage_delta` read `entry["level"]` raw. `model_health_check.parse_usage_level` falls back to *counting filled pips* when ollama.com's wording changes, so a markup drift returns a level like 0 or 7. With a champion at 7 and a genuine High (3) candidate, the report printed the champion's Usage as `unknown` **and** the gate as `quota decrease — High (3) vs champion \`qwen3.5:397b\` unknown (-2 usage levels)` — and `quota_policy` returned **adopt** ("no usage-level increase"), which is precisely the "adopted as if it were free because nobody could see its level" failure this PR exists to prevent. Fixed with one `usage_level()` normalizer behind `usage_name` **and** `usage_delta` (its returned `candidate_level`/`champion_level` now say what the table says), so an unreadable level is `unknown` → `hold` everywhere. A JSON-replayed `3.0` still compares (it read as `unknown` in the delta before, while the table printed `High (3)`), and `extra high` renders `Extra high (4)` rather than `.title()`'s `Extra High`. 3 tests added.
- **The `hold` never reached the human it is for** (`scripts/weekly_model_check.sh`) — the weekly cron is the only automated path that tells the owner a swap was recommended, and its alert printed a policy-held swap identically to an adoptable one. It now carries the quota direction and the spend-policy decision; legacy result files with neither key read `unknown` (verified against a fixture, `bash -n` clean).
- Also tightened a vacuous assertion: `test_a_dry_run_reads_no_usage_page_but_still_honours_an_override`'s `assert "unknown" in report` passed on the new footnote regardless of the candidate's row; it now asserts the scorecard row.

Checked and found sound: `unknown ≠ flat` throughout; the gate stays a pure quality verdict (no `usage` expectation added, `recommend` untouched); `hold` reported never gated; champions fetched too, so a delta always has both sides; `--dry-run` and `--no-usage-levels` make no network call; a failed fetch degrades to unknown rather than failing the run; overrides are keyed on the bare ids `champions_from_config` returns; `BENCHMARK_USAGE_LEVELS` is picked up by the weekly cron's existing `BENCHMARK_[A-Z_]+` sourcing; a pre-#842 results JSON still renders (no `usage`/`usage_delta`/`quota_policy` → `unknown`, no held block, no crash); `USAGE_LEVELS_BY_LABEL == model_health_check._USAGE_LEVELS` is asserted.

**Close honesty:** the PR body carries no closing keyword and states the five remaining criteria on #842 — correct, the live run, the keep-or-adopt call, the `gpt-oss:120b` demotion, the config PR and the post-restart smoke test are all still open there.

`pytest tests/unit -q` → **8934 passed** (146 in `test_benchmark_models.py`).

gitchrisqueen:

Issue #865, opened 2026-08-01
Verify the Ollama agent lane on the bare model ids — Phase 2 (follow-up of #844)


gitchrisqueen:
Follow-up of #844, which carried this acceptance box:

> - [ ] One Ollama-lane tick verified end to end.

#844 could not tick that box from a worktree: the config change ships in the repo, but the live
LiteLLM container reads `/opt/lem/.litellm/config.yaml` and only picks up the new ids when the
release carrying them is deployed. Everything else on #844 is done and verified — see
https://github.com/christopherqueenconsulting/linkedin_engagement_manager/issues/844#issuecomment-5149971960
for the probe (all four bare ids answer 200 against `https://ollama.com/v1` directly).

## Precondition — do not start before this is true

The release containing #844 must be live. Check `GET /api/app-info` against the deployed version, or
confirm `/opt/lem/.litellm/config.yaml` has `model: openai/glm-5.2` (no `:cloud`) for
`lem-agent-tier1`. If it does not, stop and leave this issue for the next tick.

## What to do

```bash
BASE=/home/lem/agent-pipeline . /home/lem/agent-pipeline/lib/capacity.sh
_litellm_probe lem-agent-tier2   # 0 served / 1 usage-limited / 2 broken
cat /home/lem/agent-pipeline/state/.litellm_probe_detail
```

Then let one real tick route to the Ollama lane (`OLLAMA_CAPACITY_PCT=80` pins it healthy for the
test) and confirm it produces work rather than dying on the alias.

## Acceptance
- [ ] `_litellm_probe` returns `0` for `lem-agent-tier2` on the deployed config, and
      `state/ollama.state` shows `alias_ok=1` stamped after the deploy.
- [ ] One tick dispatched on the Ollama lane reaches a real result (PR opened, or a review/fix
      commit pushed) — not an `Invalid model name` failure.
- [ ] Result recorded here; close.

If the lane fails on the bare ids after the deploy, that is a regression of #844 — reopen it with
the probe output rather than re-adding the `:cloud` tag, since the tag is invisible to the
advance-retirement scan and to shadow-cost pricing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

gitchrisqueen:
## Verification result — end to end, on the deployed config

[...]

PR #866, opened 2026-08-01
fix(litellm): use the bare catalog ids for the agent-lane Ollama tiers (closes #844)

### LINKED ISSUE #844: Agent-tier Ollama ids use a `:cloud` tag the direct ollama.com API does not list
Found during the adversarial review of PR #843 (issue #717). Out of that PR's scope, so it is filed
rather than fixed there.

## The problem

The four `lem-agent-*` aliases in `.litellm/config.yaml` are configured with a `:cloud` tag:

```yaml
lem-agent-tier1      openai/glm-5.2:cloud
lem-agent-tier2      openai/kimi-k2.7-code:cloud
lem-agent-tier2-alt  openai/minimax-m3:cloud
lem-agent-tier3      openai/nemotron-3-super:cloud
```

`OLLAMA_CLOUD_URL` is `https://ollama.com/v1` — the **direct cloud API**. Per
`docs.ollama.com/cloud` ("Cloud API access", mirrored in `tests/unit/fixtures/ollama/cloud.md`) that
endpoint serves exactly the ids `ollama.com/api/tags` lists, and the committed catalog snapshot
(`.litellm/ollama_catalog_snapshot.json`, generated from that endpoint) lists them **bare**:
`glm-5.2`, `kimi-k2.7-code`, `minimax-m3`, `nemotron-3-super`. The `-cloud` suffix (hyphen, not
colon) is the `ollama pull` form for the **local** runtime.

Every LEM serving-tier deployment — `gpt-oss:20b`, `gpt-oss:120b`, `qwen3.5:397b`, and now
`deepseek-v4-flash` / `gemma4:31b` from #843 — uses the bare id. The agent tiers are the only
deployments that do not.

## Why it matters even if the lane "works"

If the ids are wrong, three things follow:
1. Every call to that alias 404s, and LiteLLM's `fallbacks:` chain drops tier1 → tier2 → tier3, all
   of which would fail the same way. That is consistent with the "tier1 intermittently times out"
   note in `scripts/agent-pipeline/docs/agent-pipeline-routing.md` but not proof of it.
2. `plan_retirement_notices` keys configured deployments by their bare id and matches the
   retirement table verbatim, so an announced retirement of `glm-5.2` would never match
   `glm-5.2:cloud` — the advance warning silently does not cover this lane.
3. None of the four has a `.litellm/model_prices_snapshot.json` entry, so there is no shadow-cost
   reference for them either.

`tests/unit/test_litellm_roster_config.py` currently **excludes** the `lem-agent-*` groups from its
verbatim-catalog and price-snapshot assertions, with a comment pointing here. That exclusion should
be deleted once this is settled.

## What to do

1. Determine empirically which id the endpoint serves — one `curl https://ollama.com/api/tags` with
   `OLLAMA_CLOUD_API_KEY` on the box settles it, plus a one-token chat completion against
   `glm-5.2` and `glm-5.2:cloud`.
2. If bare is correct: drop the `:cloud` tags from the four aliases, update
   `scripts/agent-pipeline/lib/dispatch.sh` / `agent-pipeline-routing.md` comments, and add
   price-snapshot entries with a `shadow_reference` for each.
3. Delete the `AGENT_GROUPS` exclusion in `tests/unit/test_litellm_roster_config.py` so all Ollama
   deployments are checked verbatim.
4. Run one agent-pipeline tick on the Ollama lane and confirm it produces work.

## Acceptance
- [x] The id the direct API actually serves is established and recorded on this issue.
- [x] `.litellm/config.yaml` agent-tier ids match it; docs/scripts comments match.
- [x] Every `lem-agent-*` Ollama id has a `model_prices_snapshot.json` entry with a `shadow_reference`.
- [x] `AGENT_GROUPS` exclusion removed from `tests/unit/test_litellm_roster_config.py`; suite green.
- [ ] One Ollama-lane tick verified end to end. → needs the deployed config; tracked on #865.

Needs the box (Ollama Cloud key), so the probe half cannot run headless in a worktree.

gitchrisqueen:
The four `lem-agent-*` deployments in `.litellm/config.yaml` carried a `:cloud` tag no other Ollama deployment in that file uses. #844 filed it on the theory that those ids 404. **They do not** — and that turned out to matter, because it changes what the actual defect is.

## What the probe found

Ran it on the box against the direct cloud API (`https://ollama.com/v1`, real `OLLAMA_CLOUD_API_KEY`), 2026-08-01. Full output on the [issue](https://github.com/christopherqueenconsulting/linkedin_engagement_manager/issues/844#issuecomment-5149971960).

- `ollama.com/api/tags` lists 18 ids, **all bare** — no `:cloud` entry.
- One-token completions: `glm-5.2` **200**, `glm-5.2:cloud` **200**, and the same for the other three pairs.
- Controls: `glm-5.2:latest` 200, but `glm-5.2:bogus` / `glm-5.9` / `not-a-real-model` all **404**.

So tags *are* validated and `:cloud` is an alias the endpoint **resolves**, like `:latest`. Live lane state agrees (`state/ollama.state` `alias_ok=1`, `.litellm_probe_detail` `http=200` with a real body).

## So what was actually broken

A wrong-but-accepted spelling is the nastier case: nothing in the request path notices, only the tooling that reads the config **by id**.

1. **Advance-retirement warning did not cover this lane.** `plan_retirement_notices` matches the docs.ollama.com/cloud retirement table verbatim — an announced retirement of `glm-5.2` could never have matched `glm-5.2:cloud`. This is the one thing the weekly cron exists to give us.
2. **Shadow cost silently read zero.** None of the four had a `model_prices_snapshot.json` entry, and `estimate_shadow_cost_usd` looks the id up exactly (no substring fallback, unlike `estimate_llm_cost_usd`). Every agent-lane call reported no shadow cost — subscription traffic the margin report could not see.
3. The verbatim-catalog assertion had to exclude the whole group to stay strict about the serving tiers.

## Changes

- `.litellm/config.yaml` — bare ids for all four tiers; the "note 0" header now records that `:cloud` **serves** rather than implying it 404s, so the next reader does not re-derive the wrong conclusion.
- `.litellm/model_prices_snapshot.json` — entries for `openai/glm-5.2`, `openai/kimi-k2.7-code`, `openai/minimax-m3`, `openai/nemotron-3-super`. `shadow_reference` follows each family's committed precedent where it has one (`glm-5.1`→4o-mini, `kimi-k2.5/k2.6`→4o, `minimax-m2.5/m2.7`→4o-mini) and a size-comparable peer where it does not (`nemotron-3-super`, 230B → 4o-mini, same as `minimax-m2.5`).
- `tests/unit/test_litellm_roster_config.py` — `AGENT_GROUPS` exclusion deleted; the catalog, pricing and retirement assertions now cover **all 12** Ollama deployments. Added one guard that pins the four agent-tier ids by name, because a re-introduced tag would otherwise fail as "not in the catalog snapshot" with no hint of what to write instead.
- `scripts/agent-pipeline/docs/agent-pipeline-routing.md`, `scripts/agent-pipeline/lib/dispatch.sh` — comments/table match. Also corrected the implication that tier1's intermittent timeouts were the bad id; they are the model being slow.

No app code changed. `scripts/agent-pipeline/lib/labels.sh` already used bare ids, so config and telemetry labels now agree.

## Testing

`8903 passed` (full `tests/unit`), including the 13 in `test_litellm_roster_config.py` — 12 of which now check deployments they previously skipped.

## Scope

Follow-up: #865 — the fifth acceptance box ("One Ollama-lane tick verified end to end") needs the deployed config, since the live LiteLLM container reads `/opt/lem/.litellm/config.yaml` and only picks the new ids up after the release ships. It is filed with the precondition and the exact probe commands.

Closes #844

### DIFF
--- a/.litellm/config.yaml
+++ b/.litellm/config.yaml
@@ -4,9 +4,16 @@
 #     `gpt-oss:120b-cloud` and not `gpt-oss:120b:cloud`. OLLAMA_CLOUD_URL is the direct
 #     ollama.com/v1 API (docs.ollama.com/cloud, "Cloud API access"), which serves exactly the ids
 #     `ollama.com/api/tags` lists; the `-cloud` suffix is the `ollama pull` form for the LOCAL
-#     runtime. Get this wrong and the deployment 404s on every call — and under (1) a 404 is the
+#     runtime. An id that catalog has never listed 404s on every call — and under (1) a 404 is the
 #     FASTEST answer in the group, so the broken deployment wins the race and takes the traffic.
 #     That is how the retired ministral-3:8b kept winning lem-simple.
+#     Careful: a WRONG-BUT-ACCEPTED spelling is the nastier case, because it does not 404. Probed
+#     2026-08-01 (issue #844): `glm-5.2:cloud` and `glm-5.2:latest` both answer 200 and serve the
+#     same model as bare `glm-5.2`, while `glm-5.2:bogus` 404s — so `:cloud`/`:latest` are tag
+#     ALIASES the endpoint resolves, not typos it rejects. Nothing in the request path notices,
+#     but everything that reads this file by id does: the catalog check, the advance-retirement
+#     scan (scripts/model_health_check.py keys on the bare id and matches the retirement table
+#     verbatim) and shadow-cost pricing all key on the exact string. Keep ids verbatim-catalog.
 #  1. `routing_strategy: latency-based-routing` means the order of deployments inside a
 #     model_name group is NOT a priority list — every deployment in the group competes and the
 #     fastest wins. Order still matters for ONE thing: scripts/benchmark_models.py treats the
@@ -186,14 +193,17 @@ model_list:
   # ANTHROPIC_AUTH_TOKEN=$LITELLM_MASTER_KEY) with --model = one of these aliases. LiteLLM serves
   # the Anthropic /v1/messages endpoint and translates to the underlying Ollama Cloud model.
   # CLOUD-ONLY by policy (Ollama Max subscription) — no local models on this path.
-  #   tier1   glm-5.2:cloud        hardest long-horizon / architecture / multi-step plans
-  #   tier2   kimi-k2.7-code:cloud  default coding / patches / tests / multi-file edits
-  #   tier2a  minimax-m3:cloud      premium parallel / vision-capable alternate
-  #   tier3   nemotron-3-super:cloud fallback / reviewer / reasoning lane
+  #   tier1   glm-5.2         hardest long-horizon / architecture / multi-step plans
+  #   tier2   kimi-k2.7-code  default coding / patches / tests / multi-file edits
+  #   tier2a  minimax-m3      premium parallel / vision-capable alternate
+  #   tier3   nemotron-3-super fallback / reviewer / reasoning lane
+  # These four carried a `:cloud` tag until #844. It served fine (see note 0 — the endpoint
+  # resolves `:cloud` as a tag alias), but it is not what the catalog lists, so the retirement
+  # scan and the shadow-cost lookup could not see them. Bare ids now, like every tier above.
   # See docs/agent-pipeline-routing.md for the capacity-aware lane-selection rule.
   - model_name: lem-agent-tier1
     litellm_params:
-      model: openai/glm-5.2:cloud
+      model: openai/glm-5.2
       api_base: os.environ/OLLAMA_CLOUD_URL
       api_key: os.environ/OLLAMA_CLOUD_API_KEY
       max_parallel_requests: 2
@@ -202,7 +212,7 @@ model_list:
 
   - model_name: lem-agent-tier2
     litellm_params:
-      model: openai/kimi-k2.7-code:cloud
+      model: openai/kimi-k2.7-code
       api_base: os.environ/OLLAMA_CLOUD_URL
       api_key: os.environ/OLLAMA_CLOUD_API_KEY
       max_parallel_requests: 4
@@ -211,7 +221,7 @@ model_list:
 
   - model_name: lem-agent-tier2-alt
     litellm_params:
-      model: openai/minimax-m3:cloud
+      model: openai/minimax-m3
       api_base: os.environ/OLLAMA_CLOUD_URL
       api_key: os.environ/OLLAMA_CLOUD_API_KEY
       max_parallel_requests: 4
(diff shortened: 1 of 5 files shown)


gitchrisqueen:

Verified first: probe conclusion holds against everything in-repo that reads these ids. All four bare ids are in `.litellm/ollama_catalog_snapshot.json`; `plan_family_upgrades` still returns `[]` on the new roster (no new cron issue); `champions_from_config` unchanged; `scripts/agent-pipeline/lib/labels.sh` already used bare ids; no `:cloud` string survives outside prose. Full `tests/unit`: **8904 passed**.

- **New agent-lane guard pinned the four ids by literal name — would fail CI on the retirement cron's own correct swap.** This PR is exactly what makes the `lem-agent-*` deployments visible to `scripts/model_health_check.py`, and `scripts/weekly_model_check.sh` doesn't just warn — `apply_swaps` rewrites the `model:` line in place when a configured id retires, then opens a PR to land it. Ollama retires on a months cadence (this config already records ministral-3:8b, kimi-k2:1t, minimax-m2.7), so `assert _ollama_models("lem-agent-tier1") == ["glm-5.2"]` is a scheduled red build on a 3am automated PR, failing with a message that reads like the swap was wrong. **Fixed:** matched by SHAPE instead — no `:cloud`/`-cloud`/`:latest` suffix on ANY Ollama deployment, which is the regression actually being guarded, and the assertion message still names what to write (`glm-5.2:cloud (lem-agent-tier1) carries a ':cloud' tag … write 'glm-5.2'` — verified by reintroducing the tag). Added a second assertion that each agent tier keeps exactly one Ollama deployment, so the shape guard can't pass vacuously on an emptied group.

- **"Shadow cost silently read zero for the agent lane" is not the mechanism** — the four price entries are right to add, the reason given for them isn't. `estimate_shadow_cost_usd` is only ever called from `track_llm_call`, and agent-pipeline traffic never reaches it: the `claude` CLI hits LiteLLM's `/v1/messages` directly, so its cost is LiteLLM's native `$ai_generation` callback. `scripts/agent-pipeline/docs/agent-pipeline-routing.md` says so itself under *Token/cost accounting*, two sections below the paragraph this PR added — the doc now contradicted itself, and a reader would come away believing the margin report covers the agent lane. **Fixed:** the config comment, the routing doc and the test docstring now state the accurate rationale — the entries keep the roster uniform and price these ids from day one for the promotion case (this config's own comments have `glm-5.2` / `minimax-m3` as live `lem-complex` candidates), while the advance-retirement blind spot is the real defect the bare ids fix. No price data changed.

Checked and NOT flagged: `shadow_reference` picks follow committed family precedent and are a documented judgement call; `re` is still used in the test module after the `AGENT_GROUPS` deletion; the `openai/glm-5.2` addition doesn't perturb `estimate_llm_cost_usd`'s substring fallback for any existing id; `Closes #844` is honest — boxes 1–4 ship here, box 5 needs the deployed config and is tracked on #865 with a precondition, linked from both the PR body and the issue.

gitchrisqueen:

PR #867, opened 2026-08-01
feat(ui): prompt to refresh when /api/app-info reports a new release (closes #754)

### LINKED ISSUE #754: SPA new-version awareness: prompt to refresh from /api/app-info — Phase 2 (follow-up of #743)

gitchrisqueen:
## What & why

Phase 2 of #743 — the third, explicitly optional item from that issue.

#743's two layers are **reactive**: asset retention (`api/spa_assets.py`) and the one silent reload
(`ui/src/utils/chunkReload.ts`) both fire only once a lazy chunk has already failed to load.
Releases batch 4x daily, so a long-lived tab is routinely several builds behind — and it can keep
*working* while running old client code against a newer API, in which case nothing ever fails and
neither layer says a word.

This adds the proactive half:

- **`ui/src/hooks/useNewVersion.ts`** polls `GET /api/app-info` every `VERSION_POLL_INTERVAL_MS`
  (5 min). A hidden tab issues **no request at all** (`document.visibilityState`), and a
  `visibilitychange` listener re-checks the moment it comes back, so a backgrounded tab catches up
  instead of polling in the dark.
- The baseline is the **first version this tab ever read**, held in module scope — remounting the
  notice can't re-baseline onto a build that shipped after boot.
- It **never reloads**. All it does is flip the existing `NewVersionNotice` on; the one automatic
  reload stays owned by `chunkReload.ts`, so the two signals can't both act on a tab.
- Unknown is never "new": an unreachable endpoint, or a missing/blank/non-string version, raises
  nothing **and** does not become the baseline.
- `NewVersionNotice` is reused (no second banner). Same heading + Refresh button; only the secondary
  line differs — "part of the app could not load" for the chunk case, "Refreshing picks up the
  latest version" for this one. The chunk-failure wording wins when both are true.

## Testing

- `src/cqc_lem/ui/src/hooks/useNewVersion.test.tsx` — 12 vitest cases with fake timers: baseline on
  first read, change detected on the next poll with nothing having failed, identical version quiet
  forever, every-poll-fails quiet, blank/missing/non-string version quiet, **no requests while
  hidden**, catch-up on `visibilitychange`, polling stops once the prompt is up, interval cleared on
  unmount, plus the notice wiring (renders nothing on the current release; shows the alert and the
  Refresh button on a change, with `location.reload` never called by the hook itself).
- `chunkReload.test.tsx` mocks the api client so it stays about the chunk half only.
- Full UI suite green locally: `npm test` → 37 files, 301 tests. `npm run lint` clean on every
  touched file; `npm run build` (tsc + vite) passes.

Docs: `docs/zero-downtime-deploys.md` § "Stale lazy chunks in already-open tabs" now describes three
layers (two reactive + this prompt); `CLAUDE.md`'s gotcha updated to match.

Closes #754


gitchrisqueen:

- **Hidden-tab boot read never baselined (`useNewVersion.ts`)**: `poll()` returned early on `document.visibilityState === 'hidden'` *including the mount call*, so a tab opened in the background (ctrl-click, restored session) — which still loads and RUNS the bundle while hidden — recorded no baseline until the user first looked at it. If a release shipped in that gap, the baseline became a build the tab isn't running and the prompt could never fire for exactly the tab this feature exists to catch (it also contradicted the module's own "baseline is the version this tab booted with" invariant). Fixed: the boot read ignores visibility (`poll(true)`), every later poll still respects it — a hidden tab issues one request at load and nothing after. Added `still baselines at boot in a tab that opened in the background` to `useNewVersion.test.tsx`; CLAUDE.md + `docs/zero-downtime-deploys.md` updated to match.

Also checked, no findings: `/api/app-info` is in `_PUBLIC_API_PREFIXES`, so the poll can't trip the axios 401 interceptor and silently log a signed-in user out; the notice is mounted once in `App.tsx` (no duplicate banner) and the chunk-failure wording still wins when both signals are true; the effect tears its interval and `visibilitychange` listener down on unmount and once the prompt is up; unknown/blank/non-string versions neither raise nor baseline; `Closes #754` is honest — all five acceptance boxes are covered by the diff and #754 is itself the last phase of #743.

Local verification: `npx tsc --noEmit` and `eslint` clean on the touched files. Vitest could not run in this worktree (node 18 here; the suite's rolldown build needs `node:util.styleText`, node 20+) — CI is the source of truth for the suite.

gitchrisqueen:

PR #869, opened 2026-08-01
feat(security): identity + session hardening — public_uid, hashed sessions, httpOnly cookie, auth limits (refs #745)

### LINKED ISSUE #745: feat(security): identity + phishing-resistant auth + encrypt LinkedIn session at rest — Phase 2 (dropped follow-up of #568)

gitchrisqueen:
Phase **2b** of issue #745 — the approved Phase-2 build order (`docs/AUTH_SECURITY_DESIGN.md` §8). 2a (encryption at rest) shipped in #807; this is identity + session hardening. **2c (passkeys/TOTP/step-up) is still open, so this PR does not close #745.**

## What changed

| Before | After |
|---|---|
| `sessions.session_token` stored the token in **plaintext** | stores `SHA-256(token)` — a DB dump yields no live session |
| the token sat in `localStorage`, readable by any script on the page | lives in an **httpOnly / Secure / SameSite=Lax** cookie |
| one opaque session row, no way to see or end a device | per-device rows (label, pseudonymised IP, last seen) + revocation |
| the email address WAS the identity | `users.public_uid` is the identity; email is an attribute that can move |
| `/auth/email/init` + `/verify` unlimited against a 6-digit PIN | per-email + per-IP limiter **and** a durable PIN lockout |
| no record of logins, failures or session changes | every one appended to `auth_audit_log` |

## Threat-model note

- **T1/T6 (DB leak / insider):** the sessions table no longer contains anything that can be replayed — only SHA-256 of the token. The migration rewrites existing rows in place, so **the deploy signs nobody out**: the browser still holds the same plaintext token and the server hashes it on the way in. The hash is deliberately UNKEYED — the token is 256 bits of `secrets.token_hex(32)`, so there is nothing to brute-force, and keying it would mean a rotated `LEM_SECRET_KEY` logged every user out.
- **XSS / token theft:** the token leaves JavaScript's reach entirely. ~150 SPA call sites still pass `session_token`, so rather than rewrite all of them, `AuthContext` now holds the non-secret sentinel `'cookie'` and `api/main.get_session_user_id()` resolves the request from the cookie (a `ContextVar` set by `session_cookie_middleware`). An explicit token that RESOLVES still wins — non-browser callers, the LinkedIn OAuth `state` round trip and the tutorial capture harness are unchanged.
- **CSRF (new exposure, cookies are ambient):** `SameSite=Lax` keeps the cookie off cross-site POSTs; every mutating endpoint is JSON-bodied (no CORS middleware exists, so a cross-origin preflight has nothing to succeed against); with `API_ACCESS_TOKENS` set, `/api/*` also needs the bearer token. Written up in `docs/identity-and-sessions.md` so the next person who adds CORS sees it first.
- **PIN brute force:** Redis fixed windows (5 init / 10 verify per email per hour, 30 per IP) that **fail open** if the broker is down, plus the durable half — after `PIN_MAX_ATTEMPTS` (5) wrong guesses the outstanding PIN locks for 15 minutes and even the correct PIN is refused. A fresh `init` clears the lock, which is exactly what the per-email init limiter bounds: worst case ~25 guesses/hour against a 10^6 space.
- **Account takeover via email:** changing the address needs a PIN sent to the **new** address, writes `user_email_history`, and revokes every OTHER session. An address owned by someone else gets the same generic 400 as any other rejection (a distinct "already registered" would be an account-existence oracle).
- **Extension:** the SPA no longer has a token to hand the browser extension, so `POST /user/extension-token` mints it its OWN labelled session — visible beside every other device and revocable without signing the person out of the app.

## Migration (`V20260801031800__identity_session_hardening_phase_2b.sql`)

`users.public_uid` (+ backfill) and `email_verified_at`; `user_email_history`; per-device columns + index on `sessions`; `auth_audit_log`; `attempts` / `locked_until` on `email_pin_auth`; and one in-place `UPDATE sessions SET session_token = LOWER(SHA2(session_token, 256))`. Additive — no column or table is dropped, and nothing is destructive. The token rewrite is deliberately unconditional (a token and its hash are both 64 chars, so length cannot tell them apart; Flyway runs a migration exactly once).

## Testing

- `tests/unit/api/test_auth_hardening.py` — cookie issued/cleared, sentinel and stale-token resolution, 429s and lockout, per-device list/revoke scoping, email-change flow.
- `tests/unit/utilities/test_auth_rate_limit.py` — window bounds, IP hashing, fail-open on a dead/broken Redis.
- `tests/unit/utilities/test_db_identity_hardening.py` — plaintext never reaches SQL, revocation scoping, PIN lockout, `public_uid`, email history, audit writes.
- `tests/integration/test_identity_session_hardening.py` — the same guarantees against a **live migrated MySQL**, so the schema and the code are proven to agree.
- Full unit lane green locally: 8911 passed.

## Ops

New env in `.env.example`: `SESSION_COOKIE_{NAME,SECURE,SAMESITE}`, `AUTH_INIT_MAX_PER_HOUR`, `AUTH_VERIFY_MAX_PER_HOUR`, `AUTH_IP_MAX_PER_HOUR`, `PIN_MAX_ATTEMPTS`, `PIN_LOCKOUT_MINUTES`. All have working defaults — **prod needs no change**, since it is HTTPS behind Cloudflare. A plain-http local origin needs `SESSION_COOKIE_SECURE=false`, or the browser drops the cookie (the SPA falls back to holding the token in that case rather than locking anyone out).

Full posture: `docs/identity-and-sessions.md`.

**Remaining on #745:** **2c** — passkeys (`webauthn`) + TOTP (`pyotp`) + recovery codes (`argon2-cffi`) + step-up gate + enrollment UI, prerequisite being a valid TLS certificate on the public hostname. The issue stays open for it.


gitchrisqueen:
Held (`needs-human`, risk: `migration` + `security`). This PR rewrites every row in `sessions` and changes how the SPA proves who it is, so it wants your sign-off before it merges.

### 1. The cookie cutover — how far do we take it in this PR?
- **A. As built** — the token moves into an httpOnly cookie, the SPA holds the non-secret sentinel `'cookie'`, and if the cookie fails to stick (`http://` origin, browser blocking it) `AuthContext` falls back to holding the real token so nobody is locked out. ✅ *recommended*
- **B. Cookie set but unused** — the API issues the cookie and the SPA keeps sending the real token from `localStorage`. Zero behaviour change, but **no XSS win** — the whole point of 2b's session half is deferred.
- **C. Hard cookie-only** — drop the fallback. Strictest, but any environment where the cookie doesn't stick becomes a login loop, and that includes a plain-http local dev box until `SESSION_COOKIE_SECURE=false` is set.

### 2. Existing sessions at deploy — sign everyone out, or not?
- **A. Keep them (as built)** — the migration hashes each stored token in place, so the token in your browser still works and no one notices the deploy. ✅ *recommended*
- **B. Invalidate everything** — add a `DELETE FROM sessions` to the migration. Cleanest possible cut (nothing pre-2b survives), but every signed-in user has to request a fresh PIN, including you, right when a release lands.

### 3. Rate-limit defaults (per hour): 5 PIN emails/address, 10 verifies/address, 30 auth calls/IP.
- **A. Ship these** — bounds a PIN walk to ~25 guesses/hour against a 10^6 space; the limiter fails open if Redis is down, and a successful login clears the counters. ✅ *recommended*
- **B. Raise the IP bucket to 100/hour** — safer for a shared-NAT office or a corporate VPN where many users appear as one address; weaker against a single host walking many addresses.
- **C. Per-email only** — drop the IP bucket entirely. No false positives ever; nothing then stops one host spraying many addresses.

**My recommendation: `1A 2A 3A`.** A gets the real security win (the token leaves JavaScript's reach) without a cutover that can lock anyone out, and hashing in place means the hardening ships invisibly instead of forcing a fleet-wide re-login during a release window.

Note on scope: this PR is **2b only** and deliberately carries no closing keyword — **2c** (passkeys + TOTP + recovery codes + step-up gate) stays open on #745, gated on the public hostname serving a valid TLS certificate.

gitchrisqueen:
Ok

gitchrisqueen:
Addressed your review: you replied `Ok`, so I applied all three recommendations — **1A 2A 3A**. All three were what the PR already shipped, so **no behavioural change was needed**; I re-verified each against the diff rather than taking my own summary on trust:

- **1A — cookie cutover as built.** `AuthContext.tsx` holds the non-secret sentinel `COOKIE_SESSION = 'cookie'` and falls back to holding the real token when the cookie does not stick (`http://` origin / browser blocking), so no environment turns into a login loop. Unchanged.
- **2A — keep existing sessions.** The migration contains exactly one session statement, `UPDATE sessions SET session_token = LOWER(SHA2(session_token, 256))` — no `DELETE`/`TRUNCATE`, so the deploy signs nobody out. Unchanged.
- **3A — ship the rate-limit defaults.** `env_constants.py` reads 5 / 10 / 30 per hour plus `PIN_MAX_ATTEMPTS=5`, `PIN_LOCKOUT_MINUTES=15`, matching `.env.example`. Unchanged.

**One thing I did change, and it is not from your review:** the required check `CLAUDE.md ≤ 40,000 chars` was failing. That was staleness, not this PR — `main` compressed CLAUDE.md under the cap in #888 after this branch forked, and against the older 41,269-char base my one added bullet pushed the file to 42,351. I merged `origin/main` in (clean, no conflicts, also clears the 17-commit lag Codecov flagged); CLAUDE.md is now **31,748 chars** and the identity+sessions bullet is intact.

Full unit lane green on the merge result: **9008 passed**.

Scope is unchanged and still 2b-only — no closing keyword, **2c** (passkeys + TOTP + recovery codes + step-up gate) stays open on #745 behind the TLS-certificate prerequisite.

gitchrisqueen:

Reviewed as the adversarial gate (I did not write this). CI was green and the close is honest — the PR correctly does **not** carry `Closes #745`, states "Remaining on #745: 2c", and #745 stays open, so no phase is left living in prose. The four findings are all at the seams the cookie cutover creates; each fix ships with a test that fails against the pre-fix behaviour. Pushed as `28ae03d`.

- **The acting session could differ from the authenticating one** (`api/main.py`, medium): `current_session_token()` returned a stale explicit token even when `get_session_user_id()` had fallen through it to the cookie — the exact case the PR designs for and tests in `test_a_stale_explicit_token_falls_back_to_the_live_cookie`. Consequences: `auth_logout` called `delete_session(<stale>)`, deleting nothing and leaving the live session row valid until expiry after a "successful" sign-out; `revoke_other_sessions(keep_token=<stale>)` matched no row, so **"sign out all other devices" revoked the caller's own live session**; `is_current` was never true on the Security card; and the email-change audit wrote `changed_by_session_id=None`. Both resolvers now share one order — an explicit token only wins when it resolves, and the resolve is skipped when there is no cookie to fall through to, so non-browser callers cost no extra query.
- **The per-IP auth limiter was bypassable with one header** (`api/main.py`, medium): `_client_ip()` read the FIRST entry of `X-Forwarded-For`. A proxy *appends* to the chain the client supplied — Cloudflare included — so that entry is attacker-chosen: one host could mint itself a fresh `AUTH_IP_MAX_PER_HOUR` bucket on every request, and every `ip_hash` in `auth_audit_log` and `sessions` was forgeable. Now reads `CF-Connecting-IP` first (Cloudflare sets it on everything it proxies and overwrites what the client sent, and the origin is tunnel-only, so it cannot be chosen); XFF stays as the fallback for a deployment with no Cloudflare in front. Documented in `docs/identity-and-sessions.md` so the next edge change knows which header belongs at the top.
- **The bypass login claimed the email was verified** (`api/main.py`, low-medium): the no-mail-provider branch called `mark_email_verified()`, stamping `users.email_verified_at` on the one login path where nobody proves control of the address. Dropped — that column exists to record the proof, and 2c's step-up gate is meant to be able to trust it. The email-PIN and email-change paths still stamp it.
- **Logout cleared the cookie after an unguarded audit write** (`api/main.py`, low): `record_auth_event` ran before `_clear_session_cookie`, against the handler's own comment that "whatever happens to the audit trail, the session row and the cookie must still go". A DB hiccup there returned a 500 with the cookie still set. Cookie clear moved ahead of it and the audit write is now swallowed.

Also checked and found sound, so recorded rather than changed: the migration's unconditional token rewrite (a token and its hash are both 64 chars, `sessions.session_token` is `VARCHAR(64) UNIQUE`, Flyway runs it once, and nobody is signed out); `SameSite=Lax` against the OAuth return trip (LinkedIn's callback is a top-level GET, so the cookie rides — and every GET endpoint in `main.py` is a read, so Lax's one opening is not exploitable here); `revoke_session` scoping by `user_id` (a foreign id is a 404, verified in both the unit and integration lanes); the unkeyed session hash (256-bit token, and keying it would make a `LEM_SECRET_KEY` rotation a mass logout); and the ContextVar under `BaseHTTPMiddleware` (the downstream task is spawned from inside `dispatch`, so the value is copied into it — the cookie-resolution tests exercise this end to end through `TestClient`).

Full unit lane green after the fixes: 9014 passed.

gitchrisqueen:
🧩 phase-guard — **held before merge.** This PR closes issue #745, whose body declares work beyond this PR (phase: "Phase 2"), and no follow-up issue is linked anywhere on this PR or that issue. Merging now would close #745 and drop that scope silently — exactly how #548 was lost.

To clear this, do ONE of:
2. Remove `Closes #745` from the PR body, say what remains, and re-label this PR `agent:working` — the issue stays open.

Assigning @gitchrisqueen.

gitchrisqueen:
Follow-up: #897

Phase **2c** (passkeys + TOTP + recovery codes + step-up gate) is now filed as #897 with topical labels, `agent:ready` and `priority:high`. #745 stays open until 2c lands.

This PR's body already stated it does not close #745; #897 makes the remaining scope tracked rather than implied, which is what the guard was asking for. Re-labelling `agent:working`.

gitchrisqueen:

PR #871, opened 2026-08-01
fix(groups): rotate past a group LEM cannot post in (closes #858)

### LINKED ISSUE #858: fix(groups): a group LEM cannot post in starves the weekly rotation forever
## Why

#769 (PR #856) made the weekly group post rotate: `get_next_group_for_post` picks the
post-enabled group with the oldest `last_posted_at`, and `record_group_post` stamps that
column **only after a post actually shipped** — deliberately, so a transient failure
(session dead, selectors flaky) leaves the group next in line instead of skipping its turn.

That is right for a transient failure and wrong for a permanent one. Groups where members
cannot post (admin-only / announcement groups) render no share box, so `auto_post_to_group`
returns `"Group share box not found"` every time. `last_posted_at` stays NULL, NULL sorts
FIRST, and that group is "next" **every week forever** — no other post-enabled group ever
gets a turn, and the Groups card keeps naming it as the next post. Group sync opts every
joined group into posting by default (`post_enabled DEFAULT 1`), so one such group in a
user's list starves all the rest.

Pre-existing in effect (before #769 the weekly post was pinned to one arbitrary group), but
#769 is what states "the weekly slot rotates" in the SPA, so this is now a promise the code
can break.

## Scope

- Track when a group post RUN happened, separately from when one succeeded — e.g. a
  `last_post_run_at DATETIME NULL` column stamped on every run that reached the group
  (success and structural failure), with the rotation ordering on it while `last_posted_at`
  stays the truthful "we actually posted here" record.
- A group that turns out to be unpostable then moves to the back of the queue and is retried
  only after every other post-enabled group, instead of blocking them.
- Keep the transient-failure behaviour intact where it matters: a run that never reached the
  group (no session / profile failure) must NOT count as a try.
- Needs a TIMESTAMP-versioned migration (`V<YYYYMMDDHHMMSS>__*.sql`) — additive column only.

## Acceptance

- [x] A post-enabled group whose post attempt fails structurally is not selected again while
      another post-enabled group has gone longer without a run.
- [x] `last_posted_at` still reflects only successful posts (the API payload and any copy
      built on it stay truthful).
- [x] A failure before the group page is reached does not advance the rotation.
- [x] Unit tests cover the ordering and both stamp paths (≥80% patch coverage).
- [x] All required CI gates pass.

Follow-up of #769 / PR #856 (found in adversarial review).

gitchrisqueen:
## What & why

#769 made the weekly group post rotate on `user_groups.last_posted_at`, stamped **only after a post actually shipped** — deliberately, so a transient failure leaves the group next in line instead of skipping its turn.

That is right for a transient failure and wrong for a permanent one. A group whose members cannot post (admin-only / announcement) renders no share box, so `auto_post_to_group` returns `"Group share box not found"` every time, `last_posted_at` stays NULL, NULL sorts FIRST — and that group is "next" **every week forever**. Group sync opts joined groups into posting by default, so one such group starves every other post-enabled group and the Groups card keeps naming it as next.

## The change

Split "we tried here" from "we posted here":

- **Migration** `V20260801065636__add_user_group_last_post_run_at.sql` — additive `last_post_run_at DATETIME NULL`, backfilled from `last_posted_at` so existing rows keep their place in the rotation (without the backfill every user's rotation would restart at the alphabetically first group).
- **`get_next_group_for_post`** orders on `COALESCE(last_post_run_at, last_posted_at)` instead of `last_posted_at`. An unpostable group moves to the back of the queue and is retried only after every other post-enabled group.
- **`record_group_post`** stamps both columns (a successful post is also a run); new **`record_group_post_run`** stamps only the run column.
- **`auto_post_to_group`** calls `record_group_post_run` on the three *structural* misses (share box / editor / Post button not found) — the group loaded but its composer did not — and logs one INFO line. Everything else is unchanged: a failure before the group is reached (no session / profile error), an empty generation, or an unexpected exception stamps **neither** column, so the group keeps its turn.

`last_posted_at` keeps its exact meaning, so `GET /user/groups` and the Groups card stay truthful about what actually shipped.

## Testing

`tests/unit/app/test_groups.py`:
- ordering asserts the run column, not the success column;
- `record_group_post` stamps both, `record_group_post_run` never touches `last_posted_at`;
- parametrized over all three structural misses: run stamped, post not;
- session failure before the group is reached: neither stamped;
- empty generation: neither stamped.

`tests/unit/utilities/test_db_coverage_errors.py` covers the new function's mysql-error fallback.

```
pytest tests/unit/app/test_groups.py tests/unit/utilities/test_db_coverage_errors.py \
       tests/unit/utilities/test_db_coverage_avatar_groups.py tests/unit/api/test_api_coverage_misc.py
275 passed
```

Closes #858

### DIFF
--- a/compose/local/database/migrations/V20260801065636__add_user_group_last_post_run_at.sql
+++ b/compose/local/database/migrations/V20260801065636__add_user_group_last_post_run_at.sql
@@ -1,0 +1,22 @@
+-- Separate "we tried to post here" from "we posted here" (issue #858, follow-up of #769).
+--
+-- #769 made the weekly group post rotate on `last_posted_at`, stamped ONLY after a post actually
+-- shipped so a transient failure (dead session, flaky selectors) leaves the group next in line
+-- rather than skipping its turn. That is right for a transient failure and wrong for a permanent
+-- one: a group whose members cannot post (admin-only / announcement) renders no share box, so its
+-- `last_posted_at` stays NULL, NULL sorts FIRST, and it is "next" every week forever — starving
+-- every other post-enabled group. Group sync opts joined groups into posting by default, so one
+-- such group in a user's list is enough.
+--
+-- `last_post_run_at` is stamped by every run that REACHED the group and found it unpostable, plus
+-- every successful post; the rotation orders on it. `last_posted_at` keeps its meaning untouched —
+-- the truthful record of real posts, which is what the API payload and the Groups card show. A run
+-- that never reached the group (no session / profile failure) stamps neither, so transient
+-- behaviour is unchanged.
+--
+-- Backfilled from `last_posted_at` so existing rows keep their place in the rotation instead of all
+-- collapsing to NULL (which would restart every user's rotation at the alphabetically first group).
+ALTER TABLE user_groups
+    ADD COLUMN last_post_run_at DATETIME NULL DEFAULT NULL AFTER last_posted_at;
+
+UPDATE user_groups SET last_post_run_at = last_posted_at;
--- a/src/cqc_lem/app/run_automation.py
+++ b/src/cqc_lem/app/run_automation.py
@@ -35,7 +35,7 @@ from cqc_lem.utilities.db import get_user_password_pair_by_id, get_user_id, inse
     get_engagement_preferences, count_comments_today, get_recent_engagers, upsert_engager, \
     get_newsletter_settings, mark_newsletter_published, record_newsletter_subscriber_stat, \
     get_newsletter_edition, mark_edition_published, mark_edition_failed, \
-    upsert_user_group, get_enabled_group_ids, record_group_post, record_post_stats, \
+    upsert_user_group, get_enabled_group_ids, record_group_post, record_group_post_run, record_post_stats, \
     get_recent_posted_post_ids, get_uncaptured_posted_post_ids, \
     get_shipped_variant_keys, \
     get_lead_magnet_settings, has_received_lead_magnet, record_lead_magnet_sent, \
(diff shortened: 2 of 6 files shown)


gitchrisqueen:
Held (`needs-human`, risk: migration). This PR adds `user_groups.last_post_run_at` and **backfills it** from `last_posted_at`, so it needs your sign-off before merge.

### 1. Ship the migration as written — additive column + one backfill `UPDATE`?
`V20260801065636__add_user_group_last_post_run_at.sql`: `ADD COLUMN last_post_run_at DATETIME NULL DEFAULT NULL`, then `UPDATE user_groups SET last_post_run_at = last_posted_at;`. Nothing is dropped or rewritten; the only existing column read is `last_posted_at`.
- **A. Ship as written** — existing rows keep their position in the rotation.  ✅ *recommended*
- **B. Column only, no backfill** — every row starts NULL, so on the first run after deploy every user's rotation restarts at the alphabetically first post-enabled group (harmless, but one group loses its turn).
- **C. Don't ship** — the starvation bug stays; an admin-only group keeps being "next" every week.

### 2. Which failures count as "we tried this group" and advance the rotation?
Today the PR advances it **only** on the three structural misses (share box / post editor / Post button not found — the group loaded but its composer did not). A dead session, a profile error, an empty LLM generation, or an unexpected exception all leave the group next in line.
- **A. Structural misses only** — a group that genuinely can't be posted in rotates away after one week; transient noise never costs a group its turn.  ✅ *recommended*
- **B. Also advance on any exception after the group page loads** — recovers faster if a group breaks in some new way, but one flaky Selenium week silently skips a healthy group.
- **C. Also advance when generation returns nothing** — an LLM hiccup is user-wide, not group-specific, so this would rotate for a reason unrelated to the group.

**My recommendation: `1A 2A`.** The backfill is what keeps the rotation continuous across the deploy, and limiting the new stamp to composer-not-there keeps #769's "a transient failure never skips a turn" guarantee exactly as it was.

gitchrisqueen:
Ok

gitchrisqueen:
Addressed your review: read your `Ok` on the Decision Comment as **1A + 2A** (every ✅ recommended option). Both are already exactly what this branch does, so **no code changed** — confirming rather than guessing:

**1A — ship the migration as written.** `V20260801065636__add_user_group_last_post_run_at.sql` is unchanged: `ADD COLUMN last_post_run_at DATETIME NULL DEFAULT NULL AFTER last_posted_at`, then the single `UPDATE user_groups SET last_post_run_at = last_posted_at;` backfill. Nothing dropped or rewritten; existing rows keep their position in the rotation across the deploy.

**2A — structural misses only.** `auto_post_to_group` calls `record_group_post_run` from exactly three sites — share box / post editor / Post button not found. A dead session or profile error (returns before the group is reached), an empty generation, and any unexpected exception all still stamp **neither** column, so #769's "a transient failure never costs a group its turn" guarantee is untouched. `last_posted_at` still moves only on a real post, so `GET /user/groups` and the Groups card stay truthful.

Also ticked the five acceptance boxes on #858 — all are met by this diff and CI is fully green (Unit, Integration, E2E, CodeQL, GitGuardian, Migration Versions, codecov patch 100%). Re-ran the touched suites locally against this worktree: `tests/unit/app/test_groups.py` + `tests/unit/utilities/test_db_coverage_errors.py` → 143 passed.

Nothing deliberately left out; no side-instructions in your reply to file as issues. Handing back to the runner for merge.

gitchrisqueen:

Reviewed the full diff against #858's acceptance criteria and the surrounding code (`auto_group_posts` weekly dispatch, `find_first`/`click_first` miss semantics, `/user/groups` + `is_next_post`, the #769 migration this one builds on). The core of the fix holds up: ordering on `COALESCE(last_post_run_at, last_posted_at)` with the backfill keeps every existing row's place in the rotation; `record_group_post` stamping both columns means the success path never double-stamps; the three `_unpostable` call sites are exactly the composer-not-there misses, and a dead session / profile error / empty generation / unexpected exception all still stamp neither column, so #769's "a transient failure never costs a group its turn" guarantee is intact. `Closes #858` is honest — all five acceptance criteria are covered by the diff, and the issue declares no later phase.

Two things I would have blocked a human PR for, both fixed in `8473332`:

- **Silent failure on the new stamp**: `record_group_post_run` reported a MySQL error through the legacy `myprint` shim (INFO). Prod forwards WARNING and above to PostHog, so the *one* failure that re-creates this exact bug — stamp never lands → group stays least-recently-tried → "next" again next week — was invisible, and `_unpostable` ignores the `False` because there is nothing it can do with it. Now `log_error(..., exc=err, user_id=…, task_name=…)`, matching db.py's modern call sites (83 of them) and the `utilities/CLAUDE.md` level table. Added `test_record_group_post_run_failure_is_visible`.
- **Groups card copy no longer matched the behaviour**: it told the user the weekly slot goes to whichever group with Post on "has gone longest without one" — i.e. without a *post*. After this PR it goes to the one that has waited longest for its *turn*, and those are different groups in precisely the announcement/admin-only case the PR exists for. Rewrote the sentence and named the unpostable case explicitly, so a user seeing "never posted" on a group that is not marked **Next post** has the explanation in front of them.

Not changed (deliberate, and worth stating): a 429/auth wall or an SDUI selector break makes a *healthy* group look structurally unpostable, so it uses up its turn. That is the behaviour you signed off as `2A`, it can only cost a turn rather than starve the rotation, and `last_posted_at` stays truthful throughout — so `GET /user/groups` and the card still only ever claim real posts.

Tests: `tests/unit/app/test_groups.py` + `tests/unit/utilities/test_db_coverage_errors.py` + `tests/unit/utilities/test_db_coverage_avatar_groups.py` → 168 passed locally. The push re-triggers CI.

gitchrisqueen:

PR #879, opened 2026-08-01
fix(errors): stop the group-feed sort-control miss escalating as a defect (closes #872)

### LINKED ISSUE #872: fix(errors): RecurringWarning: Selector miss: Feed sort control (1x)
## Why

> Selector miss: Feed sort control

- First seen: `2026-07-31T19:05:20.634000Z` · last seen: `2026-07-31T19:05:20.634000Z`
- SDK: `posthog-python`

[Open the issue in PostHog](https://us.posthog.com/project/475262/error_tracking/019fb991-1489-79b2-b4b3-b07512d7ae4f) — it has the stack trace, the grouped occurrences and the affected people.

## Scope

## Files
- `tests/unit/`

## Acceptance
- All required CI gates pass.

Auto-filed from PostHog Error Tracking. Dedup marker (do not remove): `posthog-issue-019fb991-1489-79b2-b4b3-b07512d7ae4f`

gitchrisqueen:
## What & why

PostHog filed a `RecurringWarning: Selector miss: Feed sort control` against
`cqc_lem.app.run_automation.auto_comment_in_groups`.

Root cause: `_switch_feed_to_recent` is documented as a *silent* no-op when LinkedIn's "Sort by"
control isn't present, but it looked the control up with `find_first(..., required=False)`, which
emits `Selector miss: <label>` at **WARNING**. `auto_comment_in_groups` points the feed commenting
engine at `https://www.linkedin.com/groups/<id>/`, and a group feed never renders the home-feed sort
control — so the warning fired on every enabled group, every run. Per the recurrence rule in
`src/cqc_lem/utilities/CLAUDE.md` ("once is a warning, repeatedly is a defect"), `log_escalation`
re-emitted it at ERROR and captured one grouped `$exception`, which the daily error→issue cron filed
as #872. It was a defect report for behaviour that works.

## Change

- `selenium_util.find_first` gains `warn_on_miss: bool = True`. When `False`, a non-required total
  miss logs the same structured message (label, tried selectors, url, ids) at **DEBUG** instead of
  WARNING. The flag is carried through the `max_try` retry recursion — otherwise the final attempt
  would warn anyway.
- `_switch_feed_to_recent` passes `warn_on_miss=False` for the "Sort by" lookup only.
- Everything else is unchanged: the `Recent sort option` miss (menu opened, option gone) is real
  selector rot and still warns/escalates, as does every other `required=False` call site.
- One line added to `src/cqc_lem/utilities/CLAUDE.md` documenting the Selenium form of the
  "don't warn on an expected no-op" rule.

## Testing

- `tests/unit/utilities/test_selenium_find_first.py` — expected miss logs DEBUG and not WARNING with
  the context intact; and it survives the retry pass (`max_try=2`) rather than warning on the last
  attempt.
- `tests/unit/app/test_feed_sort.py` — the sort-control lookup passes `warn_on_miss=False`.
- Full suite locally: `8907 passed`.

Closes #872

### DIFF
--- a/src/cqc_lem/app/run_automation.py
+++ b/src/cqc_lem/app/run_automation.py
@@ -1130,10 +1130,12 @@ def _roster_activity_url(profile_url: str) -> str:
 
 def _switch_feed_to_recent(driver, wait) -> None:
     """Best-effort: flip the feed sort from 'Top' to 'Recent' so golden-hour posts surface for
-    commenting. Silent no-op if the 'Sort by' control isn't present."""
+    commenting. Silent no-op if the 'Sort by' control isn't present — group feeds
+    (`auto_comment_in_groups`) never render it, so a WARNING here escalates on every group run and
+    files a defect for working behaviour."""
     try:
         btn = find_first(driver, wait, [(By.XPATH, "//button[contains(normalize-space(),'Sort by')]")],
-                         "Feed sort control", required=False)
+                         "Feed sort control", required=False, warn_on_miss=False)
         if btn is None:
             return
         # The control reads "Sort by: Recent" once flipped — skip re-opening the menu so the second
--- a/src/cqc_lem/app/run_automation.py
+++ b/src/cqc_lem/app/run_automation.py
@@ -91,6 +91,7 @@ from cqc_lem.utilities.observability import track_post_outcome, track_audience_s
     track_comment_outcome, track_golden_hour_report, track_company_page_invite_run, \
     track_catchup_run, attribute_llm_cost, llm_attribution, \
     FEATURE_COMMENT, FEATURE_CONTENT, FEATURE_DM
+from cqc_lem.utilities.env_constants import MAX_WAIT_RETRY
 from cqc_lem.utilities.selenium_util import click_element_wait_retry, \
     get_element_wait_retry, get_elements_as_list_wait_stale, getText, close_tab, get_driver_wait_pair, quit_gracefully, \
     wait_for_ajax, find_first, click_first, find_all_first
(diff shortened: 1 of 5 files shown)


gitchrisqueen:

- **Over-broad suppression: the home feed lost its selector-rot signal.** `_switch_feed_to_recent` doesn't only run on group feeds — `navigate_to_feed` (`run_automation.py:151`) and `comment_on_feed_inline` (`run_automation.py:1393`) call it on the HOME feed, where the 'Sort by' control genuinely exists. A blanket `warn_on_miss=False` means a LinkedIn rename of that control would leave the recency-dominant scoring engine silently reading a 'Top' feed with no warning anywhere — the exact rot `find_first`'s structured miss logging was built to catch. Fixed by deciding it per surface: new `_feed_sort_control_expected(driver)` warns only on `/feed`; group feeds and activity pages log the miss at DEBUG. An unreadable `current_url` counts as not-expected (a missed signal is cheaper than a defect filed on a guess). Tests added for all three surfaces.
- **Same surface check drops `max_try` to 1 off the home feed** — a control that is never on the page isn't worth 2 × `WAIT_DEFAULT_TIMEOUT` (15s) plus a 5s retry sleep for every enabled group, every run.

Verified and NOT flagged: the `Recent sort option` miss still warns (only reachable once the button was found, so it is real rot); `warn_on_miss` is correctly threaded through the retry recursion and is a no-op under `required=True`; `Closes #872` is honest — all three acceptance criteria are covered by this diff with no later phase left in prose.

Full unit suite green locally: `8909 passed`.

gitchrisqueen:

gitchrisqueen:

PR #880, opened 2026-08-01
fix(errors): stop the optional reaction fly-out miss escalating as a defect (closes #873)

### LINKED ISSUE #873: fix(errors): RecurringWarning: Selector miss: Open reactions menu (1x)
## Why

> Selector miss: Open reactions menu

- First seen: `2026-07-31T18:44:15.836000Z` · last seen: `2026-07-31T18:44:15.836000Z`
- SDK: `posthog-python`

[Open the issue in PostHog](https://us.posthog.com/project/475262/error_tracking/019fb97d-cff4-78d0-8462-7eb060a283f9) — it has the stack trace, the grouped occurrences and the affected people.

## Scope

## Files
- `tests/unit/`

## Acceptance
- All required CI gates pass.

Auto-filed from PostHog Error Tracking. Dedup marker (do not remove): `posthog-issue-019fb97d-cff4-78d0-8462-7eb060a283f9`

gitchrisqueen:
## What

`react_to_post_inline` looks for the hover-revealed `button[aria-label='Open reactions menu']` on each feed card. When it is absent the function already degrades exactly as documented — it clicks the primary React toggle directly and leaves a default Like ("Better a Like than no reaction").

But the lookup ran through `click_first(..., required=False)`, which emits `Selector miss: Open reactions menu` at **WARNING**. Repeated across the cards of an `auto_comment_in_groups` run, the recurrence rule in `log_escalation.py` re-emitted it at ERROR, captured a grouped `$exception`, and the daily error→issue cron filed #873 — a defect report for a path doing precisely what it was designed to do.

## Why this shape

The miss is only a signal when there is **nothing to fall back to**:

- `click_first` now carries `find_first`'s `warn_on_miss` through (it was added in #872 but only reachable via `find_first`, and most call sites go through `click_first`).
- The opener lookup passes `warn_on_miss=trigger is None`. With a Reaction-state button or React toggle in hand, the fly-out is optional and its miss logs at DEBUG. With neither, the card's reaction controls are genuinely unreadable — that still warns and still escalates, so real SDUI rot is not silenced.

This follows the same posture as #872 (`Selector miss: Feed sort control`) and the `src/cqc_lem/utilities/CLAUDE.md` rule "do not warn on an expected no-op", which is extended here with the fallback half of the test.

## Testing

- `tests/unit/utilities/test_selenium_find_first.py` — `click_first` passes `warn_on_miss` down to `find_first` (DEBUG, not WARNING) and still warns by default.
- `tests/unit/app/test_comment_inline.py` — the opener miss is `warn_on_miss=False` when a fallback toggle exists, and `warn_on_miss=True` when `find_first` returns nothing at all (and the call still returns `False`).
- Full suite green locally: `8913 passed`.

Closes #873

### DIFF
--- a/src/cqc_lem/app/run_automation.py
+++ b/src/cqc_lem/app/run_automation.py
@@ -975,8 +975,13 @@ def react_to_post_inline(driver, wait, card, post_content: str = None, comment_t
                 time.sleep(random.uniform(0.6, 1.2))
             except Exception:
                 pass
+        # The fly-out opener is optional: with a `trigger` in hand its absence just means we take the
+        # documented default-Like fallback below, which is working behaviour — warning about it every
+        # card escalated into a filed defect (issue #873). With no trigger there is no fallback, so
+        # the card's reaction controls really are unreadable and the miss is worth the signal.
         opened = click_first(driver, wait, [(By.CSS_SELECTOR, "button[aria-label='Open reactions menu']")],
-                             "Open reactions menu", parent_element=card, required=False, user_id=user_id)
+                             "Open reactions menu", parent_element=card, required=False,
+                             warn_on_miss=trigger is None, user_id=user_id)
         time.sleep(random.uniform(0.8, 1.6))
         # The fly-out can render just outside the card subtree, so match the reaction button globally
         # (only one menu is open at a time and click_first filters to visible elements).
--- a/src/cqc_lem/utilities/CLAUDE.md
+++ b/src/cqc_lem/utilities/CLAUDE.md
@@ -50,9 +50,11 @@ Two things follow for anyone writing a call site here:
    Same idea for empty-result chatter: `db.get_ready_to_post_posts` logs INFO only when the list is
    non-empty, DEBUG otherwise. For Selenium, `find_first(..., required=False, warn_on_miss=False)`
    logs the miss at DEBUG — use it when the element is legitimately absent on some surfaces
-   (`_switch_feed_to_recent`'s 'Sort by' control does not exist on a group feed). Decide it PER
-   SURFACE, not once for the call site: the same lookup on the home feed, where the control does
-   exist, is selector rot and must still warn.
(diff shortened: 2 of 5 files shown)


gitchrisqueen:

- **`click_first`'s `warn_on_miss` passthrough only covered half its miss paths** (`src/cqc_lem/utilities/selenium_util.py:641`): the flag reached `find_first` for the "never found" miss, but the found-then-un-clickable branch (`element_to_be_clickable` timeout, `ElementNotInteractableException`, `StaleElementReferenceException`) still emitted `Click miss: Open reactions menu` at WARNING unconditionally. Both branches return `None`, so `react_to_post_inline` takes the identical default-Like fallback either way — and the fly-out opener is *hover-revealed*, which is precisely the kind of control that goes un-clickable between the find and the click. Per card across an `auto_comment_in_groups` run that re-escalates to ERROR and files the same defect #873 closes, one branch over. Fixed: `Click miss:` now emits through the same `log_warning`/`log_debug` selector as `Selector miss:`; default stays WARNING so only a caller that has declared a working fallback is quieted. Added `test_click_miss_also_honours_warn_on_miss` + `test_click_miss_warns_by_default`, and extended the `utilities/CLAUDE.md` note to say the flag covers both paths.

Checked and found clean:
- `warn_on_miss=trigger is None` is the right predicate — `trigger = state or <React toggle>`, and the only `return False` on a missing fly-out is guarded by `trigger is None`, so the miss is silenced exactly when a fallback exists.
- No second warning leaks from the same card: `if opened is None or click_first(...)` short-circuits, so a `React <reaction>` selector miss is not also emitted when the opener was absent.
- `Selector miss: React <reaction>` after the menu *did* open still warns — genuine rot is not silenced.
- New kwarg is keyword-only, so no positional call site breaks; `find_first` already carries `warn_on_miss` through its retry recursion.
- Close is honest: #873's three acceptance criteria (downgraded, unit test, CI) are all met and the issue declares no later phase.
- Full unit suite green locally: `8915 passed`.

gitchrisqueen:

Issue #878, opened 2026-08-01
fix(errors): RecurringWarning: Could not leave a reaction on post (1x)


gitchrisqueen:
## Why
PostHog error tracking grouped this exception into an issue that is still active: **RecurringWarning**

> Could not leave a reaction on post

- Occurrences (last 24h): **1** across 1 distinct actor(s)
- First seen: `2026-07-31T18:44:16.989000Z` · last seen: `2026-07-31T18:44:16.989000Z`
- Celery task: `cqc_lem.app.run_automation.auto_comment_in_groups`
- SDK: `posthog-python`

[Open the issue in PostHog](https://us.posthog.com/project/475262/error_tracking/019fb97d-cff2-7ce1-b37d-7908a35c4db0) — it has the stack trace, the grouped occurrences and the affected people.

## Scope
- Fix the root cause of the exception, not the symptom.
- If it is an EXPECTED best-effort failure that already degrades gracefully (a Selenium selector miss, a third-party timeout the caller retries), stop raising it into error tracking: catch it and `log_warning(...)` instead, or add a suppression rule in PostHog.
- Keep the change scoped to this exception — no unrelated refactors.

## Files
Start from the stack trace on the PostHog issue; add/extend the unit tests:
- `tests/unit/`

## Acceptance
- The exception no longer occurs (or is deliberately downgraded to a warning).
- A unit test covers the failing path (≥80% patch coverage).
- All required CI gates pass.

Auto-filed from PostHog Error Tracking. Dedup marker (do not remove): `posthog-issue-019fb97d-cff2-7ce1-b37d-7908a35c4db0`

gitchrisqueen:
Closing as a **duplicate of #816**, per the owner's decision on PR #881 (`2A`: close the reaction cluster's siblings as duplicates of #816 and take them off the agent queue).

`Could not leave a reaction on post` is not an independent defect — it is the fourth symptom of the one fault #816 tracks: the reaction anchors have rotted (`Open reactions menu` 30, `Reaction state` 17, `Reaction state (post-click)` 16 misses in 48h) and reactions are not landing. #873, #875, #877 and #881 have each collapsed one of the other symptoms into a quieter level; this one is the reaction cluster's remaining independent warning, so it must not follow them before #816 is actually fixed.

[...]

Issue #874, opened 2026-08-01
fix(errors): RecurringWarning: Selector miss: Reaction state (1x)


gitchrisqueen:
## Why
PostHog error tracking grouped this exception into an issue that is still active: **RecurringWarning**

> Selector miss: Reaction state

- Occurrences (last 24h): **1** across 1 distinct actor(s)
- First seen: `2026-07-31T18:43:03.734000Z` · last seen: `2026-07-31T18:43:03.734000Z`
- Celery task: `cqc_lem.app.run_automation.auto_comment_in_groups`
- SDK: `posthog-python`

[Open the issue in PostHog](https://us.posthog.com/project/475262/error_tracking/019fb97c-c83a-7460-bdc6-e64d0a38f3e0) — it has the stack trace, the grouped occurrences and the affected people.

## Scope
- Fix the root cause of the exception, not the symptom.
- If it is an EXPECTED best-effort failure that already degrades gracefully (a Selenium selector miss, a third-party timeout the caller retries), stop raising it into error tracking: catch it and `log_warning(...)` instead, or add a suppression rule in PostHog.
- Keep the change scoped to this exception — no unrelated refactors.

## Files
Start from the stack trace on the PostHog issue; add/extend the unit tests:
- `tests/unit/`

## Acceptance
- The exception no longer occurs (or is deliberately downgraded to a warning).
- A unit test covers the failing path (≥80% patch coverage).
- All required CI gates pass.

Auto-filed from PostHog Error Tracking. Dedup marker (do not remove): `posthog-issue-019fb97c-c83a-7460-bdc6-e64d0a38f3e0`

gitchrisqueen:
Parked pending a decision — the adversarial review of PR #881 found that `Selector miss: Reaction state` is **not** the benign un-reacted card the PR assumes: #816 records 17 pre-click misses vs 16 post-click misses in 48h (near 1:1), i.e. the anchor is rotted in both reads and reactions are not landing. Options + recommendation are on the PR: https://github.com/christopherqueenconsulting/linkedin_engagement_manager/pull/881#issuecomment-5150803673 — reply there (or here) with option letters, e.g. `1A 2A`, or `ok`.

gitchrisqueen:
[...]

PR #882, opened 2026-08-01
fix(errors): stop the post-click reaction-state miss escalating as a defect (closes #875)

### LINKED ISSUE #875: fix(errors): RecurringWarning: Selector miss: Reaction state (post-click) (1x)
## Why

> Selector miss: Reaction state (post-click)

- First seen: `2026-07-31T19:08:23.134000Z` · last seen: `2026-07-31T19:08:23.134000Z`
- SDK: `posthog-python`

[Open the issue in PostHog](https://us.posthog.com/project/475262/error_tracking/019fb993-dd6b-7003-beaa-5e3e43f20301) — it has the stack trace, the grouped occurrences and the affected people.

## Scope

## Files
- `tests/unit/`

## Acceptance
- All required CI gates pass.

Auto-filed from PostHog Error Tracking. Dedup marker (do not remove): `posthog-issue-019fb993-dd6b-7003-beaa-5e3e43f20301`

gitchrisqueen:
## What & why

PostHog filed `RecurringWarning: Selector miss: Reaction state (post-click)` (1 actor,
`auto_comment_in_groups`) — the third of the same family after #872 and #873: a WARNING on a
Selenium miss that the code already handles as a documented fallback, repeated per feed card until
`log_escalation` re-emitted it at ERROR and turned working behaviour into a defect.

`react_to_post_inline` re-reads the card's `Reaction button state` toggle after clicking, purely as
a best-effort confirm — its stated contract is *"If the toggle can't be re-read, trust the click
rather than false-negative."* But the lookup used the default `warn_on_miss=True`, so every card
that goes through the React-toggle / fly-out path (i.e. never exposed a Reaction-state button in the
first place) warned about not finding one afterwards.

The miss only carries information when there was something to re-read:

- **pre-click read found the toggle** → it was readable before the click and isn't after. That is
  real SDUI rot — still WARNING, still escalates.
- **pre-click read found nothing** → the miss *is* the trust-the-click fallback. Logged at DEBUG,
  and `max_try=1` so we stop paying `WAIT_DEFAULT_TIMEOUT` twice plus a 5s retry sleep, per card,
  for a control that surface doesn't render.

Return values are unchanged: an unreadable toggle still never false-negatives a reaction that
landed. Scoped to this one lookup — the pre-click `Reaction state` read is a different fingerprint
and is left warning, so the SDUI-drift signal is fully retained (it is tracked separately as #874,
alongside #877 `React toggle` and #878 `Could not leave a reaction on post`; the family root cause
is #816).

## Testing

- `tests/unit/app/test_comment_inline.py`: two new cases — the post-click confirm passes
  `warn_on_miss=False` + `max_try=1` when the card never had the toggle (and still returns `True`),
  and keeps `warn_on_miss=True` + `MAX_WAIT_RETRY` when the pre-click read found it.
- `poetry run pytest tests/unit/app -q` → 1786 passed.

Closes #875

### DIFF
--- a/src/cqc_lem/app/run_automation.py
+++ b/src/cqc_lem/app/run_automation.py
@@ -996,11 +996,17 @@ def react_to_post_inline(driver, wait, card, post_content: str = None, comment_t
             driver.execute_script("arguments[0].click();", trigger)
         time.sleep(random.uniform(0.8, 1.5))
         wait_for_ajax(driver)
+        # Best-effort confirm: label flips away from 'no reaction'. If the toggle can't be re-read,
+        # trust the click rather than false-negative — so the miss only means something when the
+        # card HAD a readable Reaction-state button before the click. Cards that never exposed one
+        # (the fly-out/React-toggle path) take the documented trust-the-click fallback, and warning
+        # about that on every card escalated to ERROR and filed a defect for working behaviour
+        # (issue #875). A control this card never carried is also not worth waiting out twice.
+        expected = state is not None
         after = find_first(driver, wait, [(By.CSS_SELECTOR, "button[aria-label^='Reaction button state']")],
                            "Reaction state (post-click)", parent_element=card, required=False,
-                           visible_only=True, user_id=user_id)
-        # Best-effort confirm: label flips away from 'no reaction'. If the toggle can't be re-read,
-        # trust the click rather than false-negative.
+                           visible_only=True, warn_on_miss=expected,
+                           max_try=MAX_WAIT_RETRY if expected else 1, user_id=user_id)
         if after is not None and "no reaction" in (after.get_attribute("aria-label") or "").lower():
             return False
         myprint(f"Reacted '{reaction}' on post")
--- a/src/cqc_lem/utilities/CLAUDE.md
+++ b/src/cqc_lem/utilities/CLAUDE.md
@@ -56,7 +56,10 @@ Two things follow for anyone writing a call site here:
    not once for the call site: the same lookup on the
    home feed, where the control does exist, is selector rot and must still warn. The other half of
    the test is whether a FALLBACK is in hand — `react_to_post_inline` warns on a missing
-   'Open reactions menu' only when it found no React toggle to default-Like instead (issue #873).
+   'Open reactions menu' only when it found no React toggle to default-Like instead (issue #873),
+   and on a missing post-click 'Reaction state' only when the pre-click read found that toggle: a
+   card that never carried one has nothing to re-read, so the miss IS the documented
+   trust-the-click fallback (issue #875).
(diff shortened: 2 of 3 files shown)


gitchrisqueen:
🔎 Claude adversarial review — PASS (0 code findings, 1 PR-body correction)

Verified against production, not just the diff. Pulled the 24h `cqc-lem` logs around the filed occurrence (2026-07-31 19:08:23Z) and reconstructed every reaction sequence:

```
18:57:30  Selector miss: Reaction state          <- pre-click MISS  => state is None
18:58:07  Selector miss: Open reactions menu
18:58:45  Selector miss: Reaction state (post-click)
```
That shape repeats for **every one of the 6 post-click occurrences** (17:19, 18:58, 19:03, 19:08, 19:12, 19:16, 19:21) — the pre-click read misses first in all of them, so `expected = state is not None` is `False` on exactly the cards that filed the issue. The fix lands on the real path, not a theoretical one. The alternative root cause I went looking for — pre-click read *succeeds*, card re-renders, post-click read goes stale (which would leave `expected=True` and fix nothing) — **does not occur in the data**.

Checks that passed:
- **Reachability.** The post-click read is only reachable when `trigger is not None` (`if trigger is None: return False` fires earlier), so the `state is None` + `expected=False` branch is genuinely exercised — confirmed by the ~37s pre-click→menu gap on these cards vs ~72s on the ones that also miss `React toggle` and bail with `Could not leave a reaction on post`.
- **No new blindness.** The pre-click `Reaction state` read still warns at full `MAX_WAIT_RETRY`, and it fires on every affected card — the SDUI-drift signal is retained in full, only the redundant second copy is downgraded.
- **Return semantics unchanged.** `warn_on_miss` never touches `find_first`'s return; `expected=True` implies the pre-click label read `no reaction` (the other branch returns `None` early), so the three-valued contract at `_engage_card` is untouched.
- **`max_try=1`** still burns one full `wait` timeout and only skips the 5s retry — matches the #872 precedent, and the pre-click read already proved absence with full retries.
- Label `Reaction state (post-click)` is unique in the tree — single call site, right place. Close is honest: #875 has no phased scope, both acceptance criteria met. Convention-compliant (`find_first`/`log_debug` escalation contract, no new `myprint`).
- `poetry run pytest tests/unit/app/test_comment_inline.py -q` → 29 passed locally; all required CI gates green; codecov/patch pass.

Fixed (no code change):
- **PR body claimed the pre-click `Reaction state` fingerprint "has not misfired"** — it has, on every affected card, and it is already filed as open issue #874. Corrected the body to say the signal is retained and cross-linked the family (#874, #877, #878, root cause #816) so the next fixer doesn't re-derive this.

gitchrisqueen:

Issue #883, opened 2026-08-01
fix(engagement): scope the own-post reply composer to its comment (same bug class as #876)


gitchrisqueen:
## Why
Found while fixing #876 — not filed as extra scope in that PR.

`post_comment_inline` used a **document-wide** `div[role='textbox']` lookup, so on a feed page that already had another composer mounted it typed into the wrong one (#876 fixed it by scoping to the post card).

`_reply_to_comment_inline` (`src/cqc_lem/app/run_automation.py`, used by `automate_reply_commenting` at the `_reply_to_comment_inline(...)` call site) still carries the un-scoped version: it clicks `button[aria-label='Reply']` scoped to `comment_el`, then looks the composer up across the whole document.

This is the exact failure `_reply_under_comment_inline` was written to fix — quoting its docstring:

> The bug: clicking a comment's Reply then taking the first page-wide role=textbox grabbed the post's main 'Add a comment' box, so the reply posted as a standalone comment. Fix: after opening the reply box, type into the composer NEAREST the comment.

So on a post with several comments, a reply can land as a top-level comment (or under the wrong comment) instead of under the one we are answering — and a composer scrolled off the top is what produces the `ElementClickInterceptedException ... at point (x, 9)` family.

Not folded into #876 because it is a different function on a different surface (replies to comments on the user's OWN posts) with no captured occurrence, and it needs a decision the post-card case did not: a reply composer is not reliably a DOM descendant of the comment element, which is why #478 picked the geometrically NEAREST visible composer instead of using containment.

## Scope
- Make `_reply_to_comment_inline` resolve the composer relative to the comment it is replying to, rather than page-wide.
- Prefer reusing `_reply_under_comment_inline`'s proven nearest-composer selection over inventing a third approach — collapse the two if they turn out to be the same operation.
- Do not change `_reply_under_comment_inline` or the #876 post-card scoping.

## Files
- `src/cqc_lem/app/run_automation.py`
- `tests/unit/app/test_reply_inline.py`

## Acceptance
- [x] A reply never types into a composer belonging to another comment or to the post's main comment box, even when those are earlier in document order.
- [x] A comment with no reachable reply composer skips (returns False) rather than borrowing another one.
- [x] Unit tests cover both, and fail without the fix (>=80% patch coverage).
- [ ] All required CI gates pass.

Follow-up of #876.

gitchrisqueen:
[...]

PR #884, opened 2026-08-01
fix(engagement): scope the inline comment composer to its own post card (closes #876)

### LINKED ISSUE #876: fix(errors): ElementClickInterceptedException: Message: element click intercepted: Element <div class="ql-editor ql (1x)
## Why
PostHog error tracking grouped this exception into an issue that is still active: **ElementClickInterceptedException**

> Message: element click intercepted: Element <div class="ql-editor ql-blank" data-gramm="false" contenteditable="true" data-placeholder="Add a comment…" aria-placeholder="Add a comment…" aria-label="Text editor for creating content" role="textbox" aria-mul

- First seen: `2026-07-31T18:47:29.866000Z` · last seen: `2026-07-31T18:47:29.866000Z`
- SDK: `posthog-python`

[Open the issue in PostHog](https://us.posthog.com/project/475262/error_tracking/019fb980-bff4-7983-8090-79ee5128dd13) — it has the stack trace, the grouped occurrences and the affected people.

## Scope

## Files
- `tests/unit/`

## Acceptance
- All required CI gates pass.

Auto-filed from PostHog Error Tracking. Dedup marker (do not remove): `posthog-issue-019fb980-bff4-7983-8090-79ee5128dd13`

gitchrisqueen:
## What

`post_comment_inline` now looks the SDUI comment composer up **scoped to the post card it was handed** (`parent_element=card`) instead of across the whole document.

## Why

PostHog grouped one `ElementClickInterceptedException` out of `auto_comment_in_groups`:

> `element click intercepted: Element <div class="ql-editor ql-blank" ... data-placeholder="Add a comment…"> is not clickable at point (890, 9). Other element would receive the click: <svg class="mercado-match">`

That reads like #815 (sticky nav steals a top-of-viewport click) — but it can't be, and the `y=9` is the tell. `_engage_card` scrolls the card it is about to comment on to the **middle** of the viewport immediately before calling `post_comment_inline`, so a composer belonging to that card cannot be at y=9. Something else was being clicked.

The feed walk — and the group walk in `auto_comment_in_groups`, which reuses it — comments on several posts **per page load** (`while posted < max_posts`, `continue` re-gathers from the same DOM), and LinkedIn leaves each composer mounted after it submits (`_composer_submitted` relies on exactly that: it checks the composer *cleared*). The composer lookup passed no `parent_element`, so `find_first` returned the first **visible** `div[role='textbox']` in document order — the *previous* post's composer, by then scrolled off the top and under the nav.

So this was never a positioning bug. Centering that composer, as #815 does, would not have rescued the run — it would have typed this post's comment into the **previous post** and submitted it there, since `_SUBMIT_NEAR_COMPOSER_JS` walks up from whichever composer it is given. The loud interception was the good outcome.

The same bug class was already found and fixed once on the reply path; `_reply_under_comment_inline`'s docstring records it verbatim ("taking the first page-wide role=textbox grabbed the post's main 'Add a comment' box", #478).

## The change

- `post_comment_inline` passes `parent_element=card` to the composer lookup. The locator chain is unchanged.
- **No document-wide fallback on purpose.** Commenting on the wrong post is worse than not commenting, and a miss is already graceful: `post_comment_inline` returns `False`, `_engage_card` releases the claim, a later run retries the post. A total miss still logs `Selector miss: Comment composer` with the tried selectors + URL, so if the composer ever renders outside the card that shows up as signal rather than silence.
- `CLAUDE.md`'s SDUI gotcha records the scoping rule beside the #815 centering rule.

## Testing

`tests/unit/app/test_comment_inline.py` — new `TestComposerIsScopedToTheCard`, driving the real `find_first` against a fake driver whose document holds an earlier post's composer ahead of this card's:

- an earlier post's composer is never typed into or clicked;
- a card with no composer of its own skips rather than borrowing one;
- the `parent_element=card` contract itself.

All three fail on `main` (verified by reverting the one-line scope and re-running). Full suite: `1789 passed` (`tests/unit/app`).

## Notes

- Not deployed-behaviour-affecting beyond feed/group commenting; no migration, no new env var, no API change.
- **Follow-up: #883** — `_reply_to_comment_inline` (replies to comments on the user's own posts) still carries the un-scoped, page-wide version of this lookup. Deliberately not folded in: different surface, no captured occurrence, and it needs a decision this one didn't (a reply composer isn't reliably a DOM descendant of its comment, which is why #478 used geometric nearest-composer selection there). Nothing remains on #876 itself.

Closes #876


gitchrisqueen:

- **CLAUDE.md size guard was red, and this PR made it worse (fixed in 6c6a0f4).** `main` is already **41,763** chars against the 40,000-char cap in `scripts/check_claude_md_size.py`; the guard only runs on PRs whose diff touches `CLAUDE.md`, so it went red here for a condition this PR did not cause — but the note added ~452 chars on top of it. The rule belongs in CLAUDE.md; the retelling (why #815's centering would not have rescued the run, how the feed walk re-gathers) is already in the `post_comment_inline` code comment. Trimmed to the invariant: +452 → **+286** chars. The overage itself can't be cleared by any PR scoped to this issue — filed **#885** for the repo-wide relocation pass, which also has to settle the workflow comment claiming this is a required check when branch protection does not list it.

Verified, no finding:

- **The diagnosis holds against the code, not just the narrative.** `_card_for_textbox` returns the nearest ancestor of the post's text box containing its `button[aria-label='Comment']`, so `card` spans the social bar and the comments section — the same element `react_to_post_inline` and the "Open comment composer" `click_first` already scope to successfully in prod. `_SUBMIT_NEAR_COMPOSER_JS` walks up from whichever composer it is handed, so the un-scoped lookup really would have submitted this post's comment on the previous post once centered; the interception was the lucky outcome, as the PR body claims.
- **The three new tests genuinely fail without the one-line change.** Re-ran them with `parent_element=card` reverted in the worktree: 3 failed / 29 deselected. With the fix: `32 passed`.
- **No document-wide fallback is the right call and fails loudly, not silently.** A miss returns `False`, `_engage_card` releases the claim so a later run retries the post, and `find_first(required=False)` emits `Selector miss: Comment composer` with the tried selectors + URL — which escalates on repeat if the composer ever stops being a descendant of the card. A stale `card` degrades the same way (skip + one warning) rather than borrowing a neighbour's composer.
- **The close is honest.** #876's acceptance criteria (root cause fixed, unit test on the failing path, required gates green) are all met by this diff. The one piece of deferred scope — `_reply_to_comment_inline`'s still-page-wide lookup — is filed as **#883** (`bug`, `priority:medium`, `agent:ready`) with quoted scope and real acceptance criteria, and linked in both the PR body and a comment on #876. Phase guard clean.
- Required checks are green (`Unit Tests (Python 3.12)`, `Integration Tests`, `GitGuardian Scan`, `UI Build`, `Migration Versions`); `codecov/patch` passes.

gitchrisqueen:

Issue #886, opened 2026-08-01
fix(engagement): harden the #478 reply-composer pick so it can never fall back to the post's main comment box


gitchrisqueen:
## Why
Found while fixing #883 — deliberately out of scope there ("Do not change `_reply_under_comment_inline`").

`_reply_under_comment_inline` (`src/cqc_lem/app/run_automation.py`, the #478 path: replies to replies under OUR comments on third-party posts) picks its composer with a SOFT penalty:

```js
let d = Math.abs(r.top - a.bottom) + (r.top < a.top ? 1e6 : 0);
if (d < bd) { bd = d; best = e; }
```

A composer ABOVE the comment is penalised, not rejected — so when the post's main "Add a comment" box is the only visible `role=textbox` on the page (our reply box never opened, the thread re-rendered, the box collapsed), it is still returned as `best` and the reply posts as a standalone top-level comment. That is exactly the failure the function's own docstring says it exists to prevent; the penalty only wins the race when a genuine reply composer is also present.

#883 fixed the sibling function (`_reply_to_comment_inline`) with a HARD filter plus an ownership check (`_reply_composer_for_comment` / `_in_same_comment`), and left this one alone as instructed. The two are the same operation, so the composer resolution should be shared rather than kept in two forms.

## Scope
- Point `_reply_under_comment_inline` at `_reply_composer_for_comment` (or hard-filter its own JS the same way): a composer above the comment is never borrowed, and no composer belonging to the comment means skip.
- Check the other call sites of `_reply_under_comment_inline` (the #478 sweep and the lead-signal reply at the `_reply_under_comment_inline(driver, wait, cont, text, ...)` site) still behave — a skip must stay a clean `False`, never an exception.
- If the two reply helpers end up identical after this, collapse them into one.

## Files
- `src/cqc_lem/app/run_automation.py`
- `tests/unit/app/test_comment_followups.py`
- `tests/unit/app/test_reply_inline.py`

## Acceptance
- [x] With the post's main comment box as the ONLY visible composer, `_reply_under_comment_inline` returns False instead of posting a standalone comment.
- [x] A composer belonging to a different comment is never typed into.
- [x] The two reply paths share ONE composer-resolution helper (or are collapsed into one function).
- [x] Unit tests cover the above and fail without the fix (>=80% patch coverage).
- [ ] All required CI gates pass.

Follow-up of #883.

Issue #885, opened 2026-08-01
chore(docs): CLAUDE.md is over the 40,000-char harness cap on main — move detail into docs/


gitchrisqueen:
## Why
`scripts/check_claude_md_size.py` caps `CLAUDE.md` at 40,000 chars, but **`main` is already 41,763 chars** — 1,763 over. The guard only runs on PRs whose diff touches `CLAUDE.md` (`.github/workflows/claude-md-size.yml` `paths:` filter), so the overage is invisible until someone edits the file — and then that PR's `CLAUDE.md ≤ 40,000 chars` check goes red for a condition it did not cause.

Found during the adversarial review of #884 (issue #876): that PR added ~450 chars of SDUI invariant and inherited a red check. The addition was trimmed to ~286 chars there, but no scoped PR can bring the file under the cap — that needs a repo-wide pass, which is why it is its own issue.

The check is NOT in branch protection's required contexts today (`Unit Tests (Python 3.12)`, `Integration Tests`, `GitGuardian Scan`, `UI Build`, `Migration Versions`), so it does not block merges — but the workflow's own comment claims it is a required gate, and every future CLAUDE.md edit ships red. Both halves are worth settling.

## Scope
- Bring `CLAUDE.md` under 40,000 chars by relocating DETAIL into `docs/*.md`, keeping CLAUDE.md as the map it says it is: locations, symbols, constants, invariants, and where to find the detail.
- Land it as ONE pass — the file is edited by nearly every agent PR, so a slow trickle just conflicts. Prefer the `Known Gotchas` / `Feature Areas` paragraphs that already carry a `docs/*.md` pointer: move the prose there and leave the one-line invariant + pointer.
- Do NOT drop any invariant. If a sentence states a rule, it stays (possibly shorter); only narrative/rationale moves.
- Decide the required-check question and make the repo consistent: either add `CLAUDE.md ≤ 40,000 chars` to branch protection's required contexts, or correct the "Required status check — gates merges" comment in `.github/workflows/claude-md-size.yml`.

## Files
- `CLAUDE.md`
- `docs/*.md` (destinations for the relocated detail)
- `.github/workflows/claude-md-size.yml` (comment accuracy only)

## Acceptance
- [x] `python3 scripts/check_claude_md_size.py` exits 0 on the branch.
- [x] Every invariant present in CLAUDE.md before the pass is still stated somewhere — in CLAUDE.md, or in the `docs/*.md` the surviving line points to.
[...]

PR #887, opened 2026-08-01
fix(engagement): scope the own-post reply composer to its own comment (closes #883)

### LINKED ISSUE #883: fix(engagement): scope the own-post reply composer to its comment (same bug class as #876)

gitchrisqueen:
## What

`_reply_to_comment_inline` (the own-post reply sweep, `automate_reply_commenting`) opened a comment's Reply box scoped to that comment, then looked the composer up across the **whole document**:

```python
composer = find_first(driver, wait, [... (By.CSS_SELECTOR, "div[role='textbox'])],
                      "Reply composer", visible_only=True, required=False, user_id=user_id)
```

`find_first` returns the FIRST visible match in DOM order, so the reply could be typed into:

- the post's main **"Add a comment"** box — the reply then posts as a standalone comment, answering nobody; or
- a composer left mounted by a comment we already replied to **earlier in the same sweep** — the reply lands under the wrong person. That composer is by then scrolled off the top, which is also the `ElementClickInterceptedException ... at point (x, 9)` family (#815).

Same bug class as #478 on the other reply path and #876 on the post card.

## How

New `_reply_composer_for_comment(driver, comment_el)` resolves the composer against the comment being answered, in two rules:

1. **Own subtree wins.** LinkedIn nests a comment's replies — and the reply box that opens at the end of them — inside the comment container, so a rendered `role=textbox` in that subtree is unambiguous. Among several, the one nearest the container's bottom edge is the one that just opened.
2. **Otherwise #478's geometry, hardened.** The visible composer nearest the comment's bottom edge, with anything **above** the comment rejected outright (an 8px slack for rounding only), and only after `_comment_container` walks the winner back up to this comment (`_in_same_comment`).

Both filters are hard — that is the one thing deliberately *not* reused from #478, whose `+1e6` penalty still returns the main comment box when it is the only candidate. Per the issue, `_reply_under_comment_inline` is untouched; its residual soft-penalty case is filed as **#886** (out of scope here, not lost).

No composer belonging to this comment means **skip** (`False`), never borrow one: the caller already writes the REPLY failure log row and the next sweep retries the comment. The skip logs at DEBUG on purpose — a warning per unanswerable comment would repeat and escalate working behaviour into a filed defect (#873/#875 lineage).

## Testing

`tests/unit/app/test_reply_inline.py` — 20 tests, all green, plus the full unit suite (8,934 passed).

Behavioural coverage of both acceptance criteria, verified to FAIL against pre-fix `run_automation.py`:

- `test_types_into_this_comments_composer_not_the_main_comment_box` — main box first in document order, this comment's box second; pre-fix the main box is typed into (`mine.send_keys` called 0 times).
- `test_comment_without_a_reachable_composer_skips_instead_of_borrowing_one` — only the main box on the page; pre-fix it returns `True` after posting into it.

Plus the resolver's paths: nested pick (end of the reply list), sibling-below pick, composer above rejected, composer owned by another comment rejected, owner-wrapper accepted, 0×0 (hidden) composers ignored, unrendered comment → `None`.

## Notes

- `CLAUDE.md` untouched: its SDUI composer-scoping gotcha already states the rule, and the file is over the 40,000-char harness cap on `main` (relocation tracked in #885), so any PR touching it runs that guard red.
- No migration, no new dependency, no behaviour change to `_reply_under_comment_inline` or the #876 post-card scoping.

Closes #883
Follow-up: #886

### DIFF
--- a/src/cqc_lem/app/run_automation.py
+++ b/src/cqc_lem/app/run_automation.py
@@ -1654,9 +1654,10 @@ def _reply_composer_for_comment(driver: WebDriver, comment_el: WebElement,
     Two rules, in order. A composer inside the comment's own subtree is unambiguous — LinkedIn nests
     a comment's replies, and the box that opens at the end of them, in the comment container. If this
     render puts it outside, fall back to #478's geometry: the visible composer NEAREST the comment's
-    bottom edge, with anything above the comment rejected outright, and only if it walks back up to
-    this comment. Both filters are HARD — #478 merely penalises composers above the comment, which
-    still hands back the main box when that is the only candidate. No match means skip.
+    bottom edge, with anything above the comment rejected OUTRIGHT — that hard above-filter is what
+    keeps the post's main box out, where #478 merely penalises it and still hands it back when it is
+    the only candidate — and with a box that resolves to a DIFFERENT comment rejected too. No
+    candidate means skip; we never borrow a composer.
     """
     anchor = _visible_rect(comment_el)
     if anchor is None:
@@ -1671,10 +1672,16 @@ def _reply_composer_for_comment(driver: WebDriver, comment_el: WebElement,
         return None
     if nested:
         return best[0]
-    # Sibling render: confirm ownership before typing. The nearest box below can still belong to a
-    # LATER comment when our own reply box never opened, and borrowing it answers the wrong person.
+    # Sibling render: reject a box that resolves to a DIFFERENT comment — the nearest box below can
+    # belong to a LATER comment when our own reply box never opened, and borrowing it answers the
+    # wrong person. An UNRESOLVED owner is not proof of that: `_comment_container` was written for a
+    # comment BODY (`expandable-text-box`) and rejects any ancestor holding a GIF/Emoji composer
+    # button, which is the composer's OWN toolbar here — requiring it to resolve would make this
+    # branch skip every time, silently, whenever LinkedIn renders the reply box outside the comment.
+    # Unresolved therefore falls through to #478's proven geometry, still under the hard above-filter
+    # that is what actually keeps the post's main comment box out.
     owner = _comment_container(driver, best[0])
-    if not _in_same_comment(driver, comment_el, owner):
+    if owner is not None and not _in_same_comment(driver, comment_el, owner):
         log_debug("Nearest reply composer belongs to another comment", action_type="reply", user_id=user_id)
         return None
     return best[0]
--- a/src/cqc_lem/app/run_automation.py
+++ b/src/cqc_lem/app/run_automation.py
@@ -1596,19 +1596,100 @@ def comment_on_feed_inline(driver, wait, my_profile: LinkedInProfile, user_id: i
     return posted
 
 
+# How far ABOVE a comment's own top edge a composer may still start and count as its reply box.
+# Only absorbs sub-pixel/rounding drift — a real reply box opens below the comment, never above it.
+_COMPOSER_ABOVE_SLACK_PX = 8
+
+
+def _visible_rect(element: WebElement) -> dict | None:
+    """Page-coordinate rect of a RENDERED element, else None. Zero-size IS the hidden case here —
+    the same width>0 && height>0 test #478 applies to composer candidates."""
+    try:
+        r = element.rect or {}
+    except Exception:
+        return None  # stale/detached element is not a candidate
+    if not r.get("width") or not r.get("height"):
+        return None
+    return r
+
+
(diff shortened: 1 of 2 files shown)


gitchrisqueen:

- **Silent dead fallback: `_comment_container` was written for a comment BODY, not a composer** (`src/cqc_lem/app/run_automation.py`). The sibling-render branch required `_comment_container(driver, composer)` to resolve back to this comment before typing. That helper walks up from `[data-testid='expandable-text-box']` and rejects any ancestor holding a `GIF` / `Repost` / `Emoji Picker` button — walking up from a *composer* that button is the composer's OWN toolbar, so every ancestor is rejected and it returns `null` for a perfectly valid reply box. Requiring proof of ownership from it makes the branch skip **every** sibling render, silently (DEBUG log + a REPLY `FAILURE` row), which is precisely the render the fallback exists to cover — #478's pure geometry works live there today. Fixed by keeping the rejection HARD only when the owner *resolves to a different comment*, and letting an **unresolved** owner fall through to #478's geometry under the unchanged hard above-the-comment filter (that filter, not the owner walk-up, is what keeps the post's main box out). Within one sweep the only open composers below a comment are its own — earlier comments are processed first, so their leftovers sit above the filter line — so the relaxation does not reopen #883. New test `test_unresolvable_owner_still_takes_the_box_below_this_comment`; the "belongs to another comment" rejection test still passes unchanged.

Verified, no change needed:
- Both acceptance criteria are covered by tests that genuinely fail on `HEAD~1` — re-ran `test_types_into_this_comments_composer_not_the_main_comment_box` and `test_comment_without_a_reachable_composer_skips_instead_of_borrowing_one` against the pre-fix `run_automation.py`: 2 failed.
- `element.rect` is page-coordinate on the W3C path and both the anchor and every candidate use it, so the geometry never mixes viewport and document space.
- Comment bodies are `[data-testid='expandable-text-box']`, not `div[role='textbox']`, so the nested lookup can't pick a comment's own text (which #478's live-validated geometry also confirms).
- `_comment_items_from_thread` returns the smallest ancestor of a Reply button carrying an author link, so a nested pick can never reach the post's main comment box.
- Sweep order is document order, so a comment's leftover composer is always ABOVE the next comment it could be stolen by.
- `WebDriver`/`WebElement` are imported; `log_debug` (not `log_warning`) on the expected no-op, per the #873/#875 escalation contract.
- The close is honest: #883's acceptance boxes 1–3 are ticked and the deliberately out-of-scope `_reply_under_comment_inline` remainder is filed and linked as #886.

`tests/unit/app` — 1804 passed after the fix.

gitchrisqueen:

PR #889, opened 2026-08-01
fix(engagement): hard-filter the #478 reply composer so the post's main comment box can never win (closes #886)

### LINKED ISSUE #886: fix(engagement): harden the #478 reply-composer pick so it can never fall back to the post's main comment box

gitchrisqueen:
## What

`_reply_under_comment_inline` (the #478 path — replies to replies under OUR comments on third-party posts) picked its composer with a **soft** penalty:

```js
let d = Math.abs(r.top - a.bottom) + (r.top < a.top ? 1e6 : 0);
```

A composer above the comment was penalised, not rejected. So when the post's main "Add a comment" box was the **only** visible `role=textbox` on the page — our reply box never opened, the thread re-rendered, the box collapsed — it was still returned as `best`, and the reply posted as a standalone top-level comment. That is exactly the failure the function's own docstring says it exists to prevent; the penalty only won the race when a genuine reply composer was also present.

## Why

#883 fixed the sibling function (`_reply_to_comment_inline`) with a hard above-filter plus an ownership check (`_reply_composer_for_comment` / `_in_same_comment`), and was told to leave this one alone. The two are the same operation, so keeping the resolution in two forms is how they drift.

## Changes

- `_reply_under_comment_inline` now resolves its composer via **`_reply_composer_for_comment`** — the ONE helper, shared with `_reply_to_comment_inline`. A box inside the comment wins; a box **above** it is rejected outright; a box that resolves to a **different** comment is rejected; no box of ours means skip. The ~10 lines of bespoke JS are gone.
- The identical type → focus → submit → verify tail of both reply paths moves into **`_type_and_submit_reply`**, so the submit/verify contract can't drift between them either.
- A missing composer is now a **DEBUG** skip (logged inside `_reply_composer_for_comment`) instead of a `log_warning` — an unopened reply box is an expected no-op, and per `CLAUDE.md` a repeated warning re-escalates as a defect (same class as #873/#875).

**The two helpers are deliberately NOT collapsed into one.** They now share everything except how they OPEN the reply box, and that step genuinely differs: the #478 thread path needs `scrollIntoView` + an `ActionChains` hover to render a hover-hidden Reply button before it can be clicked, where `_reply_to_comment_inline` uses `click_first`'s retry. Collapsing them would mean forcing one path's open-strategy onto the other.

## Call sites checked

Both callers treat the return as a boolean skip and are unaffected — the #478 sweep (`_followup_on_post_comment_replies`, gates `result["replied"]` / the ledger write on it) and the lead-signal delivery (`_reply_to_person_on_post`, returns it directly). A miss stays a clean `False`; every failure path is still inside the existing `try`.

## Testing

`8941 passed` (`pytest tests/unit -q`). New/updated coverage:

- `tests/unit/app/test_reply_inline.py::TestReplyUnderCommentComposerPick` — main comment box as the ONLY visible composer → `False` and never typed into; a later comment's composer → `False` and never typed into; our own box below the comment → typed and submitted; both reply paths call the same resolver anchored to their own comment; empty reply text is never typed.
- `tests/unit/app/test_comment_followups.py::TestReplyUnderComment` — updated for the shared resolver, plus a no-composer case asserting a clean `False` with **no** `log_warning`.

Verified the guard is real: reverting `run_automation.py` to `main` and re-running the new tests fails 5 of them, including `test_main_comment_box_as_the_only_composer_is_never_borrowed`.

Selenium DOM targeting itself is still validated on a supervised live run, as with #478/#876/#883.

Closes #886

### DIFF
--- a/docs/sdui-selenium-notes.md
+++ b/docs/sdui-selenium-notes.md
@@ -24,6 +24,25 @@ by the nav's `<svg>`: `ElementClickInterceptedException ... at point (x, 9)` (is
 still-mounted composer, which is the real y=9 source AND a comment landing on the wrong post.
 There is no fallback: no composer on this card = skip it (issue #876).
 
+## A reply composer is resolved ONE way, for both reply paths
+
+`_reply_composer_for_comment()` is the only thing that decides which `div[role='textbox']` a reply
+is typed into — `_reply_to_comment_inline` (a comment on a feed card, #883) and
+`_reply_under_comment_inline` (a reply under our own comment on someone else's post, #478/#886)
+both call it. Order: a box nested in the comment's own subtree wins; otherwise the visible box
+nearest the comment's bottom edge, with a box starting ABOVE the comment rejected **outright** and
+a box that resolves to a DIFFERENT comment rejected too. No box of ours = skip.
+
+The above-filter has to be a rejection, not a penalty. #478's original pick only scored an
+above-composer `+1e6`, so the post's main "Add a comment" box still won whenever it was the only
+visible composer (our reply box never opened, or the thread re-rendered and collapsed it) — and the
+reply posted as a standalone top-level comment, the exact failure the function exists to prevent
+(#886). The two helpers stay separate only because they OPEN the box differently: the #478 thread
+path needs `scrollIntoView` + an `ActionChains` hover to render a hover-hidden Reply button.
+
+A miss is an expected no-op and is logged DEBUG *inside* the resolver — never a `log_warning`,
+which would re-escalate as a defect on repeat (see `docs/error-tracking.md`).
+
 ## Failure naming
 
 Inline compose failures name the STEP that threw (e.g. `Inline comment post failed at focus
--- a/src/cqc_lem/app/run_automation.py
+++ b/src/cqc_lem/app/run_automation.py
@@ -1687,6 +1687,22 @@ def _reply_composer_for_comment(driver: WebDriver, comment_el: WebElement,
     return best[0]
 
 
+def _type_and_submit_reply(driver, composer: WebElement, reply_text: str, user_id: int = None) -> bool:
+    """Type into an ALREADY-resolved composer and submit (role=textbox + Ctrl+Enter fallback). Both
+    reply paths share this so the submit/verify contract can never drift between them. True only when
+    `_composer_submitted` confirms the post."""
+    reply_text = _strip_non_bmp(reply_text)
+    if not reply_text.strip():
+        return False
+    _focus_composer(driver, composer)  # sticky nav steals a top-of-viewport click (#815)
+    composer.send_keys(reply_text)
+    time.sleep(random.uniform(1, 2))
+    if not driver.execute_script(_SUBMIT_NEAR_COMPOSER_JS, composer):
+        composer.send_keys(Keys.CONTROL, Keys.RETURN)  # fallback
+    time.sleep(random.uniform(3, 5))
+    return _composer_submitted(driver, composer, reply_text)
+
+
 def _reply_to_comment_inline(driver, wait, comment_el, reply_text: str, user_id: int = None) -> bool:
     """Open a comment's inline reply box, type the reply, and submit (same SDUI pattern as
     post_comment_inline: role=textbox composer + Ctrl+Enter fallback). The composer is resolved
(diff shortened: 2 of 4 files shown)


gitchrisqueen:

Reviewed the diff against #886's acceptance criteria and the surrounding SDUI reply/composer code. The core fix is right: `_reply_under_comment_inline` now resolves through `_reply_composer_for_comment`, so the post's main "Add a comment" box is rejected outright instead of penalised. I independently verified the guard is real — reverting `src/cqc_lem/app/run_automation.py` to `origin/main` and re-running the new tests fails 5 of them, including `test_main_comment_box_as_the_only_composer_is_never_borrowed`. I also re-checked both call sites (`_followup_on_post_comment_replies`, `_reply_to_person_on_post`): a miss stays a clean `False` and the ledger write / `record_action` stay gated on it, so a failed reply is retried next sweep rather than recorded as sent. `_type_and_submit_reply` is a faithful extraction — step order, sleeps and the `_composer_submitted` verdict are byte-for-byte what both paths did before.

Findings fixed in b387b16:

- **Silent failure mode (real defect).** The diff dropped `_reply_under_comment_inline`'s own `log_warning` on a composer miss and delegated it with `# _reply_composer_for_comment logs it DEBUG` — but that helper's FIRST exit, `_visible_rect(comment_el) is None` (stale or unrendered comment), returned `None` with no log at any level. Both reply paths now depend on that helper for miss telemetry, so a stale comment mid-sweep dropped a generated reply with zero trace. Added the missing `log_debug` there; extended `test_none_when_the_comment_itself_is_not_rendered` to assert the miss is logged. The other two exits already logged, so the comment's claim is now true for all three.
- **Convention (CLAUDE.md, type hints required on all signatures).** `_type_and_submit_reply(driver, …)` left `driver` unannotated while its neighbour `_reply_composer_for_comment` annotates `driver: WebDriver`. Annotated.
- **Undocumented invariant.** This PR makes "ONE reply-composer resolver" a real invariant across both reply paths, and it was written down nowhere — `docs/sdui-selenium-notes.md` is the doc CLAUDE.md points at for the composer posture. Added a section stating the resolution order, why the above-filter must be a rejection and not a penalty (#478 → #886), why the two helpers stay separate (the hover-hidden Reply button on the thread path), and that a miss is DEBUG not WARNING.

Checked and NOT flagged: the `Closes #886` claim is honest — every acceptance box except "CI gates pass" is covered by the diff, and the "collapse them into one" criterion is conditional ("if the two helpers end up identical"), which they are not; the PR states why. Keeping the two helpers apart is the right call — forcing `click_first`'s retry onto the thread path would lose the scroll+hover that renders the hover-hidden Reply button. The `owner is None` fall-through in the sibling branch is #883's shipped behaviour, not new here.

`8941 passed` (`pytest tests/unit -q`) after the fixes.

gitchrisqueen:
🧩 phase-guard — heads up: issue #886 still has **boxes: 1 unchecked** acceptance box(es). Merging anyway (unticked boxes are common here), but if any of that scope did NOT ship, file a follow-up issue with `agent:ready` and link it, or drop the `Closes #886` from this PR.

gitchrisqueen:

PR #890, opened 2026-08-01
fix(errors): stop the React-toggle miss filing a third defect for one unreadable card (closes #877)

### LINKED ISSUE #877: fix(errors): RecurringWarning: Selector miss: React toggle (1x)
## Why

> Selector miss: React toggle

- First seen: `2026-07-31T18:43:39.976000Z` · last seen: `2026-07-31T18:43:39.976000Z`
- SDK: `posthog-python`

[Open the issue in PostHog](https://us.posthog.com/project/475262/error_tracking/019fb97d-3c12-7bf2-82d3-c95212566fcd) — it has the stack trace, the grouped occurrences and the affected people.

## Scope

## Files
- `tests/unit/`

## Acceptance
- All required CI gates pass.

Auto-filed from PostHog Error Tracking. Dedup marker (do not remove): `posthog-issue-019fb97d-3c12-7bf2-82d3-c95212566fcd`

gitchrisqueen:
## What

`react_to_post_inline` looks for the card's primary Like/React toggle **only when** the
`Reaction button state` button was absent (`trigger = state or find_first(...)`), and that toggle is
just ONE of the two ways into a reaction — the hover fly-out opener tried immediately after is the
other. So its absence alone is not a failure, and when *both* miss the condition is already
signalled twice:

1. the fly-out opener's own miss, which warns exactly for that case (`warn_on_miss=trigger is None`,
   added by #873), and
2. `_engage_card`'s `Could not leave a reaction on post` (tracked separately as #878).

The toggle lookup was still running at the default `warn_on_miss=True`, so a single unreadable card
in an `auto_comment_in_groups` run emitted a **third** warning for the same fault. `log_escalation`
re-emits a repeated warning at ERROR and captures a grouped `$exception`, which the daily
error→issue cron filed as this issue.

## Change

- `src/cqc_lem/app/run_automation.py` — the `React toggle` lookup passes `warn_on_miss=False`, so
  its miss logs at DEBUG. **One condition, one warning.**
- `src/cqc_lem/utilities/CLAUDE.md` — records that third rule beside the #873 / #875 ones, so the
  next call site in this function doesn't re-open the same escalation.

Nothing is silenced. The opener miss remains the single escalating signal for "this card's reaction
controls are unreadable", and it still fires on exactly the cards that produced this issue.

## Not in scope

Re-grounding the reaction anchors against the live DOM is **#816** (`risk:live-linkedin`,
`needs-human`), which explicitly says *do not guess* new selectors without a live capture. This PR
only stops a redundant warning escalating into a defect; it does not change which selectors are
tried or whether a reaction lands.

## Testing

- `tests/unit/app/test_comment_inline.py` — two new cases: the `React toggle` lookup is made with
  `warn_on_miss=False`, and it is never made at all when the state button is the trigger (proving
  the warning only ever fired on state-less cards).
- `pytest tests/unit/app -q` → **1812 passed**.

Closes #877

### DIFF
--- a/src/cqc_lem/app/run_automation.py
+++ b/src/cqc_lem/app/run_automation.py
@@ -970,12 +970,17 @@ def react_to_post_inline(driver, wait, card, post_content: str = None, comment_t
         reaction = choose_post_reaction(post_content, comment_text)
         # The reaction fly-out is hover-revealed off the card's primary Like/React toggle. Hover it
         # first (the menu opener is hidden until then), then click the opener.
+        # The toggle is only ONE of the two ways in: missing it still leaves the fly-out path below,
+        # so its absence alone is not a failure. And when the opener misses too, that miss already
+        # warns (`warn_on_miss=trigger is None`, issue #873) and the caller warns again — so warning
+        # here as well filed a THIRD PostHog defect for the same one condition (issue #877).
         trigger = state or find_first(
             driver, wait,
             [(By.CSS_SELECTOR, "button[aria-label='React Like']"),
              (By.CSS_SELECTOR, "button[aria-label^='React']"),
              (By.CSS_SELECTOR, "button[aria-label='Like']")],
-            "React toggle", parent_element=card, required=False, visible_only=True, user_id=user_id)
+            "React toggle", parent_element=card, required=False, visible_only=True,
+            warn_on_miss=False, user_id=user_id)
         if trigger is not None:
             try:
                 ActionChains(driver).move_to_element(trigger).perform()
@@ -985,7 +990,8 @@ def react_to_post_inline(driver, wait, card, post_content: str = None, comment_t
         # The fly-out opener is optional: with a `trigger` in hand its absence just means we take the
         # documented default-Like fallback below, which is working behaviour — warning about it every
         # card escalated into a filed defect (issue #873). With no trigger there is no fallback, so
-        # the card's reaction controls really are unreadable and the miss is worth the signal.
+        # the card's reaction controls really are unreadable and the miss is worth the signal — this
+        # is the ONE warning that stands for that condition (the React-toggle miss above is DEBUG).
         opened = click_first(driver, wait, [(By.CSS_SELECTOR, "button[aria-label='Open reactions menu']")],
                              "Open reactions menu", parent_element=card, required=False,
                              warn_on_miss=trigger is None, user_id=user_id)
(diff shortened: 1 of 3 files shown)


gitchrisqueen:

Reviewed the diff against `react_to_post_inline`, `selenium_util.find_first/click_first`, `logger.log_warning`'s escalation path, and the #873/#875 precedents this PR builds on.

- **Doc asserted an invariant the code does not yet deliver** (fixed in 1f50564): the new `utilities/CLAUDE.md` rule reads "one condition gets ONE warning … the opener's warning already stands", but one unreadable card still warns twice more — the pre-click `Selector miss: Reaction state` read at `run_automation.py:965` (default `warn_on_miss=True`, filed as **#874**, open) and the caller's `Could not leave a reaction on post` at `run_automation.py:1276` (**#878**, open). That file auto-loads, so the next agent would have read the reaction path as already compliant and left #874/#878 unfixed. The doc now names both as outstanding and says to collapse them. No code change — #874/#878 are separately tracked and correctly out of scope here.

Verified, no finding:
- **The premise holds.** `find_first` sends the miss to `log_debug` when `warn_on_miss=False` (`selenium_util.py:612`), and `log_debug` has no `log_escalation.note()` call — so this miss can no longer produce a grouped `$exception`. Sole call site: `run_automation.py:982` (grepped src/ and tests/).
- **Nothing is silenced.** Every path that previously depended on this warning still signals: toggle missing + opener missing → the opener warns (`warn_on_miss=trigger is None` stays True); toggle missing + opener opens + fly-out button missing → `React <reaction>` warns at its default; toggle missing but opener works → a landed reaction, correctly quiet.
- **Tests match behaviour.** Both new cases exercise the real branch (`args[3]` label matching is safe — all three `find_first` calls in the function pass the label positionally), and `test_missing_menu_still_warns_when_there_is_no_fallback` still pins the opener's `warn_on_miss=True`, so the #873 signal can't be silenced by regression. `tests/unit/app` → **1812 passed** locally.
- **The close is honest.** #877's three acceptance criteria (downgraded, unit-tested, CI green) are all met; no phase or unchecked box remains. #816 (live selector re-grounding, `risk:live-linkedin`/`needs-human`) is correctly left alone — this PR changes no selector and no reaction outcome.

gitchrisqueen:

gitchrisqueen:

PR #881, opened 2026-08-01
fix(errors): stop the un-reacted card's Reaction-state miss escalating as a defect (closes #874)

gitchrisqueen:
## What

`react_to_post_inline` reads the feed card's `button[aria-label^='Reaction button state']` to skip posts we already reacted to. That pre-click read ran with `required=False` and the default `warn_on_miss=True`, so its miss emitted `Selector miss: Reaction state` at WARNING; repeated across an `auto_comment_in_groups` run it re-emitted at ERROR via `log_escalation` and filed a PostHog `RecurringWarning` issue every day.

**Rationale corrected after review (`1A`).** The first version of this PR claimed LinkedIn only renders that button once a reaction is on the post, so the miss was an ordinary un-reacted card. That is wrong: #816 records this read and the post-click re-read missing at **near 1:1** (17 vs 16 in 48h), so the anchor is absent either way — it has **rotted** — and the function's own already-reacted branch tests for the label `no reaction`, which only makes sense if the button renders *before* we react.

The downgrade still stands, on the true reason: this read is a **duplicate symptom of the one rot #816 tracks**, and the `React toggle` lookup below it is its fallback, so the reaction attempt proceeds regardless. It is the same collapse #873 / #875 / #877 already made for the fly-out opener, the post-click confirm and the React toggle — **one condition gets ONE warning**.

## The change

- `src/cqc_lem/app/run_automation.py` — pre-click state read passes `warn_on_miss=False`; the comment says it is rot, cites **#816**, and names the warning that must stay.
- The reaction cluster's surviving signal is the **fly-out opener** miss (`warn_on_miss=trigger is None`, #873) — a test now pins it *beside* this downgrade, so the cluster cannot be quieted to zero one issue at a time while reactions are still broken.
- `src/cqc_lem/utilities/CLAUDE.md` — records the generalisable rule instead of the wrong anecdote: **quieting a miss claims it is not a DISTINCT actionable defect, never that the DOM is healthy** — cite the tracking issue at the call site and leave the fault's surviving warnings alone. Merged with the #878 paragraph that landed on `main` (see below) and states plainly how thin the cluster's remaining signal is now that it is fully collapsed.

No behaviour change beyond log level.

## Rebased onto current main (2026-08-01, second pass)

Three siblings merged while this PR was parked, so it was rebased onto each of them rather than reasserting its own version:

[...]

PR #892, opened 2026-08-01
fix(errors): stop the caller restating a reaction failure that already warned (closes #878)

### LINKED ISSUE #878: fix(errors): RecurringWarning: Could not leave a reaction on post (1x)

gitchrisqueen:
## What & why

PostHog filed **RecurringWarning: `Could not leave a reaction on post`** off an
`auto_comment_in_groups` run. The warning lives in `_engage_card`, one frame *above* the code that
detects the fault — it fires on every falsy-but-not-`None` return from `react_to_post_inline`, and
each of those paths had already warned where it happened:

| `react_to_post_inline` returns `False` because… | already warned by |
|---|---|
| no React toggle **and** the fly-out opener missed | the opener's `warn_on_miss=trigger is None` (#873) |
| the opener worked but the reaction button missed / wasn't clickable | that `click_first`'s own miss |
| anything raised | `Inline post reaction failed` |
| the click never registered (toggle still reads "no reaction") | *nothing* |

So one unreadable card emitted this as a **second** warning for a fault that was already signalled,
and `log_escalation` re-emitted it at ERROR and filed it as its own issue. This is the same shape as
#873 / #875 / #877, and `src/cqc_lem/utilities/CLAUDE.md` already named #878 as the next one to
collapse.

**The fix:** `_engage_card` logs the outcome at DEBUG. A reaction is best-effort and never blocks the
comment, so the caller has nothing to add that the detecting frame didn't already say.

**Nothing goes silent.** The one row above with no signal — a click the toggle never registered —
now warns exactly once, inside `react_to_post_inline`, where it is detected. Net warnings per
condition stay at one and the total strictly drops: an unreadable card goes from 2 warnings to 1.
The rule this generalises to, added to the tree's `CLAUDE.md` and `docs/error-tracking.md`: **warn
where you detect, not where you notice.**

`Selector miss: Reaction state` (#874) is the remaining duplicate on an unreadable card and stays
out of scope here; the doc still names it as open.

## Testing

- `test_a_click_that_never_registered_warns_exactly_once` — the moved warning fires, once, with a
  stable message (the escalation dedup key needs the template stable).
- `test_an_unreadable_card_adds_no_warning_of_its_own` — the `#873` path stays single-signalled.
- New `TestEngageCardReactionLogging`: a failed reaction is DEBUG and the comment still lands; an
  already-reacted post stays DEBUG; a landed reaction logs nothing at all.
- `pytest tests/unit/app -q` → **1817 passed** locally.

Closes #878

### DIFF
--- a/docs/error-tracking.md
+++ b/docs/error-tracking.md
@@ -105,6 +105,10 @@ return `False` both for "already reacted" (a no-op) and for genuine failure, so 
 working skip as `Could not leave a reaction on post`; it now returns `None` for the no-op — still
 falsy, so truthiness callers are unaffected — and only real failures warn.
 
+The second half of that pattern is **warn where you detect, not where you notice**. Restating a
+failure one frame up files a second defect for one fault, so that caller now logs the outcome at
+DEBUG and each failing path inside `react_to_post_inline` owns its single warning (issue #878).
+
 | Env | Default | Purpose |
 |---|---|---|
 | `LOG_ESCALATION_ENABLED` | `true` | master switch; false → zero Redis calls |
--- a/src/cqc_lem/app/run_automation.py
+++ b/src/cqc_lem/app/run_automation.py
@@ -1021,6 +1021,11 @@ def react_to_post_inline(driver, wait, card, post_content: str = None, comment_t
                            visible_only=True, warn_on_miss=expected,
                            max_try=MAX_WAIT_RETRY if expected else 1, user_id=user_id)
         if after is not None and "no reaction" in (after.get_attribute("aria-label") or "").lower():
+            # The card's controls were readable and the click STILL didn't take — the one reaction
+            # failure none of the selector misses above stand for, so it gets its single warning
+            # here, where it is detected. The caller's blanket warning is DEBUG (issue #878).
+            log_warning("Reaction did not register after clicking", user_id=user_id,
+                        action_type="comment")
             return False
         myprint(f"Reacted '{reaction}' on post")
         return True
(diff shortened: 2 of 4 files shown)


gitchrisqueen:

Parked from PR #881's revise tick. @gitchrisqueen answered `ok` to #881's Decision Comment, which included **2A: close #875 / #877 / #878 as duplicates of #816 and take them off the agent queue**, so that the reaction cluster keeps at least one signal while reactions are still not landing. #878 is now closed as a duplicate of #816 — but this PR was already open against it, so the merge call is yours rather than the runner's.

To be fair to this diff: it is **not** a pure quieting PR. It moves `Could not leave a reaction on post` from the caller to the point of detection and **adds** `Reaction did not register after clicking` inside `react_to_post_inline`. That is the right shape.

The gap is coverage, not intent. After #873/#875/#877 (+ #881, pending), a card where the `Reaction button state` anchor has rotted **but a React toggle is still found** produces: opener miss → DEBUG (a trigger exists), toggle click → post-click re-read → DEBUG + trusted (`expected = state is not None` is False), so the function returns `True`. With this PR the caller no longer warns either — so that path logs **nothing** and reports success, while #816 says reactions are not landing. That is the exact end state 2A was meant to avoid.

Reply one letter — or `ok` for the recommendation.

### 1. What happens to PR #892?
- **A. Hold it until #816 lands** — keep the caller's warning as the cluster's last blanket signal; re-open this PR (or re-file it) once the anchors are re-grounded and the misses are genuinely rare.  ✅ *recommended*
- **B. Merge as-is** — cleaner logging today; the rotted-anchor-with-toggle path becomes fully silent and reports success.
- **C. Merge with one addition** — keep the relocation, but also warn when the post-click re-read misses on a card whose pre-click read ALSO missed (i.e. the rot case), so the outage keeps exactly one signal. Slightly re-opens what #875 decided; I can implement it.

**My recommendation: `1A`.** The diff is good work and worth keeping, but it is the fourth collapse of one unfixed outage — #816 is the thing to spend the next tick on, and until it lands the noisy warning is doing its job.

/cc @gitchrisqueen

gitchrisqueen:
Ok

Issue #770, opened 2026-07-28
Affiliate program Part 2: LEM-generated promotional content with FTC disclosure


gitchrisqueen:
## Why this exists

Issue #737 scoped the affiliate/ambassador program into two independent pieces:

- **(A) Affiliate STATUS** — default-enrolled with opt-out, referral link, trial-time reward.
- **(B) Promotional CONTENT from the user's own LinkedIn account** — strictly opt-in, recorded consent, separate toggle.

PR #750 built all of (A) and the consent/gate plumbing for (B): `promo_content_allowed()`, `disclosure_report()`, `apply_disclosure()`, and the `post_to_linkedin` gate that blocks undisclosed affiliate content. It deliberately did **not** build a generator that actually produces LEM-promotional content, because the owner needed to confirm the reward/eligibility/FTC decisions first.

Those decisions are now finalized (#750 decision comment: **1A 2A 3A**). This issue tracks the remaining generator work so (B) becomes real rather than a consent gate with no upstream writer.

## Scope

- A new content generator (or extension of the existing post/comment/DM writers) that produces LEM-promotional material **only when** `promo_content_allowed()` is true.
- The generator must stamp the FTC disclosure automatically using `apply_disclosure()`; the existing `disclosure_report()` gate in `post_to_linkedin` remains the backstop.
- Integration with the unified content core (`content_framework.py`, `content_alignment.py`, `content_research.py`) — no parallel per-content-type prompt helpers.
- Respect all existing safety controls: human pacing, per-day caps, content-mix governor (promo slot), slop lint, and the 70/20/10 mix so affiliate content counts as the 10% promo class.
- PostHog events for content generated, published, and blocked.
- Unit + integration tests; ≥80% patch coverage.

## Acceptance

- [x] With (B) consent enabled, LEM can generate a post/comment/DM that promotes LEM and already carries the FTC disclosure.
- [x] With (B) disabled or missing consent, no promotional content is generated.
- [x] Content without a disclosure is still blocked at `post_to_linkedin` by the existing gate.
- [x] The generator respects the 10% promo mix cap and human-pacing daily budget.
- [x] Tests cover consent-off, consent-on, disclosure present, and disclosure-missing paths.

## Related

- #737 — original affiliate program issue
- #750 — Part 1 (status + consent gate + disclosure plumbing)
- #392 — first-comment link split (newsletter link mechanics apply to referral links too)
- #625 — slop lint
[...]

PR #750, opened 2026-07-27
feat(marketing): affiliate program Part 1 — default-enrolled status, opt-IN promo consent, FTC disclosure gate

gitchrisqueen:
Builds the permanent affiliate/ambassador program from #737 — **Part 1: affiliate STATUS and the (B) consent + disclosure gate**. This PR does not build the promotional-content generator; that is tracked in the follow-up issue below.

## The split

| | **(A) Affiliate STATUS** | **(B) Promotional CONTENT from the user's own account** |
|---|---|---|
| Default | **ON**, one-click opt-out | **OFF**, always |
| Turns on by | Signup | `POST /user/affiliate/promo-consent` with `consent_acknowledged` → records a timestamp + consent version |
| Gate | status column | `promo_content_allowed()` — the only reader |

`promo_content_allowed()` requires the program on, the user enrolled, the flag set **and** a consent timestamp. A flag with no timestamp is something other than the user saying yes, and it is refused. There is deliberately no env var for (B), and opting out of (A) withdraws it. (B) has consent plumbing and a publish gate but **no generator** — that work is in the follow-up issue.

## FTC 16 CFR §255

Extra trial time is compensation, so content publishing a referral link is a paid endorsement that must disclose the material connection **in the content**.

- `disclosure_report()` grades it deterministically.
- `post_to_linkedin` — the one publish choke point — flags an undisclosed affiliate post `error` instead of publishing it.
- It grades the body **and** the link #392 just carried into the first comment.
- A reworded disclosure (`#ad`, `#sponsored`, "I'm an affiliate…") counts.
- A blank `AFFILIATE_DISCLOSURE_TEXT` **blocks** affiliate content.
- The check fails **open** on an unexpected error.

## Rewards (decision finalized: 1A 2A 3A)

| Reward | Trigger | Default | Revoked on opt-out? |
|---|---|---|---|
| Enrollment bonus | Holding status | +7d | **Yes** |
| Referral bonus | Referred user **ACTIVATES** | +14d | **Never** |
| Cap | Ledger SUM per user | 90d | — |

See `docs/affiliate-program.md` §3.1 for the reasoning and rejected alternatives.

## Abuse guards, surfaces, testing

As described in the original PR body. `poetry run pytest tests/unit -q` → **7953 passed**. Integration → 131 passed (2 pre-existing Redis failures unrelated to this PR).

Contributes to #737 (Part 1 of 2).

Follow-up for the LEM-generated promotional-content generator: #770

[...]

Issue #658, opened 2026-07-26
PH11 — Marketing attribution: UTM discipline on all outbound + conversion goals + channel dashboard


gitchrisqueen:
## Context
The launch plan (docs/launch-and-marketing-plan.md, restored in PR #657) runs an organic acquisition engine — brand-account posts/comments/DMs, newsletter, video tutorials, FAQ — but nothing attributes a signup to a channel. #503 shipped funnel event instrumentation; what's missing is the SOURCE side. PostHog web analytics (free with posthog-js, PH1 #646) gives channels/UTM/conversion-goal reporting out of the box — but only if our links carry UTMs and the signup event fires with attribution intact. PostHog's paid-ads marketing analytics (beta) is explicitly OUT of scope until paid acquisition exists; the retired revenue dashboard (gone 2026-06-30) must not be used.

## Scope
- UTM discipline at the SOURCE: every outbound link LEM generates carries utm_source/medium/campaign — brand-account post CTAs and link_in_first_comment (utm_source=linkedin, campaign per post/archetype), newsletter edition links (campaign per edition), video tutorial descriptions (utm_source=youtube), FAQ/site links. Central helper (utilities/marketing/attribution.py: build_utm_url) so campaigns are consistent + testable; BRAND_SIGNUP_URL flows through it.
- Capture side: ensure posthog-js (PH1) initial-referrer/UTM person properties survive the signup flow; fire the signup/trial-start event where attribution context exists.
- Conversion goals in web analytics: signup + activation (first post published) as goals.
- Channel dashboard (extend PH5's provisioning script): signups + activations by utm_source/campaign, brand-post → site-visit → signup funnel.
- Referral groundwork: referral link format (utm_source=referral&ref=<user>) parsed into person properties — full referral program stays in the launch plan's own issue set.

## Acceptance
- Unit tests for build_utm_url + all call sites; a click on a brand CTA in staging shows channel attribution end-to-end in PostHog; dashboard renders from provisioned insights.

Issue #500, opened 2026-07-24
feat(onboarding): automated activation checklist + stalled-user nudges


gitchrisqueen:
## Why
Fully-automated onboarding/activation (docs/launch-and-marketing-plan.md (PR #495) Section A.3). Aha = first AI post published AND first automated comment/DM under the user's voice, within 7 days.

## Scope
- Server-persisted checklist state (connect LinkedIn → set voice/targeting → approve first post → enable caps → activated) surfaced in the SPA; emit each step to PostHog.
- New beat task `auto_onboarding_nudges` (daily, throttled like `auto_notify_missing_linkedin_session`) sending the next-best nudge via `utilities/linkedin/notification_email.py` + in-app banner (no session 24h, no voice 48h, no post 72h, trial T-3d).
- Nudge copy generated with `lem-simple`, brand-voice aligned.

## Files
`app/run_scheduler.py` (+beat entry in `app/my_celery.py`), `utilities/db.py`, `ui/src/`, unit tests.

## Acceptance
- Nudge selection logic unit-tested; activation event emitted; ≥80% patch coverage.

PR #903, opened 2026-08-02
feat(marketing): affiliate reward pays per activated referral, not for joining (closes #737)

### LINKED ISSUE #737: Affiliate/ambassador program: default-enrolled with opt-out, trial-time reward, FTC-compliant disclosure — permanent marketing arm
## Owner intent (2026-07-27)
LEM users become **ambassadors/affiliates of LEM by default**, notified with a clear opt-out. Affiliates receive **additional free-trial time**; opting out returns them to the standard trial (notified). This is **permanent** — not a launch-phase promo — so LEM always has a marketing arm. May be scoped to users who have a company page/profile if that is the cleaner boundary.

## ⚠️ Two things this issue MUST keep separate (owner review required before build)
There are two very different meanings of 'affiliate by default', and they carry different risk:

**(A) Affiliate STATUS by default** — the user is enrolled in the referral program, gets a referral link/code, and earns trial time for signups they drive. Default-on is normal and low-risk.

**(B) Automated promotional CONTENT from the user's own LinkedIn account** — LEM generating posts/comments/DMs that promote LEM, published as the user. **This must NEVER be default-on.** It is the user's professional identity and reputation; auto-posting marketing for our product from their account without explicit, specific, opt-IN consent is both a trust catastrophe and a legal problem.

**Recommendation: (A) default-on with opt-out; (B) strictly opt-in, separately, with its own consent screen and its own toggle.** Build them as two independent flags, never one.

## Compliance requirements (non-negotiable, build them in)
- **FTC endorsement rules (16 CFR Part 255)**: an affiliate receiving compensation (extra trial time IS compensation) who promotes LEM must **disclose the material connection** clearly and conspicuously in the promotional content itself. Any LEM-generated affiliate content must carry that disclosure automatically — it cannot be left to the user to remember.
- **Disclosure must survive generation**: the slop-linter/review gates must not strip it; add a check that affiliate-tagged content contains the disclosure before publish.
- **Clear notice + real opt-out**: the enrollment notice must state plainly what the user is enrolled in, the reward, and how to leave. Opting out must be one click and take effect immediately.
- **Framing of the trial change**: present the affiliate bonus as a **bonus on top of the standard trial**, and opt-out as returning to standard — never as a punishment. Do not construct a dark pattern where leaving feels like losing something they already owned. (Owner's phrasing 'their trial will be reduced' is fine as an accurate consequence, but the UX copy should read 'your trial returns to the standard N days'.)
- If restricted to users with company pages, say so in the notice and handle users who later add/remove a page.

## Scope
- **Data**: affiliate enrollment state (enrolled/opted-out/ineligible), referral code per user, referral attribution (who signed up from whose link — reuse the UTM/referral groundwork in #658 `utm_source=referral&ref=<user>`), reward ledger (trial days granted, why, when). Migration is timestamp-versioned.
- **Reward mechanics**: extend the existing trial machinery (`db.extend_trial_for_user`, `EARLY_ADOPTER_TRIAL_DAYS`, `early_adopter_slots`) rather than inventing a parallel path. **Research and RECOMMEND the reward size** (owner has not fixed a number): typical SaaS referral rewards, what is meaningful vs margin-destroying given LEM's per-user LLM cost (see `docs/cost-performance-margin-plan.md`), and whether the reward should be per-referral (e.g. +N days per converted signup, capped) or a flat enrollment bonus. Document the reasoning; the owner picks from options at merge (`risk:product-decision`).
- **Abuse guards**: self-referral detection, cap on total granted days per user, referral only counts on a real activated signup (not a raw click), fraud logging.
- **Notice + UX**: in-app notice on enrollment, an Affiliate section in Account settings (status, referral link, days earned, opt-out), and the separate opt-IN toggle for (B) auto-promotional content with its own explanatory copy.
- **Permanence**: no launch-phase gating on the program itself — it outlives P0/P1/P2. (Volume of any (B) content still respects phase caps + human pacing.)
- **Analytics**: PostHog events for enrolled/opted-out/referral-clicked/referral-converted/reward-granted so the marketing funnel is measurable (#650/#658 dashboards).

## Acceptance
- (A) and (B) are independently toggleable; (B) defaults OFF and cannot be enabled without explicit consent recorded with a timestamp.
- Affiliate-generated content always carries the FTC disclosure; a test proves content without it is blocked.
- Opt-out is immediate, reflected in trial end date, and the user is notified of the resulting trial length.
- Self-referral and duplicate-referral are rejected by test.
- Unit + integration tests; ≥80% patch coverage.

Related: #659 (owner activation checklist), #736 (brand account = user 1), #658 (UTM/referral attribution), #499 (early-adopter trial grants), `docs/launch-and-marketing-plan.md` §A.2.

gitchrisqueen:
Applies the owner's **2026-08-01 decision** on #737 and closes out the remaining scope.

#750 (Part 1, merged) shipped the program with **both** rewards — +7 for joining and +14 per activated referral. The later decision picked **per-referral, capped**, with the rationale that a flat enrollment bonus *"pays every enrollee whether or not they ever refer anyone"*. This PR makes the shipped policy match, and answers the decision's second half: *"research and propose N and the cap against per-user LLM cost and typical SaaS referral norms, with the reasoning documented."*

## What changed

| | Before (#750) | Now |
|---|---|---|
| Joining | +7 trial days, revoked on opt-out | **0** — `AFFILIATE_ENROLLMENT_BONUS_DAYS` kept as a knob, defaults off |
| Activated referral | +14 | +14 (unchanged) |
| Cap | 90 | 90 (unchanged) |

Three things had to move with it, because they were all hanging off a grant that no longer happens:

- **`affiliate_enrolled` fired only when the join bonus was paid.** It now fires on the call that *created* the enrollment row — `ensure_affiliate_enrollment` reports `created` (INSERT IGNORE `rowcount`), which is the only thing that counts each user once given every Account page load calls `enroll_user`.
- **The opt-out response reported a trial end date only when a reward moved.** `POST /user/affiliate/status` now reads it back off the user otherwise, so *"opt-out is immediate, reflected in the trial end date, and the user is notified of the resulting trial length"* holds without a grant.
- **The SPA copy hardcoded a join bonus** and would have rendered "0 bonus trial days" / "your trial returns to the standard 14 days" when leaving costs nothing. Copy is now written off `bonus_days`: at 0 it says *"you keep every trial day you have already earned"*. Both branches are tested — the loss-framing ban in the issue cuts both ways.

Nothing about (A)/(B) separation, consent, or the FTC gate is touched.

## Sizing (docs/affiliate-program.md §3.1, rewritten)

**Measured, not assumed.** PostHog `llm_call`, the 10 days to 2026-08-02 that carry per-user cost attribution: **mean ≈ $0.19, median ≈ $0.15, worst day $0.38** of LLM spend per active user per day (one heavily-active account, so read it as a fully-engaged trial day). So +14 days ≈ **$2.70**, and the 90-day cap ≈ **$17** of variable spend per user — well inside the margin plan's ≥70% blended / ≥60% per-user CM targets.

**Market:** B2B SaaS referral programs pay the referrer **on conversion** ($50–150 mid-market, or 100–150% of first-month value), give the referred account a trial extension or credit, rarely pay a flat join bonus, and commonly cap. Trial days are the cheapest currency a pre-revenue product has — COGS, no payout rail, no 1099.

The doc carries a levers table (7/60 conservative · **14/90 shipped** · 30/90 aggressive · uncapped rejected) so the numbers can move with one env change.

## Scope check on #737

Every acceptance criterion is now met — (A)/(B) independent with timestamped consent (#750), disclosure enforced with a test proving undisclosed content is blocked (#750), opt-out immediate + trial length reported (this PR), self-referral and duplicate referral rejected by test (#750), unit + integration coverage. Part 2 (the promo-content generator) shipped under #770, which is closed with all boxes ticked. The one scope bullet with no dedicated event, *referral-clicked*, is covered by #658's `referral` channel on the LEM Channels dashboard, as `docs/affiliate-program.md` §6 already records.

## Testing

- `tests/unit` → **9180 passed** (includes 3 new `enroll_user` event tests).
- Affiliate unit + integration → 116 passed. New integration coverage for the shipped policy: joining pays nothing and only referrals earn; the enrollment event fires once per user, not once per page load; opt-out takes nothing away but still reports the trial length. The existing 7-day-bonus fixture is kept deliberately — that path must keep working for any deployment that configures one.
- New `AffiliateNotice.test.tsx` (4 cases) + a zero-bonus case on `AffiliateCard.test.tsx`. UI tests run in CI (no local node_modules).

Closes #737

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -440,8 +440,9 @@ MARKETING_OWNED_DOMAINS=
 # the user, with recorded consent.
 AFFILIATE_PROGRAM_ENABLED=True
 AFFILIATE_DEFAULT_ENROLLED=True
-# Trial days for holding affiliate status. Revoked on opt-out (the user returns to the standard trial).
-AFFILIATE_ENROLLMENT_BONUS_DAYS=7
+# Trial days for merely holding affiliate status. 0 = the reward is per-referral only (the shipped
+# policy). A non-zero value pays every enrollee and is revoked on opt-out.
+AFFILIATE_ENROLLMENT_BONUS_DAYS=0
 # Trial days per referral that ACTIVATES (not per click, not per raw signup). Never clawed back.
 AFFILIATE_REFERRAL_BONUS_DAYS=14
 # Hard ceiling on total granted days per user.
--- a/docs/affiliate-program.md
+++ b/docs/affiliate-program.md
@@ -142,8 +142,8 @@ back off the user when no reward moved), so the user is told their trial length 
 ### 3.1 Sizing the reward — what N and the cap were measured against
 
 The owner fixed the SHAPE (per-referral, capped) and asked for N and the ceiling to be proposed
-against LEM's real per-user cost and SaaS norms. This is that reasoning; the shipped defaults are
-**+14 / 90**, and both remain one env change.
+against LEM's real per-user cost and SaaS norms. This is that reasoning; the options below were put
+to them on PR #903 and **+14 / 90 was picked (2026-08-02)**. Both remain one env change.
 
 **What a trial day actually costs (measured, not assumed).** PostHog `llm_call`, the 10 days to
 2026-08-02 that carry per-user cost attribution: **mean ≈ $0.19, median ≈ $0.15, worst day $0.38 of
--- a/docs/affiliate-program.md
+++ b/docs/affiliate-program.md
@@ -1,7 +1,8 @@
 # Affiliate / Ambassador Program
 
-**Issues:** #737 (status + consent + disclosure), #770 (the (B) writer) · **Status:** built, owner
-decision applied (1A 2A 3A) · **Created:** 2026-07-27 · **Updated:** 2026-07-31
+**Issues:** #737 (status + consent + disclosure), #770 (the (B) writer) · **Status:** built; reward
+policy set by the owner's 2026-08-01 decision (per-referral, capped) · **Created:** 2026-07-27 ·
+**Updated:** 2026-08-02
 
 LEM's marketing arm is its own users. Every account joins the referral program by default, gets a
 referral link, and earns extra free-trial time for people they bring in who actually get set up.
@@ -109,14 +110,23 @@ LEM to the author's connections, are both a different risk class and neither shi
 
 ## 3. Reward mechanics
 
-Two rewards, deliberately different kinds — and the difference is what makes the opt-out honest
-rather than a dark pattern:
+**The reward pays for outcomes, not for joining** (owner decision, 2026-08-01). Holding affiliate
+status earns nothing; a referral that ACTIVATES earns trial days, up to a per-user ceiling.
 
 | Reward | Trigger | Env | Revoked on opt-out? |
 |---|---|---|---|
-| **Enrollment bonus** | Holding affiliate status | `AFFILIATE_ENROLLMENT_BONUS_DAYS` (default **7**) | **Yes** — the user returns to the standard trial |
 | **Referral bonus** | A referred user **ACTIVATES** | `AFFILIATE_REFERRAL_BONUS_DAYS` (default **14**) | **Never** — it was earned |
 | Ceiling | Total granted days per user | `AFFILIATE_MAX_REWARD_DAYS` (default **90**) | — |
+| ~~Enrollment bonus~~ | Holding affiliate status | `AFFILIATE_ENROLLMENT_BONUS_DAYS` (default **0** = off) | Yes, when configured non-zero |
+
+The enrollment bonus is kept as a knob (a launch push may want a join incentive) but ships at **0**:
+a flat join bonus pays every enrollee whether or not they ever refer anyone. Everything the machinery
+does with it — the negative ledger row, the baseline floor, the "your trial returns to the standard
+N days" copy — still works, and is exercised by tests, for any deployment that sets it.
+
+At 0 the opt-out has no reward consequence at all: leaving takes nothing away, because nothing was
+given for joining. `POST /user/affiliate/status` still reports the resulting `trial_ends_at` (read
+back off the user when no reward moved), so the user is told their trial length either way.
 
 - A referral converts on **activation** (the #500 aha moment: a published post AND automated
   engagement), not on a click and not on a raw signup. A farm of dormant accounts pays nobody.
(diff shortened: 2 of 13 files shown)


gitchrisqueen:
Held (`needs-human`, risk: `product-decision`) because your 2026-08-01 decision said the reward numbers get picked at merge: *"`N` and the cap are **not fixed** — research and propose them… Bring options; final numbers get picked at merge."* The shape is already settled and not re-asked — **per-referral, capped, no join bonus** — this PR implements exactly that. Only the two numbers are open.

What they cost, measured (PostHog `llm_call`, 10 days to 2026-08-02): **$0.19 mean / $0.15 median LLM spend per active user per day**, so a granted trial day is ~$0.19 of variable spend. Market: B2B SaaS pays the referrer **on conversion**, $50–150 mid-market or 100–150% of first-month value, and rarely pays anything for merely joining. Full reasoning in `docs/affiliate-program.md` §3.1.

Reply one letter — e.g. `1A` — or `ok` to take the recommendation.

### 1. How many trial days per activated referral, and what per-user ceiling?
- **A. +14 per referral / 90-day cap** (what this PR ships) — legible as "another full trial", ~6 activations to cap, worst case ≈ **$17** of LLM spend per user  ✅ *recommended*
- **B. +7 / 60** — conservative: ≈ $11 worst case, but ~9 activations to reach the ceiling, so the reward reads weak next to a $50–150 cash norm
- **C. +30 / 90** — aggressive pull: 3 activations ≈ a free quarter, same $17 ceiling because the cap (not the per-referral size) is the real guardrail
- **D. +14 / uncapped** — strongest referrer incentive, but a referral chain becomes unbounded free service; the only option here with no worst case

**My recommendation: `1A`.** It matches the standard trial (so "another full trial" needs no explanation), pays only on the #500 activation event, and its worst case — ~$17 per user, ever — is a rounding error against the $50–150 the market pays in cash, while D is the one shape that can't be bounded.

Either way both numbers are single env values (`AFFILIATE_REFERRAL_BONUS_DAYS`, `AFFILIATE_MAX_REWARD_DAYS`) — no code change if you want to move them later. `AFFILIATE_ENROLLMENT_BONUS_DAYS` stays at **0** per your decision, kept as a knob only so a future launch push can turn a join incentive back on; the revoke/return-to-standard-trial machinery it needs is still tested.

gitchrisqueen:
Ok

gitchrisqueen:
Addressed your review.

**Your `Ok` = `1A`** — **+14 per activated referral, 90-day cap**. That is exactly what this PR already ships (`AFFILIATE_REFERRAL_BONUS_DAYS=14`, `AFFILIATE_MAX_REWARD_DAYS=90`), so no numbers moved. `docs/affiliate-program.md` §3.1 now records that 14/90 was *picked* on 2026-08-02 rather than proposed, so the doc and the decision don't drift apart later.

I also worked the automated review that landed on the same head. Two of its three blockers were real; one was not, and I've said which below rather than quietly skipping it.

**Fixed**

1. **The opt-out confirmation now reads the mutation result, not the query cache.** This is the finding worth having. `useSetAffiliateStatus` seeds the cache from the flip response *and then invalidates it* — so `reward_days` / `trial_ends_at` are correct for one render and gone after the refetch. `AffiliateCard` now reads `setStatus.data` directly, and `AffiliateState` no longer declares those two fields at all (`AffiliateStatusResult` does), which makes reading them off the query a compile error instead of a copy bug. Both branches are tested now, including the **#750 cohort**: a user holding the old +7 enrollment grant does get it revoked, `reward_days: 7`, and is correctly told their trial returns to the standard 14 days.
2. **The trial-date fallback is guarded on trial status.** `users.trial_ends_at` outlives the trial, so a paid or cancelled account opting out would have been told *"your trial still ends &lt;a date in the past&gt;."* It now only answers for a status where that date is live, reusing db.py's own list (promoted from `_EXTENDABLE_STATUSES` to `TRIAL_EXTENDABLE_STATUSES`). New integration test covers it.
3. **Type hints** on the new `_enroll` test helper.
4. **`created` documented as synthetic** on both `ensure_affiliate_enrollment` and `get_affiliate_enrollment`, including that the error path returns `None` so a caller can never read `created=False` from a write that didn't happen.
5. **The repeat-page-load test can no longer pass vacuously** — it now asserts the positive case in the same run, so if the patched import ever moves to module scope both halves fail together.
6. **`enroll_user`'s docstring states the semantics**: its event is once per *user*, and a re-join emits `affiliate_enrolled` from `set_status` with `source="opt_in"` — so the funnel does not actually under-count re-joins.
7. **Integration tests assert the exact date**, not `is not None`, plus reward-date precedence.
8. Nits: consistent apostrophes in the notice copy, `max(0, …)` on the enrollment-bonus env read, `queryByTestId` instead of `innerHTML === ''`.

**Not changed, with reasons**

- **"`trial_ends_at` has two producers with two shapes"** — it doesn't. Both branches go through the same `_utc_iso(...)` on the way out (`main.py`), so the wire format is identical either way. I added the exact-value assertion anyway, since that's what would catch it if someone later bypassed the serializer.
- **"`created` leaks into the API payload"** — it can't. `affiliate_state` builds its dict field by field off `get_affiliate_enrollment`, which never sets the key. Documented rather than restructured; changing the return to a tuple would touch call sites outside this PR's scope.
- **"`.env.example` / docs missing"** — both are in the diff, just past the 28k truncation the reviewer hit.
- I left `ensure_affiliate_enrollment(status: str = 'enrolled')` alone. Moving it to the `AffiliateStatus` enum is right, but it's a shared signature with callers outside this PR — worth doing, not here.

Tests: `tests/unit` **9180 passed**; affiliate unit + integration **64 passed**. UI tests run in CI (no local node_modules).

gitchrisqueen:

Reviewed as someone who did not write this. The reward-policy flip itself is right — `INSERT IGNORE` `rowcount` as the "this call enrolled them" signal is the correct primitive, the `AffiliateState` / `AffiliateStatusResult` split makes the cache-vs-mutation bug a compile error, and `_utc_iso` **is** applied to both `trial_ends_at` producers (I checked `main.py:3542`; the automated review's blocker #2 is not real). `_EXTENDABLE_STATUSES` has no surviving references, and `_Cursor.rowcount` already resets at the top of `execute()`.

What it missed, and what I changed:

- **The enrollment notice promised the grandfathered cohort a free exit, then charged them 7 days** (blocking). `AffiliateNotice.tsx` decided its opt-out line from `bonus_days` — config, now 0 for everyone — but `revoke_affiliate_enrollment_bonus` reads the **ledger** (`ENROLLMENT + REVOKED`), never the config. A user enrolled under the #750 defaults still holds a live +7. They were shown *"You keep every trial day you have already earned"* immediately before an irreversible click that takes a week off their trial: a factual misstatement, and the dark pattern #737 calls non-negotiable, inverted. `affiliate_state` now also reports `revocable_bonus_days` (this user's standing grant, same arithmetic the revoke does, computed from totals already read — no extra query), and the "how to leave" copy reads that while "what you get" keeps reading `bonus_days`. As a side effect the line is now correct for a user whose grant never landed because they were at the 90-day cap.
- **A flip that reports no date produced no confirmation at all.** `{flip && trialEnds && …}`: the new (correct) gate leaves `trial_ends_at` null for an account that is no longer on a trial, so those users saw the toggle move and nothing else — which is not "the user is notified" under the acceptance criterion, and it is this PR's own fix that opened the gap. Every flip now confirms; the dateless branch says what is true instead of quoting a stale date.
- **Type hints on the new `_iso` integration helper** (CLAUDE.md requires them on every signature, tests included).

Not changed, deliberately: `enroll_user(grant_bonus=False)` has **no call sites** (grepped), so the "quiet backfill now spams the funnel" concern is hypothetical; `created` cannot reach the API payload because `affiliate_state` builds field by field; and I left `ensure_affiliate_enrollment`'s string `status` default alone — moving it to `AffiliateStatus` is right but touches callers outside this PR.

Close is honest: every #737 acceptance criterion is covered across #750 and this PR, Part 2 shipped and closed under #770, and no later phase is left in prose.

Tests: `tests/unit` **9184 passed**; affiliate integration **20 passed**. New coverage for the grandfathered cohort (config 0 + ledger 7 → "returns to the standard 14 days", and the flip really does revoke 7), earned referral days never reading as revocable, an already-revoked bonus netting to zero, a missing ledger, and the dateless confirmation. Vitest fixtures carry the new field; UI tests run in CI.

gitchrisqueen:

Issue #905, opened 2026-08-02
feat(security): mandatory strong-factor enrolment + extension-session narrowing — Phase 2c.1 (follow-up of #897)


gitchrisqueen:
## Why

#897 shipped Phase 2c (passkeys, TOTP, recovery codes, step-up gate). Two pieces named in
`docs/AUTH_SECURITY_DESIGN.md` were deliberately left out of it, and this issue exists so they are
not lost in prose — the failure mode the runbook's phase guard exists for.

### 1. Mandatory enrolment (design §7 Stage 2, the second half)

> After a grace period (`REQUIRE_STRONG_FACTOR_AFTER`, env-controlled date), email-PIN alone stops
> being sufficient for accounts that have a strong factor enrolled, and **enrollment becomes
> mandatory for the rest at next login**.

2c shipped the first clause: an account that HAS a factor can no longer sign in on a PIN alone
(`auth_factors.has_strong_factor` → pending handle). It did **not** ship the second: an account with
no factor still signs in on a PIN and still passes the step-up gate, which is the honest limit
documented in `docs/strong-authentication.md` ("Still open after 2c").

What is needed:
- `REQUIRE_STRONG_FACTOR_AFTER` (env date, empty = never) read at call site.
- After that date, a PIN login for a factor-less account lands in a **forced enrolment** state
  rather than a normal session — it may enrol a passkey/TOTP and save recovery codes, and nothing
  else, until it does.
- A dismissible pre-deadline prompt in the SPA so nobody meets the deadline cold.
- Nobody is ever locked out: the PIN remains a valid bootstrap TO enrolment, it just stops being a
  standalone key.

### 2. Narrow the `extension`-scoped session

`POST /api/user/extension-token` mints `sessions.scope='extension'` behind a step-up, and that scope
satisfies the step-up gate at exactly one opted-in call site (`/api/user/linkedin-cookie`). But the
token is otherwise an ordinary full session — it can read and write everything a browser session
can, exactly as in 2b. It should be restricted to the endpoints the extension actually calls.

## Acceptance

- [x] `REQUIRE_STRONG_FACTOR_AFTER` gates a factor-less PIN login into forced enrolment; empty date
      keeps today's behaviour exactly.
- [x] Pre-deadline dismissible prompt in the SPA; a user who enrols never sees it again.
- [x] No account can be locked out — the PIN still bootstraps enrolment after the deadline.
[...]

PR #906, opened 2026-08-02
feat(security): passkeys + TOTP + recovery codes + step-up gate — Phase 2c (closes #897)

### LINKED ISSUE #897: feat(security): passkeys + TOTP + recovery codes + step-up gate — Phase 2c (follow-up of #745)
## Why

Phase 2 of #745 is built in three PRs (`docs/AUTH_SECURITY_DESIGN.md` §8):

| Phase | Scope | Status |
|---|---|---|
| **2a** | Encryption at rest — `utilities/crypto.py`, column migration, dual-mode read, backfill, key rotation, cookie-only `li_at` | ✅ shipped in #807 |
| **2b** | `public_uid`, email-as-attribute, hashed session tokens, httpOnly cookie, per-device sessions, auth rate limiting, `auth_audit_log` | ✅ PR #869 |
| **2c** | **This issue** | ⬜ open |

This issue exists so that merging #869 cannot silently drop 2c. **That is exactly how #568 was lost** — it closed on a docs-only PR and Phase 2 was never filed.

## Scope

- **Passkeys** — WebAuthn registration + assertion (`py_webauthn`), primary factor.
- **TOTP** — `pyotp` alternate factor, QR enrollment.
- **Recovery codes** — single-use, hashed with `argon2-cffi`.
- **Step-up gate** — re-auth required on credential-touching endpoints (LinkedIn credential changes, session revocation, email change).
- **Enrollment UI** in the account Security surface (alongside the `SecurityCard` shipped in 2b).
- **Email-PIN demoted to bootstrap-only** once a stronger factor is enrolled.

## Prerequisite

Verify the public hostname serves a **valid TLS certificate** before starting — WebAuthn requires a secure context and will not register against a bad cert. `https://lem.christopherqueenconsulting.com` currently resolves and serves 200, so this looks satisfied; confirm the cert chain explicitly as step one rather than assuming.

## Acceptance

- [x] Passkeys primary + TOTP alternate + recovery codes.
- [x] Step-up enforced on every credential-touching endpoint.
- [x] Email-PIN is bootstrap-only once a stronger factor exists.
- [x] No plaintext LinkedIn credential or session token remains in the schema (final check across 2a/2b/2c).
- [x] Security-focused tests at ≥90% patch coverage + threat-model note in the PR body.
- [x] New migrations use TIMESTAMP versions.

Follow-up of #745. Design: `docs/AUTH_SECURITY_DESIGN.md` (§0 decisions, §8 build order). Original: #568.

gitchrisqueen:
Phase **2c** of #745 — the last of the three sequential PRs in `docs/AUTH_SECURITY_DESIGN.md` §8
(2a encryption at rest → 2b identity/sessions → **2c strong factors**). The one-line summary: the
email PIN stops being a key to the account and becomes a bootstrap.

Operator + reviewer doc: **`docs/strong-authentication.md`**.

## Prerequisite — TLS, verified before any of this was written

The issue gates 2c on the public hostname serving a valid certificate, because WebAuthn silently
requires a secure context. Checked explicitly rather than assumed:

```
subject=CN = christopherqueenconsulting.com
issuer=C = US, O = Google Trust Services, CN = WE1
notAfter=Oct  4 01:13:30 2026 GMT
X509v3 Subject Alternative Name: DNS:christopherqueenconsulting.com, DNS:*.christopherqueenconsulting.com
Verify return code: 0 (ok)
```

The wildcard covers `lem.`, so RP id `lem.christopherqueenconsulting.com` is registrable.
`webauthn_util.relying_party()` re-checks the shape at call time and returns 503 (passkey UI hidden)
instead of offering a button that cannot work.

## Threat-model note

| Threat (design §2) | Before 2c | After |
|---|---|---|
| **T2 — mailbox compromised** | attacker signs in and drives the victim's LinkedIn | a PIN alone no longer signs in an account with a factor; a passkey or TOTP is still owed |
| **T3 — real-time phishing / PIN relay** | same as T2, no mailbox breach needed | **passkeys are origin-bound**, so a proxy page cannot use one. TOTP is still relayable — it defeats T2, not T3, and the docs say so |
| **T4 — XSS / stolen session** | a stolen session could write `li_at`, move the email, revoke devices | those writes need a factor proved in the last 5 minutes (`sessions.last_verified_at`) |
| **T1/T6 — DB dump** | 2a covered `li_at`/tokens/password | 2c adds: TOTP seeds are `lemv1:` envelopes, recovery codes are argon2id hashes — a dump yields no working authenticator |
| **T5 — root on the VPS** | out of scope | still out of scope, deliberately (§5.3) |

Two replay guards worth calling out. `user_auth_factors.sign_count` is a monotonic counter for
**both** factor kinds — a passkey's WebAuthn signature count and TOTP's last accepted 30-second time
step — and each must strictly increase. That is what makes a **cloned authenticator** and a
**re-typed TOTP code** fail rather than pass. TOTP is valid across a ±1 step drift window (up to 90s,
ample for a proxy to relay it twice); it cannot be made phishing-resistant, but it is now single-use.
`auth_challenges` lives in **MySQL, not Redis** for the same reason: Redis state here is allowed to
fail open, and a challenge store that failed open would let a replayed assertion through.

## What's in it

- **`utilities/auth_factors.py`** — the ONE place factor state and the step-up verdict are decided.
- **`utilities/webauthn_util.py`** — ceremonies. RP id + expected origin are **derived** from
  `PUBLIC_BASE_URL`; letting those drift from the app's hostname is how a user base gets locked out.
- **Login.** Passkey (username-less, so `/auth/passkey/login/begin` is not an account-existence
  oracle) is the only phishing-resistant path and the only one minting an already-stepped-up
  session. A PIN on an account with a factor returns a pending handle — **including on the
  no-mail-provider bypass**, which skips the PIN entirely and would otherwise have been a hole
  straight through 2c. `email_verified_at` is still stamped: the mailbox proved the address.
- **Step-up gate** on `li_at`, the LinkedIn password, email change init, session revocation,
  extension-token minting, factor removal and recovery regeneration. Refusal is **403
  `step_up_required`**, never 401 — the SPA's axios interceptor reads any 401 as a dead session and
  would have logged the user out instead of asking them anything.
- **Adding a factor is never gated; removing one always is.** A recovery code exists so you can
  enrol a new passkey after losing the old one — gating enrolment behind a factor you no longer hold
  is a lockout, not a control. A recovery-code login signs you in but does **not** step up, so a
  found sheet of codes is not a LinkedIn session.
- **The extension caveat the design told us to decide in 2c (§6.5).** `/api/user/linkedin-cookie` is
  bearer-exempt and the extension holds only a session token; it can never run WebAuthn. Its step-up
  happens ONCE, in the SPA, when `/user/extension-token` mints a `scope='extension'` session. That
  scope satisfies the gate at **exactly one opted-in call site** — a blanket exemption would have let
  a stolen extension token move the email address and revoke every device.
- **SPA**: `AuthFactorsCard` enrolment surface, `useStepUp` (wraps a call, opens the modal on the
  403, re-runs the original write), passkey + second-factor login steps, new audit-event labels.
- **Migration `V20260802020439__strong_auth_phase_2c.sql`** (TIMESTAMP version) — `user_auth_factors`,
  `user_recovery_codes`, `auth_challenges`, `sessions.last_verified_at` / `scope`. Purely additive,
  nothing backfilled: every account starts with zero factors and logs in exactly as it does today.

New deps, licenses as recorded in design §6: `webauthn` (py_webauthn, BSD-3-Clause), `pyotp` (MIT),
`argon2-cffi` (MIT). No service, no third party, $0.

## Rollback

`STRONG_AUTH_ENABLED=false` returns every account to its 2b behaviour — PIN is a full login, no
step-up gate — **without deleting a single enrolled factor**. A rollback is a flag, not a migration.

## Testing

- `tests/unit/utilities/test_auth_factors.py` (37) — policy: the demotion, which methods are offered,
  TOTP replay, recovery-code single-use and the race, all six ways the gate can be answered.
- `tests/unit/utilities/test_webauthn_util.py` (21) — RP derivation, non-secure origin refused,
  rejected ceremonies return None.
- `tests/unit/api/test_strong_auth.py` (59) — the API seam: PIN demotion on both login paths, passkey
  login, recovery-code login not stepping up, the 403 shape, every gated endpoint, 401 on every new
  endpoint without a session, and the failure paths that must not fall back to a session.
- `tests/integration/test_strong_auth.py` (19) — live MySQL: envelopes and argon2 hashes at rest,
  challenges spent once, plus the **final at-rest sweep across 2a/2b/2c** — it drives every secret
  through its real write path and asserts no raw column holds the plaintext it was given.
- SPA: `AuthFactorsCard.test.tsx` (7), `LoginModal.test.tsx` (6) — including the guard asking for a
  factor and then retrying the same write, and cancelling leaving it undone.

Local: **9,293 Python unit tests pass**, **334 SPA tests pass**, `tsc -b` and the production build
clean. Coverage on the new modules: `auth_factors.py` **96%**, `webauthn_util.py` **99%** (the issue
asks for ≥90%). Integration tests skip without a migrated MySQL and run in CI.

## Scope check

Every acceptance box on #897 is covered: passkeys + TOTP + recovery codes; step-up on the
credential-touching endpoints; PIN bootstrap-only once a stronger factor exists; the final
no-plaintext sweep; ≥90% patch coverage on the new security code + this threat-model note; TIMESTAMP
migration.

Two things named in the DESIGN (not in #897's acceptance) are deliberately not here, and are filed
rather than left in prose:

**Follow-up: #905** — mandatory enrolment (`REQUIRE_STRONG_FACTOR_AFTER`, design §7 Stage 2's second
clause) and narrowing the `extension`-scoped session to the extension's own surface. Until then an
account with **no** factor still passes the step-up gate: the rollout is opt-in first, there is
nothing to prove with, and gating those accounts would brick the cookie paste and the email change
for everyone who has not enrolled. That limit is stated in `docs/strong-authentication.md`.

Closes #897


gitchrisqueen:
Held (`needs-human`, risk: security). This PR changes how people sign in and adds a gate in front of every write that touches your LinkedIn credentials — three calls are yours, not mine.

### 1. An account with **no** strong factor still passes the step-up gate. Ship it that way?
The rollout is opt-in first (design §7 Stage 2): the gate only bites once you have enrolled something. Today nobody has, including you.
- **A. Ship as-is** — nothing changes for anyone until they enrol; the gate becomes real the moment they do; mandatory enrolment lands on #905  ✅ *recommended*
- **B. Gate factor-less accounts too** — every LinkedIn-cookie paste, email change and device revoke starts failing for everyone until they enrol a factor, with no warning
- **C. Ship with the gate off** (`STRONG_AUTH_ENABLED=false` in prod) — enrolment UI visible, nothing enforced, flip it on when you've enrolled

### 2. The browser extension's token satisfies the gate at `/api/user/linkedin-cookie`. Right call?
The extension holds only a session token and can never run a passkey ceremony. I made its step-up happen **once**, in the SPA, when `/user/extension-token` mints a `scope='extension'` session — and that scope is opted into at exactly one endpoint, so a stolen extension token still cannot move your email or revoke your devices.
- **A. Keep it** — one-click reconnect keeps working; narrowing the scope everywhere else is tracked on #905  ✅ *recommended*
- **B. Gate it too** — the extension's one-click reconnect breaks; every cookie refresh becomes a manual paste in the SPA after a passkey prompt
- **C. Drop the scope, exempt the endpoint entirely** — simplest, but then ANY stolen session can write a `li_at`, which is the thing this PR exists to stop

### 3. Step-up freshness window is **5 minutes** (design §6.5's number). Keep it?
This is how long after touching your passkey you can keep changing credentials before being asked again.
- **A. 5 minutes** — matches the approved design  ✅ *recommended*
- **B. 15 minutes** — fewer prompts while you set an account up; a stolen session has a wider window
- **C. 60 minutes** — barely noticeable, and materially weaker: it is roughly "once per sitting"

**My recommendation: `1A 2A 3A`.** Every one of them is the approved design as written, and 1A is the only option that does not risk locking you out of your own LinkedIn reconnect on the day this deploys — the moment you enrol a passkey on the account page, the gate protects you for real, and `STRONG_AUTH_ENABLED=false` is a one-line rollback that keeps every enrolled factor.

Two things worth knowing before you decide:
- **After this merges, enrol a passkey before anything else.** Once you have one, an emailed code alone will no longer sign you in — that is the point — and your recovery codes are shown exactly once, so save them on the same visit.
- The TLS prerequisite is satisfied (`*.christopherqueenconsulting.com`, Google Trust Services, valid to 2026-10-04, chain verifies), so passkeys will actually register against the live hostname.

gitchrisqueen:
Ok

gitchrisqueen:
Addressed your review. You replied **`Ok`**, so all three recommendations are taken: **1A** ship with factor-less accounts passing the gate (opt-in rollout, mandatory enrolment on #905), **2A** keep the `extension`-scoped session satisfying the gate at the cookie endpoint only, **3A** step-up freshness stays **5 minutes**. No design change was needed for any of them — that is what the PR already does.

What I did change is the set of real defects the adversarial review found *around* those decisions:

**The two-stage login's seams**
- **A correct PIN now clears the rate limiter before handing over to the second factor.** Both stages count out of the same per-email/per-IP buckets, so a user who mistyped the PIN a few times arrived at the passkey/TOTP prompt already throttled — a self-inflicted lockout on the one path with no way around it. Fixed on the PIN path and the no-mail-provider bypass.
- **A verified step-up whose session stamp fails is now a 500, not a 200.** Returning success without the stamp meant the next credential write asked again, forever, with nothing logged anywhere. Enrolment still stamps best-effort — the factor is already stored there, so a missed stamp costs one extra prompt, not the enrolment.
- **`STEP_UP_DENIED` now carries ip/user-agent.** It is the audit row that most often means someone else is holding the session; without the client on it there was nothing to chase.

**Smaller, real**
- A duplicate passkey returns the same generic 400 as a failed verification. Credential ids are globally unique, so `"That passkey is already registered"` told the caller a passkey they hold is enrolled on somebody else's account.
- `/auth/passkey/login/begin` is rate limited per client IP — unauthenticated, and it writes a challenge row per call.
- Step-up limiting keys on `user-<id>` when an account has no email row; the limiter skips a blank identity, so `""` meant no per-identity limit at all there.
- Factor removal audits the kind removed and whether a factor remains.
- Dropped the dead `attribution` field from `PasskeyLoginBeginRequest` — an account holding a passkey enrolled it while signed in, so that path can never be a signup and nothing downstream read it.

**Docs** (`docs/strong-authentication.md`): the pending handle is spent by the *attempt*, not the success — one wrong code ends the sign-in and you start from the email PIN. That IS the brute-force bound on the second stage (exactly one guess per handle, so the 6-digit space is never walked); it is a real UX cliff and is now stated rather than discovered. Also added the operator path for a user who loses every factor **and** the recovery sheet.

**Deliberately not changed, with reasons** — silence would read as done:
- **`Optional[Request]` on a route parameter is a FastAPI error**, not a style improvement: it is not a valid Pydantic field type and every affected endpoint 500s at import. I tried it, the suite caught it, `http_request: Request = None` stays. The non-route helper `_require_step_up` does use `Optional[Request]`.
- **Unconfirmed TOTP already cannot gate a login.** `count_auth_factors` filters `confirmed_at IS NOT NULL`, and `list_auth_factors` defaults to `confirmed_only=True`, so the abandoned-setup lockout is not reachable. `upsert_totp_factor` replaces the pending row rather than appending.
- **The `li_at` endpoint was already gated** (`_require_step_up(..., extension_scope_ok=True)`), so the T4 row of the threat table is accurate.
- **Signup attribution was never dropped.** An account with a strong factor enrolled it while signed in, so `is_new_user` is false on both demoted branches — the funnel call they skip could not have fired.
- **Expired challenges already self-purge** — `create_auth_challenge` deletes expired rows on every write.
- **Enums for the challenge/method constants** and a `NamedTuple` for `count_recovery_codes` are genuine polish, but they touch every call site of a security path the day it ships. Not worth the churn inside this PR; better as a follow-up if you want it.

Testing: **9,301 unit tests pass** (+8 new, in `TestTwoStageLoginSeams` — limiter clearing on both demotion paths, the 500-not-silent-loop, the duplicate-passkey oracle, per-IP limiting on passkey begin, the emailless bucket, and both enriched audit rows). One pre-existing test needed a `factor_summary` patch because factor removal now reads the kind before deleting.

gitchrisqueen:

Reviewed the full diff (not the truncated one the bot saw — `auth_factors.py`, `webauthn_util.py`, the `db.py` additions, the migration and all five test files are present and were read). The architecture holds up: 403-not-401, the username-less login ceremony, gating email change on `init` rather than `verify`, the atomic `consume_auth_challenge`, TOTP replay via the shared counter, and refusing recovery codes at step-up are all correct and defended at the call site. Three real defects at the seams, all fixed in `cf713dde`.

- **Enrolling a factor was a free step-up (T4 does not hold as claimed).** `passkey_register_complete` / `totp_enroll_confirm` stamp `sessions.last_verified_at` — correctly, the ceremony *is* a proof of possession — but enrolment itself was ungated, so `XSS → enrol my own passkey → stamped → read the li_at` walked straight through the gate standing next to it. Added `auth_factors.enrollment_allowed()`: the **first** factor stays free (nothing to prove with, and an account that could never enrol a first one could never enrol at all), every one after it needs step-up. This creates **no lockout** — every way to sign in to an account that already holds a factor either arrives stepped up (passkey login, PIN+TOTP) or is a recovery-code session, which is now marked `sessions.scope='recovery'` and let through by name (design §6.8). That session enrols but is **not** stamped for it, so a found recovery sheet still has to run one visible, audited step-up with the new factor before it reaches a credential. `AuthFactorsCard`'s enrolment mutations now go through `guard()` like every other write — without that a user adding a second passkey just saw "could not be registered".

- **One mistyped digit ended the sign-in, and the SPA looped on it forever.** `/auth/second-factor/verify` consumed the pending handle *before* looking at the code, and `LoginModal` kept the same `pending_token` on failure — so every retry after one typo returned "That sign-in expired" with no way out but closing the modal. The handle now survives a wrong code and is burned by the `SECOND_FACTOR_MAX_ATTEMPTS`-th (default 5), counted in `auth_challenges.attempts` **in the same UPDATE that claims it**, so two concurrent guesses cannot both be the last. **This reverses a trade the docs argued for**, deliberately: the durable counter is a *harder* bound than the one it replaces, because the Redis limiter in front of it fails open, and 5 tries out of 10⁶ is not an oracle. `SECOND_FACTOR_MAX_ATTEMPTS=1` restores the old behaviour exactly if you disagree. 401 now means "wrong code, sign-in still alive" and 400 means "handle gone, start over" — that split is what lets the SPA stop dead-ending.

- **A second authenticator app could be enrolled.** `upsert_totp_factor` deletes only UNCONFIRMED rows, so begin+confirm on an account that already had one left **two** confirmed TOTP factors — both counting towards `has_strong_factor` and showing on the Security card, only the newer seed ever verified, and removing the newer one silently reactivating the older secret. `begin_totp_enrollment` now refuses while one is confirmed; the endpoint says so with a 400 rather than a 500.

Not findings, checked and cleared: the signup-attribution concern from the earlier bot review (an account with a strong factor is never `is_new_user`, so nothing is dropped); `session_token` as a query param on `GET /user/auth-factors` (the SPA sends the `'cookie'` sentinel, no secret in the URL); `consume_auth_challenge` atomicity; sign-count regression (py_webauthn raises, and the counter is written back); challenge handles and `pending_token` stored as SHA-256; expired-row sweep on the write path.

Close is honest: every acceptance box on #897 is covered by the diff and I ticked them on the issue; the two design items that are not in this PR are tracked on **#905** and linked in the body.

Tests: 9,319 Python unit tests pass locally; new unit coverage for the enrolment gate (all four call sites, the recovery allowance, the no-stamp), the attempt counter and the TOTP duplicate; integration coverage for `auth_challenges.attempts` and the recovery scope against real MySQL; two SPA tests for retry-vs-restart. `tsc -b` clean. Docs + the `CLAUDE.md` invariant updated — the old text stated the reverse rule.

gitchrisqueen:

Second adversarial pass, on the head the last one produced (`abb3e4d8`) rather than the code that pass reviewed. Read the full diff again — `auth_factors.py`, `webauthn_util.py`, the `db.py` additions, the migration, the SPA and all five test files — plus the surrounding call sites in `main.py`. Both findings are at the same seam: the per-handle attempt counter that the last round introduced.

- **The second factor's guessing budget was resettable, so the durable bound the docs lean on was not a bound on the account.** `auth_challenges.attempts` bounds ONE pending login, and the stage in front of it hands out a fresh handle — and a fresh counter — for free. Both ways to reach that stage are inside 2c's own threat model: the no-mail-provider bypass proves nothing at all, and a compromised mailbox (**T2**, the threat 2c exists to defeat) mints PINs all day. Five guesses per round with unlimited rounds walks a 6-digit TOTP space. Worse, the bypass branch also called `clear_auth_limits(email, ip)` — on a branch where **no PIN is ever typed**, so the comment justifying it ("a user's own earlier typos") describes something that cannot happen there, and the only counters it forgave were an attacker's own; one unauthenticated request reset every limiter in front of the second factor. Now: `count_challenge_attempts` sums the account's spend over `SECOND_FACTOR_ATTEMPT_WINDOW_MINUTES` (15), `_begin_second_factor` refuses with **429** once it is gone and otherwise seeds the new handle with what was already spent, and an unreadable count fails **closed** — a database that will not answer is not an empty budget. A correct code clears it outright, so one typo does not follow a user into their next sign-in. The bypass branch no longer clears anything; the PIN path still does, because a proof precedes it there and the second factor's own budget is deliberately not one of the counters it resets.

- **Recovery codes were withheld in exactly the case they exist for.** `available_methods` appended `recovery_code` only `if methods` — i.e. only when something else was already offerable. A deployment that loses its secure public origin (`PUBLIC_BASE_URL` changed, TLS lapsed, `WebAuthnUnavailable`) drops `passkey` from the list, so an account whose only factor is a passkey got a demoted PIN and **an empty method list** — a total lockout, on the one path the sheet of codes is supposed to cover. Keyed on the account HOLDING a factor now, so the codes survive the passkey disappearing; an account with no factor still gets nothing (codes are not a factor, and that account never reaches the second stage). The SPA's authenticator/recovery toggle only renders when there are two methods to toggle between — otherwise a recovery-only user could switch themselves to a TOTP they never enrolled.

Not findings, checked against the code rather than the truncated diff the bot reviews: a passkey-only user is **not** locked out of a PIN-started login (`LoginModal` runs the standalone username-less ceremony when `methods` contains `passkey` — `SecondFactorVerifyRequest` never needed a credential field); `revoke-all` **is** step-up gated (same call site as single-session revoke, `all_others` is a branch inside it); `li_at` and the LinkedIn password are gated, the former with `extension_scope_ok=True`; `pending_token` cannot be spent at `/auth/step-up/verify` and vice versa (purpose-scoped claims); `session_token` on `GET /user/auth-factors` carries the `'cookie'` sentinel, not a secret; `create_auth_challenge` sweeps expired rows on every write; `count_challenge_attempts` compares against `created_at` with the connection pinned to UTC (`db.py` connect config).

Close is honest: all six acceptance boxes on #897 are covered by the diff and ticked on the issue, and the two design items that are not in this PR are on **#905** (open, `agent:ready`) and linked in the body.

Tests: **9,326 Python unit tests pass** locally (+7 for the carried budget, the 429 on both demotion paths, the fail-closed count, clear-on-success and the bypass no longer clearing; +1 for the recovery-code lockout), plus integration coverage of the per-account count and its purpose/owner scoping against real MySQL. `tsc -b` clean. Docs, `.env.example` and the `CLAUDE.md` invariant updated — the previous text described the per-handle bound as the durable one.

gitchrisqueen:

PR #907, opened 2026-08-02
fix(ops): /health/deep must require a CONSUMING worker, not just a registered one

gitchrisqueen:
Closes the gap I flagged after the v0.120.0 deploy.

## The gap

Mid-deploy, `/health/deep` returned:

```json
{"status": "healthy", "workers": 5, "lanes": {"celery@a": [], "celery@b": [], ...}}
```

Five workers, **every lane list empty**, reported as healthy. `maint begin` cancels every queue consumer, so the whole tier stays registered and answering the control channel while doing nothing.

It resolved on its own here — but a maintenance mode that never lifts would look healthy indefinitely, which is the same class of blind spot the endpoint was built to close.

**Registration was never the question a monitor is asking.** No consumer means no task will run, which is the same outage as no worker at all — it just looks better.

## Change

- `status` keys on a new **`consuming`** count (workers subscribed to ≥1 queue) rather than how many answered. Zero consuming → `degraded`, covering both "no workers" (the v0.118.0 shape) and "workers idle" (stuck maintenance).
- **One live consumer is enough.** A deploy recreates lanes one at a time; failing on a single idle lane would flap the monitor through every rollout — that's how an alert gets muted, and a muted alert is worse than no alert.
- **`maintenance`** reported alongside, so a degraded reading is legible: during a deploy it's the expected cause, outside one it's the thing to go clear. Read *outside* the control-channel `try` and defaulting to `None`, so unreadable Redis can never downgrade an answer we already trust.

## No action needed on the existing monitor

The literal `"status":"healthy"` is **unchanged**, so the configured UptimeRobot keyword monitor keeps working untouched.

A test now pins that exact string. Renaming the field or value would silently disarm every configured monitor — and a monitor that can no longer match looks exactly like one that is passing.

## Verification

Against the live stack on the v0.120.0 image:

| | |
|---|---|
| real stack | `status=healthy workers=5 consuming=5` |
| simulated idle workers | `status=degraded workers=2 consuming=0` |

Full unit suite green — **9182 passed**.

Not labelled `release:now`: this hardens a monitor for a state that isn't currently occurring, and `docs/release-fast-lane.md` reserves the lane for user-visible breaks. It rides the next window.

### DIFF
--- a/docs/stack-watchdog.md
+++ b/docs/stack-watchdog.md
@@ -84,14 +84,37 @@ broker's control channel (`_inspect().active_queues()`), so a lane whose contain
 simply is not in the reply.
 
 ```json
-{"status": "healthy", "workers": 5, "lanes": {"celery@selenium": ["se_engage"]}}
+{"status": "healthy", "workers": 5, "consuming": 5, "maintenance": false,
+ "lanes": {"celery@selenium": ["se_engage"]}}
 ```
+
+| Field | Meaning |
+|---|---|
+| `status` | `healthy` / `degraded` / `unknown` — the only field a monitor should assert on |
+| `workers` | workers that answered the control channel — **presence, not usefulness** |
+| `consuming` | workers subscribed to ≥1 queue. **This is the one that decides `status`.** |
+| `maintenance` | `true`/`false`, or `null` when Redis couldn't be read. Explains a `degraded`. |
 
 | `status` | Meaning |
 |---|---|
-| `healthy` | at least one worker is consuming |
-| `degraded` | broker reachable, **nobody consuming** — the exact v0.118.0 shape |
+| `healthy` | at least one worker is **consuming a queue** |
+| `degraded` | broker reachable, **nothing consuming** — either no workers (the v0.118.0 shape) or workers registered but idle |
 | `unknown` | control channel unreachable. **Unmeasured is never `healthy`.** |
+
+### Why `consuming`, not `workers`
+
+A worker being *present* is not the same as a worker *working*. `maint begin` cancels every queue
+consumer, so a stuck maintenance mode leaves the whole tier registered, answering, and doing
+nothing. That state was observed live during the v0.120.0 deploy and the endpoint called it
+`healthy` — workers: 5, every lane list empty.
+
+Registration was never the question a monitor is asking. No consumer means no task will run, which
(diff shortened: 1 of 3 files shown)


gitchrisqueen:

- **Every deploy would have reported `degraded`, firing the monitor this PR documents on all four release windows a day.** The "one live consumer is enough" tolerance covers lanes recreated one at a time, but `maint begin` cancels *every* lane's consumer at once — `scripts/deploy.sh` runs begin → drain (up to `DRAIN_TIMEOUT=180`s) → recreate workers → end on every release, and `consuming == 0` across that whole span. The documented UptimeRobot rule (alert when `"status":"healthy"` is absent) would have alerted on healthy deploys, which is the alert-muting the PR body itself argues against. Fixed: while the maintenance flag is set, `status` stays `healthy` — `consuming: 0` and `maintenance: true` are still in the body, so nothing is hidden from a human. The suppression is bounded by the flag's **own TTL** (deploy sets 1800s, `maint end` deletes it), so the state this endpoint exists for is still caught: a deploy that dies between begin and end leaves consumers cancelled, the flag expires, and the reading goes `degraded`. `unknown` is never suppressed, and unreadable Redis (`None`) never suppresses — a window we cannot confirm is not a window. Same posture as layer 1's `WATCHDOG_GRACE_SECONDS`.
- **Unit tests reached Redis.** The new tests called the real `is_maintenance_mode()` (→ `_redis_client()` → a live connect attempt), so the result depended on whether a broker happened to be absent in the runner. Fixed with an autouse fixture defaulting the flag to `False`; tests that care patch it explicitly.

Tests: rewrote `test_maintenance_is_reported_so_degraded_is_legible` (it pinned the behaviour above) into `test_declared_maintenance_window_is_not_an_outage`, and added `test_maintenance_that_never_lifts_still_degrades`, `test_maintenance_never_upgrades_an_unknown_reading`, `test_unreadable_maintenance_flag_never_suppresses_a_degraded_reading`. `docs/stack-watchdog.md` updated to match. `tests/unit/api` + maintenance + worker-shutdown suites green (854 passed).

Checked and clean: no `Closes #N` to verify, logger/type/enum conventions hold, `/health` stays trivial, the `"status":"healthy"` literal is unchanged and still pinned, and the endpoint exposes nothing it did not already.

gitchrisqueen:

PR #909, opened 2026-08-02
feat(newsletter): cover image upload + opt-in AI generation (closes #893)

### LINKED ISSUE #893: feat(newsletter): Add cover image upload and AI generation for newsletters
## Why
User requests the ability to upload custom cover images for LinkedIn newsletters and an option to auto-generate images based on engagement best practices via account settings.

Auto-filed from in-app feedback #23 by the feedback→auto-work loop (`docs/launch-and-marketing-plan.md` §B.2/B.3). Classifier: feature/medium, risk `product-decision`, confidence 1.00.

[Watch the session replay](https://us.posthog.com/project/475262/replay/019fb9d7-b96b-798a-8a5b-442a56c321d4) — the browser session this report was filed from.

## Scope
- Address the reported feature in the `newsletter` area of the product.
- Risk `product-decision` — see the Decision Comment below before implementing.

## Files
- `tests/unit/`

## Acceptance
- All required CI gates pass.
- A human has answered the Decision Comment before this is merged.

gitchrisqueen:
## What & why

A LinkedIn newsletter edition had no way to carry a cover image. The reporter asked for two things,
and the owner's [decision comment](https://github.com/christopherqueenconsulting/linkedin_engagement_manager/issues/893) signed off on **both**: upload your own artwork, AND an
opt-in "generate one for me" surfaced in account settings.

The two halves are deliberately **not symmetric**:

| Source | Where it comes from | Lands as |
|---|---|---|
| `upload` | The author picks their own artwork in the newsletter review queue | `approved` — it's theirs, it publishes |
| `ai` | Per-edition "Generate with AI", or `cover_image_auto` on a fresh draft | `pending_review` — waits for the author |

A cover is a **public brand asset**, so nothing generated ever publishes unreviewed.
`_approved_cover_path` (run_automation) is the ONE thing that decides a cover may reach LinkedIn —
it reads `cover_image_status` and nothing else.

## How

- **`utilities/newsletter_cover.py`** (new) — the ONE place a cover is validated, stored, and
  generated. Both sources pass the same deterministic gate first: decodable image, PNG/JPG/WEBP,
  ≤8 MB, ≥640×336, landscape-ish (LinkedIn renders covers at 1.91:1). A failing upload is a 400
  with the reason; a failing *generation* is never written to the row, so a truncated render leaves
  the edition exactly as it was. Approval is the human half layered on that.
- **Generation reuses the SHARED image path**, not a parallel per-content-type helper:
  `get_flux_image_prompt_from_ai` (which already encodes the engagement best practices a cover
  needs — one focal subject, strong foreground separation, **no text/logos/charts**) →
  `generate_post_image(..., ratio="16:9", depicts_person=False)`. 16:9 is the closest supported
  ratio to 1.91:1; `depicts_person=False` keeps the avatar LoRA out, since a cover is the
  newsletter's brand asset rather than a scene the author appears in.
- **Cost**: opt-in twice over. `cover_image_auto` is off by default, and even with it on the
  result still waits for approval. Nothing generates a cover at publish time.
- **Publish**: `attach_article_cover` writes the path into the article editor's hidden
  `<input type=file>` (clicking the styled button would open an OS file chooser Selenium can't
  drive), with a lazy-render fallback via the "Add a cover image" affordance and a best-effort
  crop-dialog confirm. `STEP_COVER` is deliberately **not** a graded editor step — the cover is
  optional, so grading it would report `MISSING` on every cover-less publish, and a cover that
  won't attach is a warning, never a `failed_step`.
- **Storage**: `cover_image_path` is relative to `assets_dir`, so the same value is the disk
  location and the `/api/assets?file_name=` value. `/api/assets` is public, hence a random filename
  suffix; `cover_abs_path` re-resolves through `realpath` and re-checks containment so a
  hand-edited row can't hand the publish flow an arbitrary file. The API returns `cover_image_url`
  and drops the path — the browser never sees a filesystem path.
- **Migration** `V20260802032347__add_newsletter_cover_image.sql`: three nullable columns on
  `newsletter_editions` + `cover_image_auto` on `newsletter_settings`. Additive only.
- **SPA**: cover panel in `NewsletterQueue.tsx` (preview, status badge, upload / generate /
  approve / remove; the generation poll is bounded so a failed generation can't disable the panel
  forever) and the opt-in toggle in `NewsletterCard.tsx`.

Full posture: `docs/newsletter-covers.md`; one-paragraph invariant added to `CLAUDE.md`.

## One deviation worth naming

The owner's note said "reuse the existing `lem-image` path and the carousel-imagery machinery."
`lem-image` is the DALL·E-3 alias, but the app's actual post/carousel image machinery is
`generate_post_image` → Flux via Replicate (that's what `carousel_creator.select_slide_image` uses).
I reused **that** path, so covers match the visual style of everything else LEM renders and inherit
its avatar guardrails and C2PA handling. Say the word if you'd rather covers go through DALL·E — it's
a one-line swap in `generate_cover_for_edition`.

## Testing

- `tests/unit/utilities/test_newsletter_cover.py` (new, 28) — the gate from both sides (accepts
  PNG/JPG/WEBP; rejects empty, oversized, undecodable, too-small, portrait, panorama, wrong format),
  save/resolve/remove round trips, traversal refusal, and every generation failure mode including
  "generation worked but the render failed the gate → nothing stored".
- `tests/unit/utilities/linkedin/test_article_editor.py` — attach writes the path into the hidden
  input, clicks the trigger when the input renders lazily, confirms a crop dialog, and is `False`
  (never an exception) on a missing file / missing control / WebDriver error. Plus: a cover that
  can't attach **still publishes**, and no cover path never touches the upload control.
- `tests/unit/app/test_newsletter_publish.py` — `pending_review` is never published, `approved`
  resolves and is threaded to the editor, a vanished file warns and publishes without it; the
  generation task lands `pending_review`, writes nothing on failure, and survives a profile that
  won't load; the `cover_image_auto` queue hook fires once per new draft and never when off.
- `tests/unit/utilities/test_db_newsletter.py` — the new DB helpers (owner-scoped writes, status
  only on an edition that has a cover, clear nulls all three, DB errors are `False`), and both
  reads select the cover columns.
- `tests/unit/api/test_newsletter_api.py` + `tests/integration/test_newsletter_cover_api.py` (new)
  — the endpoints, and the full lifecycle against a stateful editions store + a real temp assets
  dir: upload → approved on the queue, re-upload deletes the old file, rejected image leaves no
  cover, generated cover is pending until approved, remove takes the file, and every cover route is
  a 404 (spending nothing, writing nothing) for another user's edition.

`poetry run pytest tests/unit -q` → **9259 passed** locally; the 231 newsletter/article-editor/cover
tests re-run green. SPA `tsc -b` is clean (Node 18 here is too old to run `vite build`/vitest — CI
covers those).

The attach selectors are **not** live-validated against LinkedIn (no live session available
headless), same posture as the rest of the article-editor ladder — routes are keyed on
aria-label / name / accept / text, never class names, and a miss degrades to publishing without
the cover.

Closes #893


gitchrisqueen:

Reviewed the full diff against the issue's acceptance criteria and the code it seams into (`resolve_article_editor_step`, `find_first`, `create_newsletter_edition`, `/api/assets`, the settings upsert). The close is honest: both halves the owner signed off on are built, the decision comment was answered, no phase is left in prose. Three real defects, all fixed on `eed3bebb`:

- **The cover would never have attached in production** (`article_editor.py`): `attach_article_cover` targets LinkedIn's hidden `<input type=file>` and passes `visible_only=False` to say so, but `resolve_article_editor_step` ran `_is_displayed_safe()` unconditionally *after* `find_first` — so the hidden input was skipped on every route and the attach fell through to "Article cover upload control not found". Since a cover miss is (correctly) a warning and never a `failed_step`, editions would have published cover-less forever and it would have read as selector rot, not as a bug. The tests missed it because `FakeElement` defaults to `displayed=True`. Fixed: the visibility filter now follows `visible_only`; all four existing steps use the default `True`, so nothing else moves. Added a resolver test (hidden element survives `visible_only=False`) and an attach test against a hidden input.
- **The SPA called the old cover ready** (`NewsletterQueue.tsx`): the generation poll's success condition was "this edition has a cover", so re-generating over an existing cover matched on the very first tick — "Cover ready — review it below" while the new render was still running, polling stopped, and the new cover only appeared on some later refetch. Now it records the URL the edition carried when generation started and waits for a different one.
- **Generated covers leaked files** (`run_scheduler.generate_newsletter_cover`): the task overwrote `cover_image_path` without deleting the render it replaced — the upload endpoint does this, the task didn't — so every re-generation left an unreferenced file on the assets volume. It also ignored the result of `set_edition_cover_image`, leaving a stored file with no row behind a failed write. Both cleaned up, with tests.

Also checked and found sound: ownership is a 404 (never a 403) on all three routes and spends nothing; `_approved_cover_path` is the only status reader and pending never publishes; `cover_abs_path` re-resolves through `realpath` with a containment re-check; `create_newsletter_edition` really does return the new id (so the `cover_image_auto` hook queues an id, not a bool); `_find_asset_file` handles the nested `images/newsletter_covers/<uid>/` path; the migration is additive with a timestamp version; `rowcount >= 0` matches the established `db.py` convention for idempotent updates.

`poetry run pytest tests/unit -q` → **9263 passed**. SPA type-check not run here (no `node_modules` in this worktree) — CI covers it.

gitchrisqueen:

Issue #26, opened 2026-06-17
[M2] Write integration tests for content plan generation


gitchrisqueen:
Create `tests/integration/test_content_plan.py` testing the content generation pipeline.

## Coverage targets (mock DB, AI client, LinkedIn API)
- `plan_content_for_user(user_id)` — verify 30 days planned with balanced PostType distribution
- `create_text_post(user_id, planned_post)` — verify correct AI function called per source type (blog/website/news/story/engagement)
- `auto_generate_content()` Celery task — mock `get_active_users()`, verify task fires for each user

gitchrisqueen:
Resolved in PR #87 (backend fixes)

PR #911, opened 2026-08-02
fix(ui): mobile responsiveness — tables scroll, nav shrinks, modals fit (closes #894)

### LINKED ISSUE #894: feat(ui): Improve mobile responsiveness and table visibility
## Why
User cannot see the table element on mobile and reports that the entire site needs to be mobile friendly with resizing and layout changes to ensure elements remain displayed regardless of device or rotation.

Auto-filed from in-app feedback #26 by the feedback→auto-work loop (`docs/launch-and-marketing-plan.md` §B.2/B.3). Classifier: enhancement/medium, risk `none`, confidence 1.00.

[Watch the session replay](https://us.posthog.com/project/475262/replay/019fbdf3-adac-7e12-8104-8ab6edec564a) — the browser session this report was filed from.

## Scope
- Address the reported enhancement in the `ui` area of the product.

## Files
- `src/cqc_lem/ui/src/`
- `tests/unit/`

## Acceptance
- All required CI gates pass.

gitchrisqueen:
## What & why

In-app feedback #26: *"cannot see the table element on mobile… the entire site needs to be mobile
friendly with resizing and layout changes to ensure elements remain displayed regardless of device
or rotation."* I swept the SPA for the classes of breakage that produce that report and fixed the
three real ones at their source.

**1. Data tables were crushed or clipped.** A `w-full` table has no way to say "I need more room
than this", so a 375px phone honours the width and squeezes every column into an unreadable sliver.
Worse, the feedback-triage card rounds its corners with `overflow-hidden`, so the columns that
overflowed it were *clipped away entirely* — the Actions column did not exist on a phone. New
`components/TableScroll.tsx` is the ONE place that decides how a table survives a narrow screen:
the table keeps a readable floor width and the **wrapper** scrolls sideways. It renders as a
focusable, labelled `region`, because a scroll container only a wheel or a touch drag can reach is
unreachable by keyboard (WCAG 2.1.1). Applied to the dashboard per-post table (9 cols, floor 720px)
and the admin feedback table (8 cols, floor 900px). `LineChart`'s two-column data table already
fits and was left alone.

**2. The nav pushed the whole page sideways.** `LEM + 4–5 links + email + Log out` does not fit
375px, and nothing in that row could shrink — so the bar overflowed and carried "Log out" off the
right edge on every page. The link row is now the only element allowed to shrink (`min-w-0`) and it
scrolls horizontally instead; the account cluster is `shrink-0` so its controls always stay put.
The scroll gutter would eat visible height out of a 56px bar, so it's hidden with a new
`scrollbar-none` utility in `index.css` (touch, wheel and keyboard scrolling all still work).

**3. Modals did not survive rotation.** A phone in landscape is ~375px tall. The login, survey,
PostHog-survey, step-up and delete-confirm panels were uncapped, so they ran off the bottom of the
screen together with their submit buttons — and there was nothing to scroll, because the panel
itself *was* the overflow. All five are now capped at `max-h-[90vh]` with `overflow-y-auto`; the
login panel also drops from `p-8` to `p-6` below `sm`.

Plus the Dashboard header, whose title and two CTAs need ~390px of a single line, now wraps.

Swept and deliberately **not** changed: no fixed-pixel widths exist anywhere in the SPA; the
remaining unprefixed multi-column grids are small tile/thumbnail/button grids that fit a phone;
`SettingsHub`, `ContentStudio` and the review pages already scroll or wrap their control rows; the
feedback widget and version notice already clamp to `calc(100vw - 2rem)`.

## Testing

- New `TableScroll.test.tsx` (scroll behaviour, floor width, focusable named region, class
  passthrough) and `Layout.test.tsx` (link row shrinks + scrolls, account cluster is `shrink-0` and
  outside it, labels never wrap, email hides below `sm`).
- Extended `PerPostTable.test.tsx`, `AdminFeedbackPage.test.tsx` (table is inside a scroll region
  the card cannot clip) and `LoginModal.test.tsx` (panel is height-capped and scrolls itself).
- Full suite on Node 22: `npm test` → **45 files / 357 tests passing**; `npm run build`
  (tsc -b + vite) clean.

Closes #894

### DIFF
--- a/src/cqc_lem/ui/src/components/FeedbackWidget.tsx
+++ b/src/cqc_lem/ui/src/components/FeedbackWidget.tsx
@@ -122,7 +122,10 @@ export default function FeedbackWidget() {
   return (
     // Positioned by FloatingDock (issue #596); `relative` keeps the close button anchored to the
     // panel now that the panel is no longer the fixed element.
-    <div className="relative w-[22rem] max-w-[calc(100vw-2rem)] bg-white border border-gray-200 rounded-xl shadow-xl p-4">
+    // The width clamp is not the whole story on a phone: this form is ~350px tall and the dock
+    // grows it UPWARDS off the top of a landscape screen (worse with the keyboard open, which
+    // autoFocus guarantees), so it needs the same height cap the modals wear (issue #894).
+    <div className="relative w-[22rem] max-w-[calc(100vw-2rem)] max-h-viewport overflow-y-auto bg-white border border-gray-200 rounded-xl shadow-xl p-4">
       <button
         onClick={close}
         className="absolute top-2 right-3 text-gray-400 hover:text-gray-600 text-xl leading-none"
--- a/src/cqc_lem/ui/src/components/FloatingDock.test.tsx
+++ b/src/cqc_lem/ui/src/components/FloatingDock.test.tsx
@@ -92,6 +92,22 @@ describe('FloatingDock (issue #596)', () => {
     expect(pinned).toEqual([stack])
   })
 
+  // The dock is pinned to the BOTTOM, so an open panel grows upwards — off the top of a landscape
+  // phone, taking the "Send feedback" button with it. A width clamp does not cover that; the panel
+  // needs the same height cap the modals wear (issue #894).
+  it('caps the open feedback panel to the viewport and scrolls its own content', () => {
+    render(
+      <MemoryRouter>
+        <FloatingDock><FeedbackWidget /></FloatingDock>
+      </MemoryRouter>
+    )
+    fireEvent.click(screen.getByRole('button', { name: 'Feedback / Report a bug' }))
+    const panel = screen.getByRole('button', { name: 'Close feedback' }).parentElement as HTMLElement
+    expect(panel.contains(screen.getByRole('button', { name: 'Send feedback' }))).toBe(true)
+    expect(panel.className).toContain('max-h-viewport')
+    expect(panel.className).toContain('overflow-y-auto')
+  })
+
   it('keeps Save All reachable once the feedback panel is open', () => {
     const { container } = harness()
     fireEvent.click(screen.getByRole('button', { name: 'Feedback / Report a bug' }))
--- a/src/cqc_lem/ui/src/components/Layout.test.tsx
+++ b/src/cqc_lem/ui/src/components/Layout.test.tsx
@@ -1,0 +1,64 @@
+import { describe, expect, it, vi, afterEach } from 'vitest'
+import { cleanup, render, screen } from '@testing-library/react'
+import type { ReactNode } from 'react'
+import { MemoryRouter } from 'react-router-dom'
+import Layout from './Layout'
+
+const auth = {
+  user: { userId: 1, email: 'a-fairly-long-address@example.com' },
+  logout: vi.fn(),
+  openLoginModal: vi.fn(),
+  isAdmin: true,
+}
+vi.mock('../contexts/AuthContext', () => ({ useAuth: () => auth }))
+vi.mock('./AccountReadinessBanner', () => ({ default: () => null }))
+vi.mock('./FeedbackWidget', () => ({ default: () => null }))
+vi.mock('./FloatingDock', () => ({ default: ({ children }: { children: ReactNode }) => <div>{children}</div> }))
+vi.mock('./Footer', () => ({ default: () => null }))
+vi.mock('./PostHogSurveyModal', () => ({ default: () => null }))
+vi.mock('./ShippedNotice', () => ({ default: () => null }))
+vi.mock('./SurveyModal', () => ({ default: () => null }))
+
+afterEach(cleanup)
+
+function harness() {
+  return render(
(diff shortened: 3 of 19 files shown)


gitchrisqueen:

- **Modal height cap measured the wrong viewport.** All five panels used `max-h-[90vh]`, but `vh` is the *large* viewport — it measures the screen as if the browser's address bar were retracted. On the landscape phone this cap exists for (the PR's own justification), 90vh is still taller than the visible area, so the submit button can stay below the fold and the fix doesn't land. Replaced with a new `max-h-viewport` utility in `index.css` that caps on `dvh` and keeps the `vh` line as the pre-Safari-15.4 / pre-Chrome-108 fallback; verified both declarations survive the Tailwind v4 compile.
- **Test gap on 4 of the 5 modals.** `SurveyModal`, `PostHogSurveyModal`, `useStepUp` and the ContentStudio delete confirm were changed with no coverage, and nothing stopped the *next* modal or table from reintroducing the bug. Added `src/mobileViewport.test.ts`: every `fixed inset-0` panel must be capped + scrollable, and every data table must be wrapped in `TableScroll` (LineChart's two-column table is exempted with the reason inline).

Checked and found clean: `@utility scrollbar-none` is valid Tailwind v4 and compiles (confirmed against the real config, not assumed); the viewport meta tag is present; the nav's `min-w-0` + `shrink-0` split does make the link row the only shrinkable item, so `Log out` stays put; `AdminFeedbackPage`'s 900px floor still fits the 992px desktop content box, so this is not a desktop regression; the card's `overflow-hidden` can no longer clip columns now that the child scrolls itself; the PR's "swept and not changed" claims hold up — `SettingsHub`, `ContentStudio` and the review pages already scroll or wrap, the charts are `viewBox`-scaled, and the only unwrapped table is LineChart's two-column one. `Closes #894` is honest: the issue has no later phase and no unchecked criterion left behind.

Not run locally: the vitest suite needs Node 20+ and this box has Node 18. `tsc -b` and `eslint` are clean, and the two new invariants were verified against the tree with an equivalent script (0 offenders). CI is the source of truth.

gitchrisqueen:

Reviewed as the adversarial gate: read the full diff, the surrounding components, and the linked issue's acceptance criteria. Because every new assertion in this PR checks a **class string** rather than rendered CSS, I installed Node 22, ran `npm ci`, and inspected the **built** stylesheet — that is where both findings came from.

- **The feedback widget had the exact bug this PR fixes for modals.** The PR signs it off as already safe ("clamps to `calc(100vw - 2rem)`"), but that is a *width* clamp; the shape that broke the modals is *height*. `FloatingDock` pins the panel to the bottom, so the ~350px feedback form grows **upwards** off the top of a landscape phone — worse with the keyboard open, which `autoFocus` on the textarea guarantees — and "Send feedback" goes with it, with nothing to scroll. It is also the widget #894 was reported through. Fixed: the panel now wears the same `max-h-viewport overflow-y-auto` cap, with a rendering test in `FloatingDock.test.tsx`. (`mobileViewport.test.ts` can't catch this — it keys on `fixed inset-0`, and the panel isn't the fixed element.)
- **The `90vh` fallback was not in the shipped CSS.** `@utility max-h-viewport { max-height: 90vh; max-height: 90dvh }` is two plain declarations of one property, so the minifier drops the first: the built bundle contained `.max-h-viewport{max-height:90dvh}` only. On a browser without `dvh` (the browsers the comment names as the reason the line exists) the cap silently evaluated to nothing — the modal fix would have been a no-op there while every test still passed. Fixed by moving the `dvh` line into `@supports (height: 1dvh)`; both rules are now present in `dist/assets/*.css`, verified.

Checked and found sound, no change needed:
- `max-h-viewport` and `scrollbar-none` do resolve — I confirmed both emit real rules from the build, so the `@utility` names aren't being swallowed by Tailwind's `max-h-*` namespace.
- All five `fixed inset-0` panels in the tree are capped; the only `role="dialog"` is `useStepUp`, and it's covered.
- The `mobileViewport.test.ts` modal guard is stronger than it looks: `max-h-viewport` is the binding count, so an added-but-uncapped modal is caught even in a file (`ContentStudio`) that uses `overflow-y-auto` elsewhere.
- `TableScroll`'s `minWidth` floors (720 / 900) both fit inside `max-w-5xl` minus `px-4`, so desktop layout is unchanged; the `overflow-hidden` card on `AdminFeedbackPage` cannot clip a child that scrolls itself.
- The nav shrink is correct — `LEM` and the account cluster are `shrink-0`, only the link row has `min-w-0`, so `ml-auto` still parks "Log out" on the right.
- Sweep claims spot-checked: no arbitrary/fixed pixel widths in the SPA, and the remaining unprefixed `grid-cols-2/3` are small tiles whose text wraps rather than overflowing.
- `Closes #894` is honest — the issue's acceptance criteria are generic and met, and no later phase is left in prose.

Verification: `npm test` → **46 files / 361 tests passing**; `npm run build` (tsc -b + vite) clean; `npx eslint` clean on the touched files.

gitchrisqueen:

PR #913, opened 2026-08-02
fix(ci): CodeQL PR gate waits for THIS commit's analysis, not any analysis (closes #904)

### LINKED ISSUE #904: fix(ci): CodeQL PR gate compares against stale alerts — 'an analysis exists' is not 'an analysis for THIS commit'
## Symptom

A PR can fail `CodeQL PR Quality Gate` with alerts that **have already been fixed**, and pass on a bare re-run with **no code change**. It reads as a genuine quality regression and isn't one.

Observed on #899 (2026-08-02): the gate reported 5 new alerts, all of which the pushed commit had fixed. Re-running the gate after CodeQL analysis completed → **SUCCESS**, no code change.

The report contradicted itself, which is what gave it away:
- it flagged `py/unused-local-variable — Variable opened is not used` at a line where the `opened =` binding **had been deleted**
- the alert IDs (2641–2645) were **byte-identical** to the previous scan's

## Root cause

`scripts/codeql_pr_gate.py`:

```python
def analyses_exist(self, ref: str) -> bool:
    data = self.get("/code-scanning/analyses", params={"ref": ref, "per_page": 1})
    return bool(data and isinstance(data, list) and len(data) > 0)
```

The wait loop (~line 276) polls this and returns as soon as **any** analysis exists for `refs/pull/<n>/merge`:

```python
if client.analyses_exist(ref):
    log_info("CodeQL analysis available", ref=ref)
    return True
```

On any push after the first, an analysis for that ref **already exists from the previous commit**. So the wait returns instantly, and `fetch_alerts(ref)` returns the *previous* commit's alerts.

`.github/workflows/codeql-pr-gate.yml` triggers on `pull_request` — the **same event** as `CodeQL Advanced`, with no `needs:` between them. They run concurrently. On #899 both started at `01:25:29Z`.

So the gate is not "waiting for analysis" at all in the common case; it is confirming that analysis has run *at some point in this PR's history*.

## This is the same bug the fast-lane doc already documents

`docs/release-fast-lane.md`, on the v0.115.0 incident:

> Step 2 waits on the *run*, not on "a release PR exists" — those look equivalent and are not.

Identical shape, different file: here it waits on "an analysis exists" rather than "an analysis **for this commit** exists". Worth fixing with that precedent in mind — the failure mode is known, it just wasn't generalised.

## Why it matters beyond one re-run

- **It burns the signal.** A gate that cries wolf teaches people to re-run on red rather than read the report — which is precisely how a *real* new alert gets waved through.
- **It's silent and non-deterministic.** Whether you hit it depends on which workflow finishes first, so it will look flaky rather than wrong.
- **It can fail open too.** The same staleness means a genuinely-new alert on the current commit can be missed when the previous commit was clean — the gate compares yesterday's alerts and sees nothing new. **That direction is the dangerous one** and would never be noticed, because nobody investigates a green gate.

## Fix

The API already returns what's needed — `/code-scanning/analyses` includes `commit_sha` and `created_at`:

```json
{"category": "/language:python", "commit_sha": "31496b31...", "created_at": "2026-08-02T01:44:22Z"}
```

Options, roughly in order of robustness:

1. **Match the commit.** Require an analysis whose `commit_sha` equals the SHA the gate is analysing. Note `refs/pull/<n>/merge` resolves to the **merge** commit, not `head.sha` — compare against the merge ref's resolved SHA, not the PR head.
2. **Gate on freshness.** Require `created_at` newer than the workflow run's start time. Weaker (clock skew, re-runs) but a one-line change.
3. **Order the workflows.** Make the gate `workflow_run`-triggered on `CodeQL Advanced` completing, instead of racing it on `pull_request`. Removes the race rather than tolerating it — but changes how the check reports on the PR, so it needs care with required-checks config.

(1) and (3) are complementary; (1) alone fixes the correctness bug.

Whichever is chosen, the timeout path should stay **fail-open with a loud warning** as it is today — a gate that blocks on CodeQL being slow is its own outage.

## Acceptance

- [x] A gate run on commit N never compares against alerts from commit N-1.
- [x] A test covers the stale case: an analysis exists for the ref but for an older commit → the gate waits rather than comparing.
- [x] Timing out still fails open, with the warning it logs today.
- [ ] Verified on a real PR: push a fix for a flagged alert and confirm the next gate run passes **without** a manual re-run.

gitchrisqueen:
## What

`scripts/codeql_pr_gate.py` polled `/code-scanning/analyses?ref=refs/pull/<n>/merge` and returned as soon as **any** analysis existed. On every push after a PR's first, the previous commit's analysis is already there — so the wait returned instantly and the gate then diffed **commit N-1's** alerts.

The gate and `CodeQL Advanced` both fire on `pull_request` with no `needs:` between them, so this is a genuine race, not a rare one.

## Why it matters

- **False red** — it reports alerts the push already fixed. Observed on #899: five alerts, IDs byte-identical to the previous scan, one of them naming a `opened =` binding that had been deleted. A bare re-run then passed with no code change, which teaches people to re-run on red instead of reading the report.
- **False green** — a genuinely new alert on the current commit is missed whenever N-1 happened to be clean. **That direction is the dangerous one**, and it would never have been noticed, because nobody investigates a green gate.

Same shape as the v0.115.0 release incident already written up in `docs/release-fast-lane.md`: waiting on *"a thing exists"* rather than *"the thing for this run exists"*.

## How

Option (1) from the issue — match the commit — since it alone fixes the correctness bug:

- `GitHubClient.list_analyses()` replaces `analyses_exist()`; `wait_for_analysis()` now takes the commit SHA the ref resolves to and requires an analysis **for that SHA**.
- "Complete" means every **category** the previous commit on that ref produced (`/language:python`, `/language:python/advanced`, `/language:javascript-typescript`) has landed for our commit — a partial upload is still a stale diff for the categories not yet in. The required set is read off the ref rather than hardcoded, so adding or removing a CodeQL workflow later needs no change here. On a PR's first push there is no previous commit, so any analysis for the commit counts.
- The workflow passes `--head-sha ${{ github.sha }}` — the ephemeral **merge** commit, which is what CodeQL records as `commit_sha`. Verified against the live API: analysis `commit_sha=547c534b` for `refs/pull/911/merge` is `"Merge d556f4cf into ed097652"`, i.e. neither `head.sha` nor `merge_commit_sha`.
- `--wait-timeout` 300 → 900. The old default was never actually spent (the wait returned instantly); now that it waits for real it has to outlast a CodeQL run (~2–4 min) plus queue time, or the gate would just fail open every time.
- Timeout still **fails open**, with the same loud `::warning` + job-summary note, now naming the commit. It deliberately does **not** fall back to comparing an older commit — that is this bug, and it is worse than not comparing at all.

Option (3) (reordering via `workflow_run`) is left alone: the issue notes it is complementary, and it changes how the check reports on the PR, which needs care with required-checks config.

## Testing

New `tests/unit/test_codeql_pr_gate.py` (16 tests; the script had none). Covers the stale case directly — an analysis exists for the ref but for an older commit → `wait_for_analysis` returns False, and at the `main()` level `fetch_alerts` is **never called** and the gate fails open. Also: partial-category upload keeps waiting, the wait returns once the commit's analysis lands, a PR's first push accepts the only analysis there is, and `workflow_call` (no SHA) keeps its old behaviour.

```
poetry run pytest tests/unit/test_codeql_pr_gate.py -q   # 16 passed
poetry run ruff check tests/unit/test_codeql_pr_gate.py  # clean
```

## Acceptance

The first three boxes on #904 are ticked. The fourth — *"verified on a real PR"* — is observation, not code: this PR's own gate run is the evidence. Look for `CodeQL analysis available for commit <sha>` in the `CodeQL PR Quality Gate` log with the SHA matching this PR's merge commit (or, if CodeQL is slow, the fail-open warning naming that same SHA). Reproducing the full *"push a fix for a flagged alert"* scenario would require deliberately introducing a real CodeQL alert, which is not worth doing to a PR.

Docs: `docs/codeql-pr-gate.md` gains an **Analysis freshness** section recording the trap.

Closes #904

### DIFF
--- a/.github/workflows/codeql-pr-gate.yml
+++ b/.github/workflows/codeql-pr-gate.yml
@@ -36,17 +36,23 @@ jobs:
         id: refs
         shell: bash
         run: |
-          # TWO different identifiers, deliberately. `head_ref`/`base_ref` are COMMIT SHAs, used by
-          # actions/checkout and for the auto-fix commit. `analysis_ref`/`analysis_base` are GIT
+          # THREE different identifiers, deliberately. `head_ref`/`base_ref` are COMMIT SHAs, used
+          # by actions/checkout and for the auto-fix commit. `analysis_ref`/`analysis_base` are GIT
           # REFS, used for the code-scanning API — which returns ZERO analyses for a bare SHA and
-          # only resolves a ref. Conflating them is why this gate timed out on all 24 of its runs
-          # and failed open every time (verified against the live API 2026-07-31:
+          # only resolves a ref. Conflating those two is why this gate timed out on all 24 of its
+          # runs and failed open every time (verified against the live API 2026-07-31:
           # `?ref=<sha>` -> 0 analyses, `?ref=refs/pull/824/merge` -> 5).
+          #
+          # `analysis_sha` is the SHA `analysis_ref` RESOLVES TO, which the gate waits for so it
+          # never diffs the previous commit's alerts (#904). For a PR that is the ephemeral MERGE
+          # commit — `github.sha` — NOT `pull_request.head.sha` and NOT `merge_commit_sha`; CodeQL
+          # records `commit_sha` as the merge commit it analysed ("Merge <head> into <base>").
           if [ "${{ github.event_name }}" = "pull_request" ]; then
             echo "head_ref=${{ github.event.pull_request.head.sha }}" >> "$GITHUB_OUTPUT"
             echo "base_ref=${{ github.event.pull_request.base.sha }}" >> "$GITHUB_OUTPUT"
             echo "analysis_ref=refs/pull/${{ github.event.pull_request.number }}/merge" >> "$GITHUB_OUTPUT"
             echo "analysis_base=refs/heads/${{ github.event.pull_request.base.ref }}" >> "$GITHUB_OUTPUT"
+            echo "analysis_sha=${{ github.sha }}" >> "$GITHUB_OUTPUT"
             echo "pr_number=${{ github.event.pull_request.number }}" >> "$GITHUB_OUTPUT"
             echo "branch=${{ github.event.pull_request.head.ref }}" >> "$GITHUB_OUTPUT"
           elif [ "${{ github.event_name }}" = "merge_group" ]; then
@@ -56,6 +62,7 @@ jobs:
             # and refs/heads/main), so they are used verbatim.
             echo "analysis_ref=${{ github.event.merge_group.head_ref }}" >> "$GITHUB_OUTPUT"
             echo "analysis_base=${{ github.event.merge_group.base_ref }}" >> "$GITHUB_OUTPUT"
+            echo "analysis_sha=${{ github.event.merge_group.head_sha }}" >> "$GITHUB_OUTPUT"
             echo "pr_number=0" >> "$GITHUB_OUTPUT"
             echo "branch=" >> "$GITHUB_OUTPUT"
           else
@@ -63,6 +70,7 @@ jobs:
             echo "base_ref=${{ inputs.base-ref }}" >> "$GITHUB_OUTPUT"
             echo "analysis_ref=${{ inputs.head-ref }}" >> "$GITHUB_OUTPUT"
             echo "analysis_base=${{ inputs.base-ref }}" >> "$GITHUB_OUTPUT"
+            echo "analysis_sha=${{ inputs.head-ref }}" >> "$GITHUB_OUTPUT"
             echo "pr_number=${{ inputs.pr-number || '0' }}" >> "$GITHUB_OUTPUT"
             echo "branch=" >> "$GITHUB_OUTPUT"
           fi
@@ -87,6 +95,7 @@ jobs:
           # The code-scanning API needs REFS, not SHAs — see the "Determine refs" step.
           HEAD_REF="${{ steps.refs.outputs.analysis_ref }}"
           BASE_REF="${{ steps.refs.outputs.analysis_base }}"
+          HEAD_SHA="${{ steps.refs.outputs.analysis_sha }}"
           PR_NUMBER="${{ steps.refs.outputs.pr_number }}"
           BRANCH="${{ steps.refs.outputs.branch }}"
 
@@ -95,6 +104,7 @@ jobs:
           args=(
             --repo "${{ github.repository }}"
             --head-ref "$HEAD_REF"
+            --head-sha "$HEAD_SHA"
             --base-ref "$BASE_REF"
             --pr-number "$PR_NUMBER"
             --fix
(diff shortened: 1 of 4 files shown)


gitchrisqueen:

**1. (blocking) The fix left the same false green one step in — confirmed live on this PR.** Matching `commit_sha` closed the stale-*commit* hole, but `wait_for_analysis` returned as soon as *any* category landed for that commit whenever `previous_commit_categories` came back empty — which is every PR's first push, the most common case. This repo produces three categories from two workflows (`/language:python`, `/language:python/advanced`, `/language:javascript-typescript`) and they upload up to a minute apart, so head gets compared two-categories-deep against a three-category base and every alert in the missing one is invisible. Not theoretical — **PR #913's own gate run** (run `30737225889`) logged `CodeQL analysis available for commit` at `07:10:36`, while the live `/code-scanning/analyses` record shows `/language:python` did not upload until `07:10:54`. It compared two-deep. Same failure direction the issue calls the dangerous one.
   → `previous_commit_categories` → **`expected_categories`**: the largest category set among the newest **3** other commits on the ref (largest, not newest — one commit short a category, from a failed matrix leg or a run superseded mid-upload, no longer lowers the bar for everything after it). On a first push, where the ref has no earlier commit at all, the required set is calibrated off the **base ref**, which runs the same workflows and always has one; fetched once, not per poll. Nothing analysed anywhere → accept what landed, so a CodeQL-less repo is never blocked.

**2. `workflow_call` was handed a ref string as a commit SHA.** That branch has one input, `head-ref`, already used as `analysis_ref`; `analysis_sha=${{ inputs.head-ref }}` would compare it against `commit_sha` and never match, so any future caller passing a real ref would burn the full 900s and fail open every run. No caller exists today, which is exactly why it would have been found the hard way.
   → passes **no** `--head-sha`, keeping the documented pre-#904 wait, with the reason in the workflow.

**3. Doc claimed a property the code does not have.** "adding or removing a CodeQL workflow needs no change here" — *removing* one now costs up to 3 commits of fail-open timeouts while the category ages out of the lookback.
   → `docs/codeql-pr-gate.md` states the calibration rule, its cost, and the live 07:10:36/07:10:54 evidence.

Also: wait/timeout logs now name the missing categories (the actionable bit when it fails open), and `/analyses` sends `sort`+`direction` explicitly since newest-first is what the calibration is read off.

**Verified:** `tests/unit/test_codeql_pr_gate.py` 16 → **21 passed**; new cases cover largest-not-newest calibration, the lookback bound, the base-ref fallback landing and timing out, base fetched once, and an unanalysed base ref not blocking. `ruff check` clean on the test file (it had a >88-char line that CI's `ruff check src/ tests/` would have reported).

**On acceptance box 4** (`verified on a real PR`, still unticked): this push is that test. Commit `3f988ff4` is the PR's **second** commit, so `refs/pull/913/merge` now carries the previous commit's analyses — precisely the stale case. The gate run on this commit should log `Waiting for CodeQL analysis` with `missing_categories` until all three land for the new merge SHA, then `CodeQL analysis available for commit`. Not ticking it myself, since it is an observation nobody has made yet.

gitchrisqueen:
🧩 phase-guard — heads up: issue #904 still has **boxes: 1 unchecked** acceptance box(es). Merging anyway (unticked boxes are common here), but if any of that scope did NOT ship, file a follow-up issue with `agent:ready` and link it, or drop the `Closes #904` from this PR.

gitchrisqueen:

Issue #914, opened 2026-08-02
fix(security): /api endpoints that authenticate on an email/user_id parameter instead of the session


gitchrisqueen:
## Why

Found while reviewing #905 (session-scope narrowing, phase 2c.1). #905 restricts an
`extension`-scoped session to the ONE endpoint the browser extension calls, and holds a
`scope='enroll'` session to the enrolment surface. Both are enforced in
`api/main.get_session_user_id`, which is the one resolver every *session-authenticated* handler
calls.

**A set of `/api` endpoints don't call it at all.** They authenticate on an `email` or `user_id`
**query/body parameter** and act on whatever account that names. The only thing in front of them
is the shared bearer token (`API_ACCESS_TOKENS`), which the SPA ships in its build
(`VITE_API_TOKEN`, `ui/src/api/client.ts`) and is therefore known to anyone who loads the page.

This is **pre-existing** — it predates 2a/2b/2c and #905 neither introduces nor worsens it. But it
is the ceiling on what any of those phases can claim: a stolen extension token, or a session held
for enrolment, can still reach account data through these paths, so the blast-radius statement in
`docs/strong-authentication.md` is written with this limit named rather than pretending otherwise.

## Scope

Convert every `/api` route that identifies a user from a request parameter to resolve the caller
from the session (`get_session_user_id`) and treat the parameter as a *target* that must match —
or drop the parameter entirely. Confirmed instances in `src/cqc_lem/api/main.py` (verify the full
list while implementing; grep for handlers taking `email:` / `user_id:` that never call
`get_session_user_id`):

- `PUT /user/` — **changes the account email** given only the current one
- `GET /posts/`, `DELETE /posts/`, `POST /posts/bulk_update/`, `POST /update_post/`, `GET /post_url/`
- `GET /dashboard/stats/`, `GET /dashboard/planned-tasks/`, `GET /activity/`
- `POST /schedule_post/`
- `POST /create_weekly_content/`, `POST /invite_to_li_company_page/` (queue automation by `user_id`)

Watch for callers that legitimately have no session (the tutorial-capture harness, Celery, the
LinkedIn OAuth round trip) — those should use an explicit token or an admin path, not the
parameter.

## Acceptance

- [x] No `/api` route derives the acting user from a request parameter; each resolves the session
      and 401s without one.
- [x] Where a parameter names a target (an admin acting on another account), it is authorised
      explicitly rather than trusted.
[...]

PR #915, opened 2026-08-02
feat(security): mandatory strong-factor enrolment + extension-session narrowing — Phase 2c.1 (closes #905)

### LINKED ISSUE #905: feat(security): mandatory strong-factor enrolment + extension-session narrowing — Phase 2c.1 (follow-up of #897)

gitchrisqueen:
Phase **2c.1** — the two pieces #897 deliberately left out of 2c, filed as #905 so they could not be lost in prose. Design: `AUTH_SECURITY_DESIGN.md` §7 Stage 2 (second half) and §6.5.

## 1. Mandatory enrolment

`REQUIRE_STRONG_FACTOR_AFTER` — an env date, **empty by default and empty in every deployment today**, so this ships built and off.

| Date | PIN login for a **factor-less** account |
|---|---|
| empty | a full session — 2c behaviour, byte for byte |
| future | a full session, plus a dismissible SPA prompt naming the date |
| past | a session scoped **`enroll`** — signed in, and held |

A held session is **not a lockout**, and that distinction is the whole design. The PIN still signs the account in; what it stops handing over is a session that can do anything. `scope='enroll'` reaches the enrolment surface — who am I, sign me out, the SPA's boot payloads, the passkey/TOTP ceremonies, **and recovery codes** (being forced to enrol and then unable to save the sheet would be the worst possible order) — and nothing else, with a 403 `enrollment_required` everywhere else.

**The hold belongs to the ACCOUNT, not the session row.** `_scope_checked` re-asks `enrollment_required(user_id)` and releases the moment the answer is no. Deciding it from the row alone is a dead end on every *other* device: enrol on the laptop and the phone still holds a row saying `enroll` while the account now HAS a factor — so enrolling again is step-up gated, and the step-up ceremony is deliberately outside the enrolment surface. The same re-ask is what makes the rollback real (clear the date, move it forward, or `STRONG_AUTH_ENABLED=false` releases everyone *already* held) and what survives a promotion that failed to write.

The SPA reads the state off the **same** resolve that authenticated the request, so the browser can never be told "not held" by a page whose every request is then refused.

## 2. Extension sessions narrowed

`scope='extension'` was an ordinary full session that merely *also* satisfied the step-up gate at the cookie endpoint — a stolen token could read every post, DM template and setting the SPA can. It now reaches exactly `/api/user/linkedin-cookie`. Anything else is a 403 plus an audited `session_scope_denied` row: the extension calls one path, so that row cannot happen by accident and is the clearest available signal that someone else holds the token. A held *enrolment* session deliberately writes no such row — those refusals are constant and harmless while the SPA settles, and auditing them would bury the one that means something.

## Where it is enforced

Both holds live in `api/main.get_session_user_id` — the ONE resolver every session-authenticated handler already calls — and **not** at ~150 call sites, because a narrowing that has to be remembered per endpoint is a narrowing that leaks. The scope rides along on the new `db.resolve_session`, so the check costs **no extra query**. A scope we do not restrict — including a legacy `NULL` — behaves exactly as it did in 2b.

`sessions.scope` is `VARCHAR(32)`, so `enroll` needs **no migration**.

## Threat-model note

- **T2 (compromised mailbox)** is what mandatory enrolment closes: past the deadline the mailbox stops being sufficient for *any* account, not just the ones that opted in.
- **T4 (stolen session)** is what the extension narrowing shrinks: a token living in extension storage on a machine we do not control is now worth "someone can overwrite my `li_at`", not "someone can read my whole account".
- **Neither adds a lockout path.** The PIN remains a valid bootstrap, the first factor stays ungated, recovery codes stay reachable, and signing out is always available.
- **Honest limit — please read.** The narrowing binds every route that authenticates from the SESSION. A set of `/api` endpoints predating 2a identify the user from an `email` / `user_id` **request parameter** (`PUT /user/` — which changes the account email — `GET /posts/`, `GET /dashboard/stats/`, …) behind nothing but the shared bearer token the SPA ships in its build. They never call the resolver, so no session scope constrains them. **Pre-existing**: this PR neither introduces nor worsens it, but it is the ceiling on the blast-radius claim above, so it is named in the doc rather than glossed. Filed as **#914** (`risk:security`, `priority:high`).

## Testing

- **9,496** Python unit tests and **375** SPA tests pass.
- New: `tests/unit/api/test_session_scopes.py` (37) — both surfaces, the stolen-extension-token refusal, the multi-device release, the rollback, the audit row and its failure path, path-normalisation fail-closed; `test_auth_factors.py` (+11) — deadline parsing, kill switch, prompt logic; `test_db_identity_hardening.py` (+9) — `resolve_session` and the conditional promotion; `StrongFactorGate.test.tsx` + `StrongFactorPrompt.test.tsx` (18).
- **100% patch coverage on the changed Python** (`auth_factors.py` 98% overall, no new line uncovered).
- An adversarial pass over the diff found the multi-device dead-end, a UI/server disagreement on a DB hiccup, and a per-request WARNING flood on a malformed date — all three are fixed in this branch, with a regression test each.

## Operator notes

Nothing changes until someone sets a date. `.env.example` and `docs/strong-authentication.md` carry the rollout instructions; the doc's "Still open after 2c" section is rewritten to match reality.

Closes #905
Follow-up: #914


gitchrisqueen:

Held (`needs-human`, risk: security). This PR changes how sessions are authorised — it can hold a login and it refuses an existing token class — so it does not merge on the runner's review alone. **Everything here ships INERT**: `REQUIRE_STRONG_FACTOR_AFTER` is empty, so no user's login changes until you set a date. The extension narrowing is live on merge.

### 1. The extension narrowing goes live the moment this merges. Ship it as-is?
An `extension`-scoped token currently reaches your whole account; after this it reaches only `POST /api/user/linkedin-cookie`. You hold the only extension token in existence, and the extension only ever calls that one path — so the expected user-visible impact is zero.
- **A. Ship it — one path, 403 + audited `session_scope_denied` for anything else** — a stolen token is worth "overwrite my `li_at`", not "read my whole account". ✅ *recommended*
- **B. Also let it reach `/auth/session` and `/auth/logout`** — the extension could show "connected as you" and sign itself out; two more endpoints a stolen token can read.
- **C. Log the refusal but don't enforce yet** — you'd see would-be refusals in `auth_audit_log` for a release before anything breaks; the hole stays open meanwhile.

### 2. Do you want a `REQUIRE_STRONG_FACTOR_AFTER` date set now, or leave it empty?
Setting it is a one-line `.env` change on the VPS, no deploy. The SPA starts nudging the moment a date exists (dismissible), and holds factor-less logins once it passes.
- **A. Leave it empty in this PR; you set a date later when you're ready** — merges with literally zero login-behaviour change; you pick the date after watching the nudge land. ✅ *recommended*
- **B. Set it now to ~30 days out** — I'd add `REQUIRE_STRONG_FACTOR_AFTER=2026-09-01` to `.env.example` as a live default and you'd copy it to prod; the deadline is real and you are the first person it applies to.
- **C. Set it now to a date already past** — enrolment mandatory at your very next sign-in; strongest posture, and you'd be gated out of the app until you enrol a passkey (you can, and recovery codes are reachable).

### 3. #914 — I found a pre-existing hole while doing this. How do you want it handled?
A set of `/api` endpoints (`PUT /user/`, which **changes the account email**, plus `GET /posts/`, `/dashboard/stats/`, …) authenticate on an `email`/`user_id` query parameter and never touch the session, behind only the shared bearer token the SPA ships in its build. It predates 2a and this PR neither introduces nor worsens it — but it means no session scope, this one included, constrains those routes. I filed **#914** (`risk:security`, `priority:high`, `agent:ready`) and named the limit in `docs/strong-authentication.md` rather than letting the doc overclaim.
- **A. Leave #914 for the pipeline to pick up in its normal queue** — this PR merges now; #914 lands on its own with its own security review. ✅ *recommended*
- **B. Block this PR until #914 is fixed too** — one bigger security change, one review; nothing from #905 ships in the meantime.
- **C. Bump #914 to `priority:critical`** — it jumps the queue ahead of whatever else is `agent:ready`.

**My recommendation: `1A 2A 3A`.** Ship the narrowing (zero expected impact, real reduction in what a stolen extension token is worth), merge mandatory enrolment switched off so you choose the cutover date on your own timing rather than mine, and let #914 run as its own reviewed change instead of growing this diff into a second security fix.

gitchrisqueen:
Heads-up from #914 → PR #918 (now open, held for security review).

This PR's `docs/strong-authentication.md` diff adds:

> **Honest limit, and it is not small:** … A set of `/api` endpoints predating 2a still identify the user from an `email` / `user_id` **request parameter** … That is tracked as **#914** and is the ceiling on what this section can claim until it lands.

#918 lands it — all 14 of those routes now resolve their caller through `api/main.require_session_user_id()`, and a parameter is a target authorised to 403, never the actor. #918 writes the closed-hole statement into the **same region** of that file (right after the `extension`-scoped-session limit), so whichever of us rebases second will hit a conflict there rather than letting the stale claim land quietly.

**Ask:** when this rebases onto a `main` that has #918, drop the "#914 is the ceiling" paragraph or fold it into the one #918 added. The rest of the section is unaffected.

(Merge order is question 3 of the Decision Comment on #918 — the owner may instead choose to merge #915 first, in which case #918 carries the rebase and there is nothing to do here.)

gitchrisqueen:
ok

gitchrisqueen:
Addressed your review — you replied `ok`, so **1A 2A 3A**:

- **1A** — extension narrowing ships as-is: `scope='extension'` reaches only `POST /api/user/linkedin-cookie`, everything else is 403 + an audited `session_scope_denied`. No code change; it is now also **tested against `popup.js` itself** rather than asserted in prose (see below).
- **2A** — `REQUIRE_STRONG_FACTOR_AFTER` stays empty. `.env.example` documents it commented-out, so nothing about any login changes on merge. You set a date on the VPS when you want the cutover; no deploy needed, it is read at the call site.
- **3A** — #914 left in the normal queue at `priority:high`. Not folded into this diff.

Since all three were "keep as-is", the push you are looking at is the **automated review pass** on this PR, not your decision. Most of its blockers were reading a truncated diff and were already false (the Python tests, `db.py`, and the migration question — `sessions.scope` is `VARCHAR(32)`, no migration). Four findings were real and are fixed:

1. **Unknown scope failed OPEN.** `_scope_allows` granted everything to any scope missing from `_SCOPE_SURFACES` — a typo, a hand-edited row, or a scope a later phase adds and only half wires up. For a change whose whole thesis is "a narrowing you have to remember leaks," leaving the surface table itself opt-in was that same failure one level up. `_UNRESTRICTED_SCOPES` now names `full` and `recovery` explicitly; a legacy `NULL` row still resolves to `full`, so 2b sessions are untouched, and anything unrecognised is refused.
2. **No drift guard on either surface** — both are hardcoded path literals. Rename a route and the `enroll` surface becomes a **lockout** (the gate's own fetch 403s and the held user has nowhere to go). Two tests: every surface entry is a route the router really serves, and the extension surface **equals** the `/api` paths `browser_extension/popup.js` fetches — an equality both ways, because a missing entry breaks your reconnect click and a spare entry is blast radius handed to a stolen token for nothing. Writing the first one caught that `app.routes` is not the route table to read here (the router carries all 165), so the guard is real rather than vacuously passing.
3. **The hold was decided at two mint sites.** Now one — `_mint_login_session`. The strong-factor login paths (`/auth/passkey/login/complete`, `/auth/second-factor/verify`) deliberately still do not call it: reaching either proves the account holds a factor, so `enrollment_required` is false there by construction and a hold would be unreachable code posing as a control.
4. **Log levels.** A lost `session_scope_denied` row and a failed hold release were both `log_debug`. That row is the only signal that someone else is holding an extension token, so the refusal and the failed write now WARN. The enrolment refusal stays DEBUG — those are constant and harmless while the SPA settles, and per CLAUDE.md warning on an expected no-op files a defect for working behaviour.

Also dropped the discarded `session_token` argument from `_enrollment_held()` — in an auth module, a signature that answers about the current request no matter whose token you pass is a footgun, not call-site shape.

**Deliberately not done**, so it is not read as silence:
- **Session scope as an `Enum`** — a type refactor across `db.py` and every scope constant, inside a security PR, on the eve of merge. The fail-closed default in (1) removes the bug the enum was proposed to prevent. Worth doing; not worth doing here.
- **ContextVar single-stamp assertion** — theoretical. Both readers (`/auth/session`, `/user/auth-factors`) resolve exactly once and first.
- **`_request_object` → two narrower ContextVars** — cosmetic.
- The review's items 4 and 12 needed no change but did need checking, and both hold: `has_strong_factor` does not count recovery codes (so regenerating a sheet cannot satisfy the mandate — `test_recovery_codes_alone_are_not_a_strong_factor`), and a held session can still reach `/user/recovery-codes/regenerate` because `step_up_satisfied` passes for a factor-less account. No dead end.

**On #918:** its comment above asks me to drop the "#914 is the ceiling" paragraph when this rebases onto a `main` that has it. `main` does **not** have #918 yet (still open), so there is nothing to fold and the paragraph is still accurate. If #918 merges first, the rebase will conflict in exactly that region of `docs/strong-authentication.md`, which is what it was written to do.

9,500 unit tests pass locally.

gitchrisqueen:

Reviewed as if I had not written it: full diff, the surrounding auth code (`get_session_user_id`, `current_session_token`, `step_up_satisfied`, `enrollment_allowed`, `db.resolve_session`), every SPA surface a held session actually renders, `browser_extension/popup.js`, and issue #905's five acceptance criteria. 9,501 Python unit tests and `tsc --noEmit` pass locally on the fix commit; vitest could not run in this worktree (no vitest binary / Node 18), so the SPA suite is CI's word.

**Findings fixed (211382a8):**

- **The pre-deadline nudge could be silenced permanently by one click** — `StrongFactorPrompt` wrote a bare `'1'` to `localStorage` on `Not now` and every later render read it. So a dismissal from weeks before the cutover swallowed two things that are new information: an operator **moving** the date (forward is the dangerous direction — the user was told a later one) and the date **arriving**, which changes the message from "from `<date>`" to "from your next sign-in". Both the PR body and `docs/strong-authentication.md` claim "Nobody meets the deadline cold", which held only for someone who never dismissed. The dismissal is now keyed to the notice it dismissed, so each of those two is shown once more; dismissing the same notice twice still sticks. Two regression tests + the doc paragraph corrected.
- **The extension surface is method-blind** — a surface entry is a path, and `_EXTENSION_SESSION_SURFACE` pins `/user/linkedin-cookie`, not `POST /user/linkedin-cookie`. Adding a `GET` there later would hand every extension token, including a stolen one, a READ of the LinkedIn cookie — with nothing in the source saying so and neither existing drift guard noticing (one checks the path is a real route, the other checks it equals what popup.js fetches). Added a third guard pinning that entry to `POST` alone, so widening it has to be a deliberate edit to a test that explains why. Test-only; no runtime change.

Also added the missing **2c.1 row to `docs/AUTH_SECURITY_DESIGN.md`**'s build-order table — that is where the §6.5 extension caveat ("decide this in 2c, don't discover it") was raised, and the table still ended at 2c.

**Checked and found sound — no change needed:**

- Both narrowings really are enforced only in `get_session_user_id`; all four `create_session` call sites are accounted for (`_mint_login_session` covers both PIN paths incl. the no-mail bypass; the passkey and second-factor paths prove a factor by construction; there is no password-login path that could bypass the hold).
- `sessions.scope` is `VARCHAR(32)` (V20260802020439) and `auth_audit_log.event` is `VARCHAR(50)` — `enroll` and `session_scope_denied` genuinely need no migration.
- The `enroll` surface is complete for what the SPA actually mounts while held: `/auth/session`, `/auth/logout`, `/app-info` (Footer + `useNewVersion` + FeedbackWidget), `/flags`, `/user/auth-factors`, both ceremonies, recovery codes. `useStepUp`'s `/user/step-up/*` is correctly absent — a factor-less account passes `step_up_satisfied` and `enrollment_allowed` outright, so the gate never needs it, and it is available again the moment the row is promoted.
- Fail-closed paths hold: unknown scope, unknown path, WebSocket/no-middleware (`_request_path` `None`), and a restricted token presented alongside a full cookie (no fall-through).
- `get_session_user_id` now raises 403 from 121 call sites; only `auth_logout` wraps it in a `try`, and that one deliberately still deletes the row and clears the cookie — so an extension token can always sign itself out.
- `log_warning(..., path=, scope=)` are safe kwargs (`_extra` coerces, and neither collides with a `LogRecord` attribute), so a refusal cannot turn into a 500 on the logging path.
- `set_session_scope` has no production caller, so nothing can legitimately produce the unrecognised-scope refusal.
- Close is honest: all five acceptance boxes are implemented and ticked, and the deferred surface (#914, open, `risk:security` + `priority:high`) is linked in the body and named in the doc rather than left in prose.

gitchrisqueen:

gitchrisqueen:

PR #918, opened 2026-08-02
fix(security): /api endpoints resolve the caller from the session, not an email/user_id parameter (closes #914)

### LINKED ISSUE #914: fix(security): /api endpoints that authenticate on an email/user_id parameter instead of the session

gitchrisqueen:
## What & why

A set of `/api` routes read the **acting user** out of an `email` / `user_id` / `post_id` **request
parameter** and acted on whatever account it named. The only thing in front of them was the shared
bearer token (`API_ACCESS_TOKENS`), which the SPA ships in its build (`VITE_API_TOKEN`,
`ui/src/api/client.ts`) — so it is held by everyone who has ever loaded the page. `PUT /user/` was
the worst of them: it **moved the account email** given only the current one, which is the whole
account for one query parameter.

Every one of them now resolves its caller through **`require_session_user_id()`** — the ONE resolver
plus a 401 — and treats what the request named as a *target to authorise*.

| Helper | Rule |
|---|---|
| `require_session_user_id(token)` | the acting user, or **401**. Nothing below it runs without one. |
| `_reject_foreign_email(user_id, email)` | an `email` parameter must be the caller's own → **403** |
| `_reject_foreign_user_id(user_id, target)` | same, by id |
| `_require_own_posts(user_id, post_ids)` | **403** unless `db.user_owns_posts` proves EVERY id |

**Converted (15):** `PUT /user/` · `GET /posts/` · `DELETE /posts/` · `POST /posts/bulk_update/` ·
`POST /update_post/` · `GET /post_url/` · `GET /dashboard/stats/` ·
`GET /dashboard/planned-tasks/` · `GET /activity/` · `GET /user_id/` · `POST /schedule_post/` ·
`POST /create_weekly_content/` · `POST /invite_to_li_company_page/` ·
`POST /automate_reply_commenting` · `POST /aws_test_get_my_profile/`.

The last three were not in the issue's list but fail the same acceptance criterion — the first two
queue Selenium/LLM work against an account named by a parameter, the third is an email→id oracle
that turns a known address into the id those routes wanted.

### Three decisions worth reviewing

1. **`db.user_owns_posts` fails closed and judges a batch whole.** An empty list and a missing row
   both answer `False` — "we could not prove ownership" must never be spelled the same way as "they
   own it". One foreign id rejects the whole `post_ids` list. It answers **403**, not 404-per-id:
   which post ids exist is exactly the enumeration these endpoints used to give away. A database
   fault is the one case that is NOT 403: it raises `OwnershipUnprovable` and the API answers
   **503**. The action is refused either way, but "you may not touch these posts" and "we could not
   find out" are different facts, and collapsing them files a security-shaped defect through
   `_deny`'s recurrence escalation every time the database blips.
2. **`new_email` is REMOVED from `PUT /user/`, not gated.** The address moves through
   `POST /user/email/change/init|verify` — PIN to the NEW address, step-up gated, every other
   session revoked. Nothing in the SPA ever sent it.
3. **The parameter is checked, not ignored.** Answering a mismatch with the caller's own data would
   be a silent substitution, and a legacy client naming its OWN address keeps working either way.

### Also fixed (one line, same invariant)

`POST /generate-carousel` imported `db.get_session_user_id` **around** the module resolver, so the
SPA's cookie sentinel never resolved there (403 for every cookie session) and no session scope
reached it. There is one resolver; a route that imports around it is a bug.

### SPA

`Dashboard`, `ContentStudio`, `ComposePost` and `ContentProfileCard` send `session_token` instead of
`email`. The `/user_id/?email=…` → `/create_weekly_content/?user_id=…` hop is now one call — the
account is the session's, so there is no id to look up first.

## Testing

- **`tests/unit/api/test_param_auth_scoping.py` (new, 36 cases)** — the standing proof. Per
  converted route: no session → **401**, another account's identifier → **403**, and each case
  asserts *the db call behind it was never reached* (a handler that 403s after reading the row has
  already leaked it into a log line and a latency signal). Plus: own-target still works, one foreign
  id rejects a batch, `new_email` is ignored.
- **`tests/unit/utilities/test_db.py::TestUserOwnsPosts` (7 cases)** — fail-closed on DB error,
  duplicate ids, missing row, `user_id` scoping in the SQL.
- Rewrote the affected suites onto the new contract (`test_main_posts`, `test_main_general`,
  `test_post_url`, `test_planned_tasks_api`, `test_api_coverage_misc`,
  `test_schedule_timezone_breadcrumb`, `test_content_generation_status_api`) and the integration
  ones (`test_api`, `test_avatar_preview_api`, `test_post_gate_reason_api`).
- `poetry run pytest tests/unit tests/integration` → **9674 passed**. The two
  `test_link_first_comment` integration failures also fail on a clean checkout of `main` — unrelated.
- `tsc --noEmit` clean on the SPA.

## Notes for the reviewer

- **Coordination with PR #915 (issue #905).** #915 adds a paragraph to
  `docs/strong-authentication.md` naming #914 as the open ceiling on what that section can claim.
  This PR writes the closed-hole statement into the same region of that file, so a rebase of #915
  will conflict there and must reconcile — **drop #915's "#914 is the ceiling" paragraph**, or fold
  it into this one. Flagged as a comment on #915 too.
- **`src/cqc_lem/streamlit/`** still calls the pre-#914 shapes. It is legacy, not built, and appears
  in no compose file — deliberately left alone rather than widening this diff.
- No migration. No new env var. No behaviour change for a signed-in SPA user.

## Revise round — finishing the remaining parts of #914 (`1f6b143`)

The boundary held under review; four things around it did not say what they meant.

- **An outage is not a denial.** `user_owns_posts` answered `False` on a DB error, so
  `_require_own_posts` spelled an infrastructure fault "403 Forbidden" — and `_deny` warns with
  recurrence escalation, so a database blip filed a security-shaped defect. It raises
  `OwnershipUnprovable` now; the API answers 503 and refuses the action exactly as before.
- **A denied target is audited, not just logged.** `AuthAuditEvent.FOREIGN_TARGET_DENIED` lands in
  `auth_audit_log` beside the warning, for the same reason `_scope_checked` writes one: "has THIS
  account been naming other people's?" is a query, not a grep. Kind of identifier + path only — the
  caller-supplied address never goes in.
- **The browser cache was keyed on nothing.** The SPA's `sessionToken` is the same `'cookie'`
  sentinel for every account since #745, so moving Dashboard/`post_url` query keys from `email` to
  `sessionToken` gave them NO identity component: sign out, sign in as someone else in the same tab,
  and the previous account's dashboard renders from cache. Keys carry the user id, and `logout()`
  calls `queryClient.clear()` — which also covers the pre-existing hooks keyed the same way.
- **`schedule_post` 403 "User not found"** for a session that RESOLVED is our data fault, not the
  caller's permission → 500, logged.

Plus: `update_db_post_rejection_reason` gains the `user_id=` scope every sibling write on that table
already had; `_deny` is `-> NoReturn` (the dead `return` under it goes); `session_token` on every
request model is `repr=False`, so the class of line the deleted `myprint(f"…{post}")` belonged to is
harmless rather than absent; `get_post_by_email` is deleted rather than left an import away — it
turned an ADDRESS into somebody's posts, which is the shape `GET /posts/` used to authenticate on.

**On the two automated reviews' blocking item #1 (session tokens in query strings):** the SPA does
not put a credential in a URL. Since #745 the real token lives in an httpOnly cookie and the SPA
sends the non-secret sentinel `'cookie'` in `session_token` (`COOKIE_SESSION_SENTINEL`,
`AuthContext.COOKIE_SESSION`), which is why ~150 existing call sites keep their shape while holding
nothing. The parameter carries a real token only for non-browser callers, which is the pre-existing
`/content_generation_status/` shape. Their other blocking items (no tests, no `db.py` diff) were a
truncated diff — both have been in the PR since `2a0f5ab`.

`poetry run pytest tests/unit` → **9760 passed**. `CLAUDE.md` guard: 39,995 / 40,000.

Follow-up: #950 — the shared `/api` bearer token still ships in the SPA build (`VITE_API_TOKEN`).
#914 made it non-sufficient, which is the important half; retiring it needs an owner decision on
what the middleware becomes, so it is tracked rather than crammed in here. #950 also carries the
`db.update_user(email=…)` footgun this PR made unreachable but did not remove.

**`release:now`:** a live parameter-auth hole where any holder of the SPA's build-time bearer token could act on another account — the fast lane skips the ~168-minute release wait, never a check.

Closes #914


gitchrisqueen:

Held (`needs-human`, risk: **security**). PR #918 converts 14 `/api` routes that authenticated on an
`email` / `user_id` / `post_id` request parameter so they resolve the caller from the session. Three
calls in it are yours, not mine — all three are about how much of the old shape to keep.

### 1. Keep the `email` / `user_id` parameter as a validated no-op, or delete it?

Today the converted routes still ACCEPT the parameter and 403 when it names another account
(`_reject_foreign_email` / `_reject_foreign_user_id`). The SPA no longer sends it.

- **A. Keep it, validated (what the PR does)** — a legacy or external client naming its OWN address
  keeps working; a mismatch is a loud 403 rather than a silent substitution. Costs one extra
  `get_user_email` read per request that sends it (the SPA sends none).  ✅ *recommended*
- **B. Delete the parameters entirely** — one shape per endpoint, nothing to get wrong later, and no
  extra read. Breaks any caller still sending `email=` with a 422, including
  `src/cqc_lem/streamlit/` (legacy, unbuilt, not in any compose file) and anything you have in a
  bookmark or a curl script.
- **C. Keep it and ignore it** — never 403, just serve the caller's own data. Rejected: quietly
  answering a request about account X with account Y's data is its own bug class.

### 2. `POST /automate_reply_commenting` and `POST /aws_test_get_my_profile/` — session or admin?

Neither is called by the SPA. Both are duplicated by the `X-Admin-Secret`-gated `/admin/test/*`
endpoints, and both queue Selenium/Celery work.

- **A. Session-gated for the caller's own account (what the PR does)** — a normal user can still
  kick a reply run on their OWN post; the parameter becomes a target checked against
  `db.user_owns_posts`.  ✅ *recommended*
- **B. Admin-gate them (`_require_api_and_admin`, matching `/admin/test/reply`)** — narrower blast
  radius, and arguably where these belong. Takes the capability away from users entirely.
- **C. Delete both** — `/admin/test/reply` and `/admin/test/comment` already cover the operator
  case. Smallest surface, but it is a removal I would not make without you saying so.

### 3. Merge order vs PR #915 (issue #905) — who drops the stale doc paragraph?

#915 adds a paragraph to `docs/strong-authentication.md` naming **#914 as the open ceiling** on what
that section can claim. This PR edits the same region of that file to say the hole is closed, so a
rebase will conflict there and force a reconciliation rather than letting the stale claim land.

- **A. Merge #918 first; #915 rebases and drops its "#914 is the ceiling" paragraph** — the conflict
  makes the reconciliation unavoidable, and #915 is the newer PR so it carries the rebase.
  ✅ *recommended*
- **B. Merge #915 first; I rebase #918 and delete the paragraph here** — same end state, the rebase
  lands on me instead. Costs one more `MODE=rebase` tick.
- **C. Merge in either order and fix the doc in a follow-up** — leaves `main` briefly claiming an
  open hole that is closed. Not dangerous, just wrong in the file people read to reason about blast
  radius.

**My recommendation: `1A 2A 3A`.** Keeping the validated parameter is what makes this a security fix
rather than a compatibility break, session-gating leaves the two utility endpoints usable without
handing anyone another account, and letting #915 carry the rebase puts the conflict in front of the
PR that wrote the now-stale sentence.

gitchrisqueen:
ok

gitchrisqueen:
Addressed your review — `ok` applied as **1A 2A 3A**, plus a pass over the automated review. Pushed as `d554459`.

### Your decision (`ok`)

| | Choice | State |
|---|---|---|
| **1A** | Keep `email` / `user_id` as a validated parameter | Already what the PR does — unchanged. A caller naming their OWN address still works; a mismatch is a 403. |
| **2A** | `/automate_reply_commenting` + `/aws_test_get_my_profile/` session-gated for the caller's own account | Already what the PR does — unchanged. |
| **3A** | #918 merges first; #915 rebases and drops its "#914 is the ceiling" paragraph | Nothing to do here. **#915 needs the rebase after this merges** — flagging so it isn't missed. |

### Fixed from the automated review

- **Denials are visible.** A session that resolves and then names *another* account is a broken client or somebody working this hole — `_deny()` now `log_warning`s it (so the recurrence escalation files a repeat) with the caller's `user_id` and which parameter was denied. The named address is deliberately **not** logged: it's the caller-supplied half, and the audit log isn't the place to accumulate other people's addresses. **401 stays DEBUG on purpose** — sessions expire in the ordinary course of things and the SPA polls, so warning there would file a defect for working behaviour (`utilities/CLAUDE.md`).
- **`new_email` is refused, not discarded.** Real catch: dropping the field let Pydantic ignore it and answer `200 "User settings unchanged"`, so a client still sending it would believe the address had moved. It stays **declared** and answers **400** pointing at `POST /user/email/change/init|verify`. (Went with the declared-field-400 option rather than `extra="forbid"` — the pointer is worth more than a bare 422.) No SPA change needed: `ContentProfileCard` sends only `session_token` / `blog_url` / `sitemap_url`, and nothing in the SPA has ever sent `new_email`.
- **The post writes carry the rule twice.** `bulk_update_posts`, `soft_delete_posts` and `update_db_post` take an optional `user_id=` that scopes their `WHERE` clause, and the API passes it. The check in front is still the gate; the scope closes the check→write window and makes a future caller that forgets the check harmless rather than cross-account. Optional, so the non-API callers are unchanged.
- **`_reject_foreign_user_id` fails closed on a non-numeric target** instead of 500ing on `int()`.
- **`automate_reply_commenting_for_post_id` gets its missing `-> ResponseModel`.**

Tests: 5 new `db` cases for the owner-scoped writes, 4 for the denial-logging split, and the two `new_email` tests moved onto the new 400 contract. Full unit suite green locally — **9492 passed**.

### Not fixed, and why

- **"No tests in the diff" / "`db.py` isn't reviewable"** — both are already there and were before that review: `tests/unit/api/test_param_auth_scoping.py` (36 cases, one 401 + one 403 per converted route, each asserting the db call was never reached), `TestUserOwnsPosts` (7), and `user_owns_posts` in `utilities/db.py`. The review was reading a truncated diff.
- **`401` missing from `error_responses`** — it's there (`main.py:421`), so the `if k in [401, 403]` filters resolve.
- **DB error inside `user_owns_posts` → 503 instead of 403.** Not done. It's a fair point about support triage, but making it tri-state changes the fail-closed contract's shape on a security PR at merge time, and the current answer is never *wrong* — the action doesn't run either way. The DB error already `log_error`s with the `user_id`, which is the signal support needs. Happy to file it as a follow-up if you want the distinction.
- **`session_token` out of query strings / bodies.** Doesn't apply as written: the SPA's `sessionToken` is the non-secret sentinel `'cookie'` (`AuthContext.tsx:51`) and the real token rides in the httpOnly cookie, so nothing live writes a credential into an nginx log. Moving the parameter to a `Depends` across ~150 call sites is a repo-wide refactor, not this PR.
- **Rename `user_id` → `target_user_id` on the three routes.** That renames a public query parameter, which contradicts **1A** — a legacy client sending `user_id=` would start getting a 422.
- **`/post_url/` returning `200 {"post_url": null}` for a foreign id.** Left as-is: the lookup is already user-scoped (`get_post_url_from_log_for_user`), so it's non-disclosing, and it's the one route where a null is the honest answer for "your own post has no URL yet" — 403 there would make the SPA treat a not-yet-posted draft as an auth failure.
- **`get_post_by_email` is now orphaned in `src/`** (tests still cover it). Left in place rather than removing a public `db.py` helper in a security PR.

### Coverage grep (item 15)

```
grep -nE "^\s*(async )?def [a-z_]+\(.*\b(email|user_id): " src/cqc_lem/api/main.py
```
After this PR the only route-level hit left is `GET /auth/linkedin/`, whose `email` parameter is **unused** — it builds a redirect URL and the callback resolves the user from `session_token` carried in `state`. Everything else the grep returns is a module-private helper taking an already-authenticated `user_id`, not a request parameter. The `/admin/*` routes are separately gated by `X-Admin-Secret`.

gitchrisqueen:

Reviewed the full diff plus the seams it touches: every `/api` route signature and body model in `main.py` (to confirm nothing still derives the actor from a parameter), the `db.py` write paths that gained a `user_id=` scope, the SPA callers of all 15 converted routes, and non-SPA callers (extension, Celery, scripts, the OAuth round trip).

**What held up.** The conversion is complete: sweeping every route decorator + signature + risky body model, the only remaining parameter-named accounts are the `_require_api_and_admin` / `x-admin-secret` admin routes and the `/auth/email/*` bootstrap — both are the explicit exemptions in #914. `/auth/linkedin/`'s `email` parameter is dead (the user comes from `session_token` in the OAuth `state`). No route takes a `post_id` target without an ownership check. Parameter order in the new `owner_clause` SQL is correct in both `bulk_update_posts` and `update_db_post`, `get_posts(user_id, …)` is exactly what `get_post_by_email` wrapped, and dropping `myprint(f"Received Post Request: {post}")` from `/update_post/` matters more than it looks now that `PostRequest` carries `session_token`. The middleware logs path/method/status only, so no token reaches PostHog.

**Findings fixed** (`fbe4d38`):

- **`Closes #914` left acceptance criterion 4 unmet, and the doc contradicted itself.** `docs/strong-authentication.md` still carried the "Honest limit" paragraph naming #914 as the *open* ceiling — "Those never call the resolver, so no session scope … constrains them … the ceiling on what this section can claim until it lands" — in the present tense, ~100 lines above the new paragraph saying every route now resolves. The PR body predicted this ("a rebase of #915 will conflict there and must reconcile"), but #915 merged first, so no conflict ever fired and the reconciliation never happened. Replaced the stale paragraph in place, where the blast-radius claim is made, and dropped the now-duplicate one.
- **`CLAUDE.md ≤ 40,000 chars` was failing** — 40,719 against the cap; the PR added 611 chars to a file already 108 over on `main`. Compressed the three #745/#905 auth bullets back to the map (every symbol, endpoint, constant and invariant kept; the justification prose that `docs/identity-and-sessions.md` and `docs/strong-authentication.md` already hold, dropped) → **39,975**, guard green.
- **The CSRF section's second layer no longer covers four of the converted routes.** It claims "every mutating endpoint is POST/PUT with a JSON body", but `POST /create_weekly_content/`, `/invite_to_li_company_page/`, `/aws_test_get_my_profile/` and `/automate_reply_commenting` take query parameters and no body. Before #914 they authenticated on a parameter, so the cookie was irrelevant to them; they are cookie-authenticatable now, which is what puts them under that heading. Not exploitable — `SameSite=Lax` and the bearer token both hold, and each only queues work for the caller's own account — but it is two layers, not three, and the section itself says to revisit it when exactly this happens. Named the exception rather than leaving the claim overbroad.

**Not changed, worth knowing:** `db.update_user(email=…)` is now unreachable from the API (the only caller stopped passing it) but still exists as a way to move an account email without `user_email_history` or the session revoke — a latent footgun, not a live one, and removing it is outside this diff.

Tests: `tests/unit/api` + `tests/unit/utilities/test_db.py` → 1076 passed locally. This commit is docs-only.

gitchrisqueen:
🧩 phase-guard — **held before merge.** This PR closes issue #914, whose body declares work beyond this PR (phase: "phase 2"), and no follow-up issue is linked anywhere on this PR or that issue. Merging now would close #914 and drop that scope silently — exactly how #548 was lost.

To clear this, do ONE of:
2. Remove `Closes #914` from the PR body, say what remains, and re-label this PR `agent:working` — the issue stays open.

Assigning @gitchrisqueen.

gitchrisqueen:
Please finish the remaining parts of #914

gitchrisqueen:
Addressed your review — **finished the remaining parts of #914** rather than splitting them out. Pushed `1f6b143` + `1af8b74`.

The authorisation boundary itself held up under re-reading; what didn't was **what the code said around it**. Four things, all of which would have read as working:

1. **An outage was being reported as a denial.** `db.user_owns_posts` answered `False` on a database error, so `_require_own_posts` spelled an infrastructure fault "403 Forbidden" — telling a user they lack permission to their own drafts, sending on-call after an authorisation bug, and (because `_deny` warns with recurrence escalation) **filing a security-shaped GitHub issue every time the database blipped**. Failing closed was right and hasn't moved; the diagnosis was wrong. A DB fault raises `db.OwnershipUnprovable` and the API answers **503**.
2. **A denied foreign target is now audited, not just logged.** `AuthAuditEvent.FOREIGN_TARGET_DENIED` lands in `auth_audit_log` beside the warning — same reason `_scope_checked` writes one for a misused extension token: "has *this* account been naming other people's?" is a query, not a grep. Only the kind of identifier and the path go in; the caller-supplied address never does.
3. **The browser cache had stopped carrying an identity.** This is the one I'd flag hardest — it was introduced by this PR. Since #745 the SPA's `sessionToken` is the same non-secret `'cookie'` sentinel for *every* account, so moving the Dashboard and `post_url` React Query keys from `email` to `sessionToken` gave them no identity component at all: sign out, sign in as someone else in the same tab, and the previous account's dashboard renders straight out of the cache. A security PR must not open a client-side hole while closing a server-side one. Keys carry the user id now, and `logout()` calls `queryClient.clear()` — which also covers the pre-existing hooks keyed the same way.
4. **`schedule_post` answered 403 "User not found"** for a session that *resolved*. That's our data fault, not the caller's permission → 500, logged.

Plus the smaller ones from the same pass: `update_db_post_rejection_reason` gains the `user_id=` scope every sibling write on that table already had (the claim "forgetting the gate is harmless" is only true if *every* write carries it); `_deny` is `-> NoReturn`, so the dead `return` under it goes and a type checker can see the raise; `session_token` on every request model is `repr=False`, so the *class* of line the deleted `myprint(f"Received Post Request: {post}")` belonged to is harmless rather than merely absent; `get_post_by_email` is **deleted** rather than left one import away — it turned an address into somebody's posts, which is the exact shape `GET /posts/` used to authenticate on; and `/auth/linkedin/` loses its `email` parameter, the last signature in `main.py` naming an account it never read (it was also putting the address in a URL for nothing).

### On the two automated reviews' blocking item #1 — session tokens in query strings

Worth stating plainly because both reviews led with it: **the SPA does not put a credential in a URL.** Since #745 the real token lives in an httpOnly cookie and the SPA sends the non-secret sentinel `'cookie'` in the `session_token` field (`main.COOKIE_SESSION_SENTINEL` / `AuthContext.COOKIE_SESSION`) — that's the whole reason ~150 existing call sites keep their shape while holding nothing worth stealing. The parameter carries a real token only for non-browser callers, which is the pre-existing `/content_generation_status/` shape rather than something this PR introduced. Their other two blocking items (no tests, no `db.py` diff) were a truncated diff — both have been in the PR since `2a0f5ab`; `tests/unit/api/test_param_auth_scoping.py` is 54 cases now.

### Side-instruction captured, not crammed in

**#950** — the shared `/api` bearer token still ships in the SPA build (`VITE_API_TOKEN`). #914 made it *non-sufficient*, which is the important half, but it's still trusted as a gate while every page visitor holds it, and `docs/identity-and-sessions.md` still counts it as a CSRF layer. Retiring it needs a decision on what the middleware becomes, so it's tracked with real acceptance criteria rather than widened into this diff. #950 also carries the `db.update_user(email=…)` footgun this PR made unreachable but did not remove (it can still move an address with no `user_email_history`, no PIN, no session revoke). Linked as `Follow-up: #950` in the PR body, which also clears the 🧩 phase-guard hold.

### Scope check on `Closes #914`

Re-read the issue: all five acceptance boxes are met and I've ticked them in the issue body. The phase-guard fired on the words *"phase 2c.1"* in the Why section — that's #905, already merged — not on remaining scope of #914. Caveat on box 5 (`≥90% patch coverage`): Codecov is the judge, not me; I can only report the suite.

**Verification:** `pytest tests/unit` → **9762 passed**. `tsc --noEmit` on the SPA: no new errors. `CLAUDE.md` size guard: **39,995 / 40,000** (the new invariants are in; the justification prose that `docs/identity-and-sessions.md` already holds came out to pay for them).

**Not done, deliberately:** the `src/cqc_lem/streamlit/` tree still calls pre-#914 shapes — legacy, not built, in no compose file. Left alone rather than widening this diff.

gitchrisqueen:
✅ CI green, review satisfied & all threads resolved — merging.

PR #920, opened 2026-08-02
fix(litellm): weekly catalog PR describes its own diff + keeps the offline fixture in lockstep

gitchrisqueen:
Opened by the weekly model-health check (issue #716) on 2026-08-02, against 18 tags on ollama.com/api/tags.

## What changed

**`.litellm/model_upgrades.yaml`** — unchanged. No configured model has a newly published retirement date this week.

**`.litellm/ollama_catalog_snapshot.json`** (+ the offline fixture built from the same fetch) — the record of what the catalog looked like this week, so next week's scan reports a delta instead of the whole catalog:

- **new** `deepseek-v4-flash:0731` — 167GB; published 2026-07-31
- **new** `kimi-k3` — 1561GB; published 2026-07-27
- **gone** `kimi-k2.5`
- **gone** `minimax-m2.5`

Neither removed tag is configured. Both were retired by the vendor on 2026-07-31 and were already mapped to their replacements in `.litellm/model_upgrades.yaml` (`minimax-m2.5` → `minimax-m2.7`, `kimi-k2.5` → `kimi-k2.6`) — so this is the bookkeeping catching up, not news. Both new tags were benchmarked and **declined** in #924 (`:0731` did not beat the build we ship on any tier; `kimi-k3` is extra-usage-only billing, HTTP 402).

## What this PR does NOT do

- It does not touch `.litellm/config.yaml`, so no `lem-*` alias changes model here. Adopting a new tag is a separate, deliberate config change (the evaluation issue this scan files is where that decision gets made).
- The snapshot is a bookkeeping file read only by `scripts/model_health_check.py`; nothing loads it at runtime.

## Plus the fix for why this PR read the way it did

The commit above (`1d87573`) is not part of the weekly refresh — it is the fix for two defects in the job that opened it. Both are the reason the automated review blocked:

1. **Every PR this job opens arrives red.** `mutate_catalog` refreshed the snapshot but never the offline fixture the unit suite diffs it against, so `test_the_repo_config_and_committed_snapshot_are_clean_today` fails the moment the live catalog moves. `--catalog-apply --tags-fixture PATH` now writes both from the same fetch.
2. **The title and body were fixed strings.** The map-pre-approval half and the snapshot half fire independently, but every PR claimed both — so a reviewer was sent hunting for a `model_upgrades.yaml` diff that was never there. Title and body are now rendered from the plan (this body is that renderer's output).

Also new: a tag that leaves the catalog while `config.yaml` still points a deployment at it now rides the plan as `catalog.removed_configured`, prints `[STILL CONFIGURED]` in the dry run, and gets its own "needs a look before merge" section — that case is next week's 410 found early, and it used to look like ordinary housekeeping.

**Testing:** 9,464 unit tests pass locally, including 13 new ones covering the PR renderers, the fixture lockstep, and the configured-tag-removed path.

### DIFF
--- a/.litellm/ollama_catalog_snapshot.json
+++ b/.litellm/ollama_catalog_snapshot.json
@@ -1,10 +1,16 @@
 {
   "_comment": "Committed snapshot of ollama.com/api/tags, refreshed by scripts/model_health_check.py --catalog-apply. New tags here vs the live catalog become ONE agent:ready evaluation issue; do not hand-edit.",
-  "updated": "2026-07-27",
+  "updated": "2026-08-02",
   "models": {
     "deepseek-v4-flash": {
       "modified_at": "2026-04-24T00:00:00Z",
       "size": 140000000000,
+      "parameter_size": "",
+      "family": ""
+    },
+    "deepseek-v4-flash:0731": {
+      "modified_at": "2026-07-31T08:00:00-07:00",
+      "size": 166878536440,
       "parameter_size": "",
       "family": ""
     },
@@ -44,12 +50,6 @@
       "parameter_size": "",
       "family": ""
     },
-    "kimi-k2.5": {
-      "modified_at": "2026-01-26T00:00:00Z",
-      "size": 1118481408000,
-      "parameter_size": "",
-      "family": ""
-    },
     "kimi-k2.6": {
       "modified_at": "2026-03-31T00:00:00Z",
       "size": 595148192736,
@@ -62,9 +62,9 @@
       "parameter_size": "",
       "family": ""
     },
-    "minimax-m2.5": {
-      "modified_at": "2026-02-12T00:00:00Z",
-      "size": 230000000000,
+    "kimi-k3": {
+      "modified_at": "2026-07-27T08:00:00-07:00",
+      "size": 1560860324864,
       "parameter_size": "",
       "family": ""
     },
--- a/scripts/model_health_check.py
+++ b/scripts/model_health_check.py
@@ -48,6 +48,8 @@ Options:
   --snapshot PATH        Committed catalog snapshot (default .litellm/ollama_catalog_snapshot.json).
   --today YYYY-MM-DD     Override today's date (dry-run proofs / tests).
   --catalog-fixture DIR  Read cloud.md + tags.json from DIR instead of the network (offline proof).
+  --tags-fixture PATH    With --catalog-apply, also rewrite that offline tags.json from the SAME
+                         fetch the snapshot was written from, so the two can never drift apart.
   --no-usage-levels      Skip the per-model usage-level fetch (one page request per candidate).
 Exit: 0 healthy/no-op, 2 swaps/actions planned or applied, 3 manual alert needed, 1 error.
 """
@@ -443,6 +445,25 @@ def render_snapshot(models: dict, updated: str) -> str:
     return json.dumps(doc, indent=2, sort_keys=False) + "\n"
 
 
+def render_tags_payload(models: dict) -> str:
+    """Render a parsed catalog back into the /api/tags shape, for the committed offline fixture.
+
+    The fixture and the snapshot are two views of the SAME fetch, so a run that refreshes only the
+    snapshot leaves `tests/unit/fixtures/ollama/tags.json` describing a catalog that no longer
+    exists — and the shipped-state guard then fails on every PR this cron opens. Only the fields
+    `parse_catalog` reads are written (the live payload's `digest` is not one of them), so a
+    parse -> render -> parse round trip is the identity."""
+    payload = {"models": [{"name": name,
+                           "model": name,
+                           "modified_at": (models[name] or {}).get("modified_at") or "",
+                           "size": int((models[name] or {}).get("size") or 0),
+                           "details": {"family": (models[name] or {}).get("family") or "",
+                                       "parameter_size": (models[name] or {}).get(
+                                           "parameter_size") or ""}}
+                          for name in sorted(models or {})]}
+    return json.dumps(payload, indent=2) + "\n"
+
+
 def diff_catalog(snapshot: dict, current: dict) -> dict:
     """Tag NAMES that appeared or disappeared. A tag whose BUILD moved under an unchanged name is
     invisible here by construction — that is `plan_repoints`' job (issue #925)."""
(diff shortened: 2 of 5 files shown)


gitchrisqueen:
Addressed the review. You left no note with the `agent:revise` label, so I took the automated **Changes Required** verdict as the feedback — and it was right to block, for a reason it couldn't see: **CI was red**, and would have been on every PR this cron has ever opened.

### The two real defects (fixed in `1d87573`)

1. **This PR arrived red, structurally.** `mutate_catalog` refreshes `.litellm/ollama_catalog_snapshot.json` but never `tests/unit/fixtures/ollama/tags.json` — the offline fixture the shipped-state guard diffs that snapshot *against*. So the guard fails the moment the live catalog moves, which is precisely when this job runs. `--catalog-apply` now takes `--tags-fixture PATH` and writes both views from the same fetch (`render_tags_payload` is the inverse of `parse_catalog`; the round trip is asserted). The fixture change in this PR is exactly what the fixed job would have written.

2. **The title and body were fixed strings, wrong on most weeks.** The two halves — pre-approving a scheduled retirement into `model_upgrades.yaml`, and refreshing the snapshot — fire independently, but every PR claimed both. This week nothing was pre-approved, so the reviewer was sent looking for a map diff that did not exist. Title and body now render from the plan the orchestrator already read (the body above *is* that output).

Plus one thing neither of us had: a tag leaving the catalog **while `config.yaml` still points at it** is next week's 410, found a week early. It used to be indistinguishable from housekeeping; it now rides the plan as `catalog.removed_configured`, prints `[STILL CONFIGURED]` in the dry run, and gets its own "needs a look before merge" section.

### What I checked rather than fixed

- **"The diff is empty."** It wasn't — the reviewer's diff was truncated. The original diff was ±10 lines in one JSON file.
- **"Ollama doesn't retire models / a local endpoint doesn't 410."** These are **Ollama Cloud** hosted models (`api_base: os.environ/OLLAMA_CLOUD_URL`), not a local runtime. Ollama Cloud publishes a retirement schedule at `docs.ollama.com/cloud.md` and does return 410 — `kimi-k2.5` and `minimax-m2.5`, the two tags this PR records as gone, retired on 2026-07-31 from that published table. The retirement source is now named in the PR body.
- **"This changes production inference behavior / must not merge as a chore."** It doesn't. The snapshot is read only by `scripts/model_health_check.py`; nothing loads it at runtime, and `.litellm/config.yaml` is untouched, so no `lem-*` alias changes model. The body now says this explicitly rather than leaving it to be inferred.
- **"Pre-approved swaps without evaluation" / "verify the replacements exist."** Nothing was pre-approved this week. Where it does happen, the map only takes a replacement the vendor names *and* the live catalog still carries, and `weekly_model_check.sh` smoke-tests every tier on the box and rolls back before opening a PR. Separately, both new tags here were benchmarked and **declined** in #924.
- **"Snapshot must be secret-free."** Confirmed — the file is model names, sizes, dates and (empty) family/param strings. No hostnames, base URLs, or tokens; `api_base` is an `os.environ` reference in the config and never enters the snapshot. GitGuardian is green.
- **"Unbounded append growth."** `append_upgrade_mappings` is keyed and idempotent — an existing key is never duplicated — and the map is retirement-only, not a fallback chain.

### What I did NOT do

- **Splitting the snapshot refresh from model changes into two PRs** — they already are two things, and the map half simply didn't fire this week. The generated title/body now make which one you're looking at obvious, which was the actual problem.
- **A CODEOWNERS / no-auto-merge rule for bot config PRs** — real, but a repo-policy change, not this diff. Say the word and I'll file it.

**Testing:** 9,464 unit tests pass locally (the suite was failing on this branch before), including 13 new ones over the PR renderers, the fixture lockstep, and the configured-tag-removed path.

gitchrisqueen:
Retitled `chore(litellm):` → `fix(litellm):` — the diff now carries a real bug fix (the cron's PRs were arriving red), so it should land a changelog entry rather than pass as housekeeping. The generated title for a future pure-data refresh stays `chore(litellm):`, which is what those are.

gitchrisqueen:

Reviewed the diff, the surrounding scan/plan code, and the seams with `tests/unit/test_litellm_roster_config.py`. Both halves of the change are sound (`render_tags_payload` is a true inverse of `parse_catalog`, the plan-rendered title/body match what `_catalog_apply` writes, `removed_configured` keys off the same `bare` id `plan_repoints` does). Two real defects in the consequences of it, both fixed in `621ccf8f`:

- **The cron's PR still arrives red — just on a different assertion.** Rewriting `tests/unit/fixtures/ollama/tags.json` from the live catalog is the right fix for the lockstep, but that file is also the input to unit tests that pin a candidate NAME. Simulated against the shipped fixture: publishing `glm-5.3` — a tag LEM neither configures nor serves — fails `test_minor_bump_is_reported_at_lower_urgency`, which pins `glm-5.1 -> glm-5.2`. Fixed: `TestPlanFamilyUpgrades` now reads `_frozen_catalog()` (real names, pinned at 2026-08-02; `plan_family_upgrades` reads tag names only, so no assertion loses meaning). The checks that must track the vendor — the shipped-state guard, the parse-every-tag check, the snapshot/fixture lockstep — still read the live fixture. Same rule `test_litellm_roster_config.py` already states: never pin a moving id, or the cron's own correct output fails CI.
- **`catalog.removed_configured` reached nobody.** `[STILL CONFIGURED]` prints from `_print_catalog_plan`, and the cron never runs that path (it uses `--catalog-json`) — so the only surfacing was a paragraph in a PR body that the agent pipeline opens *and merges* on its own, and merging re-baselines the snapshot, so that run is the only one that can ever report it. Fixed: `weekly_model_check.sh` now emails it through the same `alert()` path the retirement notices use, silent on any parse failure like the rest of that best-effort block.

Verified, not fixed — flagging for the record:

- The week a configured family gets a newer version, `test_the_repo_config_and_committed_snapshot_are_clean_today` and `test_the_family_scan_has_nothing_left_to_file` both go red on the cron's PR (reproduced with the `glm-5.3` simulation). Both predate this branch, both read the committed snapshot, and both look deliberate — the roster is genuinely trailing at that point and the scan files an `agent:ready` evaluation issue for it. Weakening a guard whose whole job is "the shipped roster does not trail its own family" is a product call, not a review fix, so I left it. Worth knowing that "this job's PRs no longer arrive red" holds for tag churn but not for a family release.
- No `Closes #N` in the PR body, so nothing is claimed and no phase is left untracked.

Testing: 9,714 unit tests pass locally; `bash -n` clean on the orchestrator, and the new `removed_configured` extractor exercised against a real plan shape and against garbage input (stays silent).

gitchrisqueen:

gitchrisqueen:

PR #922, opened 2026-08-02
fix(benchmarks): recalibrate the deterministic floor so the gate can open (closes #910)

### LINKED ISSUE #910: Benchmark gate is unactionable as calibrated — no model (champions included) clears the 90% deterministic floor (follow-up of #842)
Discovered by the #842 measurement run (`bm-20260802-20ae40`, report:
`docs/model-benchmarks/2026-08-02-bm-20260802-20ae40.md`).

## Why

The #721 gate emits a swap recommendation only when a candidate clears its tier's **absolute**
floors (`deterministic_pass_rate: 0.9`) *and* meets-or-beats the champion. On the first real run
across `lem-medium` and `lem-complex`, **no model cleared the deterministic floor — including both
reigning champions**:

| Tier | Model | Role | Deterministic | Judge |
|---|---|---|---|---|
| lem-complex | `qwen3.5:397b` | champion | 50% | 100% |
| lem-complex | `gemma4:31b` | candidate | 80% | 57% |
| lem-complex | `deepseek-v4-flash` | candidate | 70% | 57% |
| lem-complex | `glm-5.2` | candidate | 70% | 86% |
| lem-complex | `minimax-m3` | candidate | 40% | 75% |
| lem-medium | `gpt-oss:120b` | champion | 60% | 50% |
| lem-medium | `deepseek-v4-flash` | candidate | 60% | 83% |
| lem-medium | `minimax-m3` | candidate | 50% | 80% |
| lem-medium | `gemma4:31b` | candidate | 40% | 100% |
| lem-medium | `glm-5.2` | candidate | 40% | 75% |

A floor that the incumbent itself fails is not a floor — it is a gate that can never open. #842's
verdicts are still correct on the *relative* half (no candidate beat its champion outright), but as
calibrated the harness cannot ever produce a `recommend`, so a genuinely better model would land as
`reject` too and nobody would notice.

The failures themselves are REAL model behaviour, not harness artifacts — the run shows them
case-by-case: long-form overruns the fixture's `max_chars` (champion `qwen3.5:397b` on 4 of 5),
`slop_lint` contrastive frames, and on `lem-medium` the #617 `comment_contract` ("adds no
experience, data point, disagreement, or genuine question") on 4–6 of 10 comment cases for every
model measured. In production those are exactly what the regeneration gates catch and retry, so a
single-shot pass rate of 40–80% is expected. The suite scores a first draft; production ships an
n-th draft.

## Scope

1. Decide what the deterministic floor is measuring and recalibrate it — options: score per-case
   *after* the same retry/repair loop production uses, drop the absolute floor to a
   champion-relative rule, or keep it and record it as advisory rather than gating.
2. Reasoning-model measurement stability: `minimax-m3` returned an EMPTY completion twice in that
   run after exhausting the shipped 2-doubling truncation retry (`BENCHMARK_TRUNCATION_RETRIES`);
   re-running the same two cases at the same effective budget produced passing answers both times
   (`complex-thought-leadership` 2122 chars ✅, `medium-comment-deploy-post` 424 chars ✅). One run of
   one case is therefore not a stable measurement for a reasoning model — the harness needs either
   repeats or an explicit variance note per case.

## Acceptance

- [x] The deterministic floor is either recalibrated against measured reality or explicitly
      documented as advisory, with the reasoning written into `docs/model-benchmarks/README.md`.
- [x] A re-run of the #842 roster under the new calibration produces a verdict that could, in
      principle, be a `recommend` — i.e. the gate is demonstrably openable.
- [x] Reasoning-model variance is addressed: repeats, or a per-case variance note the report
      renders, so an empty-then-fine case can never read as a quality verdict.
- [x] Cases whose `max_tokens` mirrors a production call site are either exempted from the truncation
      retry or scored on their first attempt, so a model cannot pass a budget production would not
      give it. *(added from the calibration comment below; exempted via `budget_mirrors_production`)*

gitchrisqueen:
## Why

`bm-20260802-20ae40` measured 40–80% deterministic pass rates for **every** model in the roster, both reigning champions included, against a 90% absolute floor. A floor the incumbent itself fails is a gate that can never open: a genuinely better model would have landed as `reject` and nobody would have noticed.

The per-case failures are real model behaviour, not harness artifacts. What was wrong was grading them all the same way — **the suite scores a first draft; production ships an n-th.**

## What changed

**1. The floor is recalibrated, not deleted.** Every assertion is now classified by what production does with its failure:

| Class | Meaning | Examples |
|---|---|---|
| `contract` | the call site consumes it — nothing repairs it | broken JSON, a preamble, a classification returned as prose, a post over LinkedIn's own 3000-char limit, no output |
| `repairable` | a regeneration gate catches and redrafts it | `slop_lint`, `comment_contract`, `max_similarity`, `min_burstiness` — `lint_repaired`, the #617 comment gate (then the post is skipped), `_review_generated_post` + `evaluate_post_gates` (then the post is held) |

The ABSOLUTE floor moves to `contract_pass_rate` (default 0.9). `deterministic_pass_rate` stays as an **advisory** reading, rendered with its target and the redraft count, and is still a meets-or-beats against the champion — needing more drafts than the incumbent is a real cost (tokens, latency, skipped comments) even when nothing broken ships. Repairable-ness is per fixture assertion, not per type: on long-form the tier's craft length target is repairable and the hard 3000-char limit beside it is not, so both live on the same case.

**2. The gate is demonstrably openable — shown by replay.** A live re-run spends metered Ollama Cloud quota and needs `BENCHMARK_ENABLED` + the key, so the demonstration re-scores the #842 report's OWN measured failures — one entry per ❌ line of the committed report — through the real classifier and the real gate (`TestBm842Replay`). Result: `deepseek-v4-flash` on `lem-medium` clears every expectation and lands as **`recommend`** (tied the champion deterministically, beat it 83% vs 50% on judge, same Medium usage level — exactly the model the #842 write-up called "the model to re-measure first"). It is the only one; every other rejection now names a quality reason instead of the floor everybody failed. `lem-complex`'s champion still misses the floor at 70% contract — that is now a TRUE signal (three drafts over LinkedIn's hard character limit) and it is named on every verdict for that tier rather than left to be inferred.

**3. Production-mirror budgets are exempt from the truncation retry** (owner's third calibration comment). `simple-relevance-yes-no` is `max_tokens: 3` because `ai_helper.py`'s relevance check is; retrying at 6 and 12 credited a model with something LEM could never run there. Those cases carry `"budget_mirrors_production": true` and are never escalated; ones that end empty are named under 🔒 *production budget*. A test fails the build if the flag is ever set on a case whose budget is above 10 tokens, so it can't quietly re-disable the #842 reasoning-headroom retry.

**4. Reasoning-model variance.** An empty completion is re-measured at the SAME budget (`BENCHMARK_EMPTY_REPEATS`, default 1 — `minimax-m3` answered both of its empty #842 cases on a re-run), every case that needed it is named under ⚖️ *measurement variance*, and one that stays empty is recorded as **no output** rather than as a 0-char answer that rendered as `min_chars: 0 chars < 700` — a quality verdict on a model that was never measured.

Leaderboard gains a **Contract** column; pre-#910 rows are still parsed and re-emitted at the new width with `n/a`, so adding a column doesn't silently evict committed history.

## Testing

- `tests/unit/test_benchmark_models.py`: 186 tests (was 156) — classification defaults + per-fixture override, both scorecard rates, the advisory-not-a-floor gate posture, the champion-misses-its-own-floor note, the budget exemption, the empty re-measurement, no-output-vs-0-chars, legacy leaderboard rows, and the #842 replay.
- Full unit suite green: 9,480 passed.
- `--print-suites` and a full `--dry-run` render exercised end to end.

## Scope

Every acceptance box on #910 is covered, including the extra one from the owner's comment. Criterion 2 asks for a re-run of the #842 roster: a live one needs the metered key, so this PR delivers the replay of that run's own measurements as a committed test and the next weekly `scripts/weekly_model_check.sh` run produces the first real scorecard carrying both columns.

Closes #910

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -225,8 +225,13 @@ property can't be decided locally). Three registered: `cost-routing-arm`, `comme
 `scripts/benchmark_models.py` measures candidates vs each tier's contract suite (NO user data),
 always beside the current champion. Deterministic is source of truth (in-repo linters, not a
 copy); LLM judge is PostHog Evaluations filtered on `benchmark_run_id` (production never carries
-it → customer never billed). Only `recommend` becomes a swap; recommendations are RENDERED, never
-written (`.litellm/model_upgrades.yaml` is the retirement map). Full posture:
+it → customer never billed). **The suite scores a FIRST draft, production ships an n-th** (#910), so
+every check is classed by what production does with its failure: `contract` (the call site consumes
+it — the ABSOLUTE floor) vs `repairable` (a regeneration gate retries it — advisory, still
+champion-relative). A case whose `max_tokens` mirrors a call site (`budget_mirrors_production`) is
+never retried at a bigger budget; an EMPTY completion is re-measured at the same one, and a still-
+empty one is no output, never a 0-char answer. Only `recommend` becomes a swap; recommendations are
+RENDERED, never written (`.litellm/model_upgrades.yaml` is the retirement map). Full posture:
 `docs/model-benchmarks/README.md`.
 
 ### Content-quality telemetry (issue #630)
--- a/docs/model-benchmarks/README.md
+++ b/docs/model-benchmarks/README.md
@@ -28,7 +28,7 @@ Two scoring layers, in this order:
 1. **Deterministic** (free, in-repo, the source of truth) — length caps, comma-list shape, JSON
    validity, `slop_lint.lint_report`, `content_framework.comment_contract_report`, burstiness,
    lexical self-similarity, and `routing_policy.complexity_tier`. A case that fails here never
-   spends a judge call.
+   spends a judge call. Reported as **two** rates — see *What the deterministic rates mean* below.
 2. **LLM judge** (paid, capped) — PostHog Evaluations scoring the harness's own tagged
    `$ai_generation` events, filtered to a `benchmark_run_id` so it can never bill against
    production traffic. With no judge provider configured, the runner degrades to an in-runner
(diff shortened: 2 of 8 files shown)


gitchrisqueen:

Reviewed the full diff against #910's four acceptance criteria, the committed `bm-20260802-20ae40` report, and the surrounding harness. Verified by replay and by exercising the real CLI (`--print-suites`, a full `--dry-run` render including the leaderboard merge, then reverted the artifacts).

- **A case that produced NOTHING rendered as "production would ship this"** (`render_report`, `scripts/benchmark_models.py`) — fixed in 1c5fff62. `evaluate_case` correctly classes a missing output as a *contract* failure (`"no output from the model"`), so the new branch — which keys purely on `contract_failures` being non-empty — sent every provider error and every empty completion to `❌ \`case\` — production would ship this`. Nothing was shipped; nothing was even measured. This PR turns an empty completion into no-output, so it is now the **common** path, not an edge case, and it is exactly the misreading the contract/first-draft split exists to prevent — `minimax-m3`'s two empty #842 cases would have rendered as shipped defects. Reproduced end to end before the fix. `failed_cases` now carries the case's error and the renderer branches on it first (`❌ … — no output; nothing was measured here` + the error), with a test. The gate is unchanged: no output still fails both rates and `provider answered every case`.
- Same commit tightens the 🔒 *production budget* note, which reads as if it names every exempted case when `budget_locked` is only ever set for a case that spent the mirrored budget **without answering**.

Checked and found sound, no change needed:
- **The replay is faithful to the committed report.** Walked all 10 scorecards ❌-line by ❌-line. It correctly catches that `complex-personal-story` at 3034 chars breaches the *new* 3000 hard cap as well as the 2600 craft cap, which is what puts `qwen3.5:397b` at 70% contract rather than 80%. `deepseek-v4-flash`/`lem-medium` reaching `recommend` is real: contract 100% (tie), first draft 60% (tie), judge 83% ≥ 80% floor and ≥ champion's 50%, 6 judged ≥ min 4, flat quota → `adopt`.
- **The advisory genuinely cannot gate.** `advisories` is a separate list; `verdict` and `blockers` read `expectations` only.
- **Leaderboard migration works on the real file**, not just in the test — a live `--dry-run` render re-emitted all 10 pre-#910 rows at the new width with `n/a` under Contract and appended the new row. No history evicted, no header/divider re-ingested.
- **The `complete()` rewrite** — traced every path (escalate→repeat, repeat→escalate, production-budget, unbounded, mid-loop provider exception). The repeat re-asks at the same budget and never grows it; `budget_locked` is only recorded when the case ends with no answer; per-attempt timing is preserved.
- **The close is honest.** All four acceptance boxes are covered, including the owner's calibration comment (`budget_mirrors_production` + the ≤10-token build guard). No later phase is left in prose. Criterion 2 is met by replay rather than a live re-run — that substitution is disclosed in the PR body, in `docs/model-benchmarks/README.md` and in the test module's own header, and a live run needs metered Ollama Cloud quota, so it is the right call.

Not findings, flagged for the record: classing `comment_contract` as repairable makes `lem-medium`'s contract floor easy to clear (every model in the replay lands at 90–100%), so discrimination on that tier now rests on the judge rate and the champion-relative first-draft comparison. That is the documented intent of the recalibration, and the alternative is the gate that could never open.

Unit tests: 187 passed (`tests/unit/test_benchmark_models.py`), all CI checks were green on the previous head.

gitchrisqueen:

Issue #923, opened 2026-08-02
benchmark: a harness-wide outage renders as a full 0% quality report and lands in the rolling leaderboard


gitchrisqueen:
## Why

`scripts/benchmark_models.py` treats an **infrastructure** failure as a **quality measurement**, and
then commits it to the permanent record.

`ProviderClient.complete` catches every exception and turns it into a case result
(`{"text": None, "error": ...}`), which is right for a provider 500 or a rate limit. But it is also
what happens when the harness itself cannot run — a broken venv, a revoked `OLLAMA_CLOUD_API_KEY`, a
DNS outage. The run then completes "successfully": it renders a full report, scores every model
including the champions at **0%**, and `write_report` appends one leaderboard row per model to
`docs/model-benchmarks/README.md`, where they are indistinguishable from a real bad run forever
after.

Hit for real while working #921 (2026-08-02): a worktree whose venv had no `openai` installed
produced a complete, plausible-looking report — six rows, `reject` on every verdict, champions at
`0% (0/10)` — off a run that never made a single HTTP request. The only trace of the real cause was
`No module named 'openai'` repeated under each case's ❌ detail. It was caught by eye, not by the
harness.

This matters because the harness runs **unattended**: `scripts/weekly_model_check.sh` invokes it and
opens the report as a PR (#842 decision `1B`). An expired key would ship a PR asserting that every
model LEM runs scores zero.

The gate is not the problem — "provider answered every case" failed correctly. The problem is that
the run is still published, and that "the model produced nothing" and "the harness could not call
anything" render identically in the scorecard and in the leaderboard.

## Scope

- Distinguish a **provider/case** error from a **harness-wide** one. A case error that repeats across
  *every* case of *every* model — champion included — is not a measurement.
- On that condition the run must fail loudly (non-zero exit, no leaderboard write) rather than
  publish. `#910` already established the principle for one case: a completion that was never
  measured is `no output`, "not a zero-length answer". This is that principle at run scope.
- Keep the honest half: a real per-case provider failure (one model 500s, one case times out) must
  still render as it does today.
- The distinction should be visible in the report header, not only in per-case detail.

## Acceptance

- [x] A run where every case of every model errors writes **no** report and **no** leaderboard rows,
      exits non-zero, and names the underlying error once.
- [x] A run where one model (or some cases) fail is unchanged from today's behaviour.
- [x] Unit tests in `tests/unit/test_benchmark_models.py` covering both, including the champion-side
[...]

PR #924, opened 2026-08-02
docs(models): the #921 tag scan — decline deepseek-v4-flash:0731 and kimi-k3 (closes #921)

### LINKED ISSUE #921: Evaluate new Ollama Cloud models: deepseek-v4-flash:0731, kimi-k3
## Context
The weekly model-health check (2026-08-02) found tags on ollama.com/api/tags that are not in `.litellm/ollama_catalog_snapshot.json`. New Ollama Cloud models ship roughly monthly and one of them may be a better fit (or a cheaper one) for a LiteLLM tier.

## New tags

- `deepseek-v4-flash:0731` — 167GB; updated 2026-07-31; family `deepseek-flash`, version 4; size tag `0731`
- `kimi-k3` — 1561GB; updated 2026-07-27; family `kimi`, version 3

## Scope
- Read each model's page (https://ollama.com/library/<name>) for its usage level and capabilities; a higher usage level costs more Ollama Cloud quota per call.
- Decide whether any belongs in a `lem-simple` / `lem-medium` / `lem-complex` deployment in `.litellm/config.yaml`. Adopting one is a config change, not a `.litellm/model_upgrades.yaml` entry (that map is retirement-only and auto-swaps).
- Declining is a valid outcome — say so on the issue so the next scan's reader knows it was considered.

## Acceptance
- A decision per tag above.
- If a model is adopted, `scripts/weekly_model_check.sh`'s tier smoke test still passes and `poetry run pytest tests/unit -q` is green.

Auto-filed by `scripts/model_health_check.py --file-issues` (issue #716). Dedup markers (do not remove): `deepseek-v4-flash:0731`, `kimi-k3`

gitchrisqueen:
## What

The weekly model-health scan (#716) found two new Ollama Cloud tags. **Both are declined** — no
change to the `.litellm/config.yaml` roster. This PR is the record of *why*, so the next scan's
reader starts from a measurement instead of the spec sheets.

| Tag | Decision | Basis |
|---|---|---|
| `deepseek-v4-flash:0731` | **Decline** | Benchmarked on all three content tiers beside the champion **and** the incumbent `deepseek-v4-flash` build. Did not beat the build we already ship on the contract rate anywhere, at the same usage level. Margins are one case wide, so this is a "did not carry the burden" decline, not a "worse build" verdict. |
| `kimi-k3` | **Decline** | HTTP **402** from Ollama Cloud — extra-usage-only, outside plan usage. Not a quality question. |

## `deepseek-v4-flash:0731`

`:0731` is a **different build**, not a re-tag: 167GB against the unversioned tag's 140GB, published
2026-07-31 vs April. So `bm-20260802-20ae40`'s measurement of `deepseek-v4-flash` (#842) said nothing
about it — and would not have been comparable across the #910 recalibration in any case. It was
measured beside the incumbent build in one run so the comparison is build-vs-build under one
calibration, which is the actual decision:

| Tier | `:0731` contract | incumbent `deepseek-v4-flash` | champion |
|---|---|---|---|
| lem-complex | 80% | **90%** | 70% (`qwen3.5:397b`) |
| lem-medium | 80% | **90%** | 90% (`gpt-oss:120b`) |
| lem-simple | 40% | 50% | 50% (`gpt-oss:20b`) |

Both builds are Medium (2), so there is no quota saving to trade against. No `recommend` on any
tier — and since every margin above is exactly one case on a ten-case suite, that is what the
decline rests on, not the size of the gap. Adoption requires beating the incumbent; a tie inside
noise is not that.

`lem-simple` is a shape mismatch rather than a near miss: a reasoning model bills chain-of-thought
against `max_tokens`, so on the three `budget_mirrors_production` cases (3 / 5 / 8 tokens, mirroring
`ai_helper.py`'s own call sites) it returns nothing at all — and both deepseek builds are **+1 usage
level** against `gpt-oss:20b` there.

## `kimi-k3`

Never benchmarked, and that is the finding. The Ollama Cloud API answers:

```
HTTP 402: this model uses extra usage only (not included plan usage) and your
extra usage balance is empty
```

Its page prices per token ($3.00 / $15.00 per 1M, $0.30 cached) instead of publishing a usage-level
pip, so `fetch_usage_level` reads it as `unknown` — which the #842 standing spend policy holds by
default. Every deployment in `.litellm/config.yaml` is served inside plan usage; adopting `kimi-k3`
would move a tier onto metered extra-usage billing. That is an owner spend decision, not a roster
refresh — and measuring it would itself require buying extra-usage balance first.

## Two runs, not one

`bm-20260802-5fff18` and `bm-20260802-b84f19` are both committed because the second **contradicts the
first by more than this decision's margins**, on identical fixtures ninety minutes apart:

- `:0731` on lem-medium: 100% → 80% contract
- `qwen3.5:397b` on lem-complex: 90% → 70%
- `gpt-oss:20b` on lem-simple: 40% → 50%

Ten cases per tier means one case is ten points, so a single-digit gap between two models is noise.
That is the ⚖️ *measurement variance* note stated at scorecard scope, and it is the strongest
argument in the file against pinning anything on one run. Written up in
`docs/model-benchmarks/README.md`.

## Testing

- `poetry run pytest tests/unit -q` → **9481 passed**.
- Config is comments-only; re-verified `champions_from_config` still resolves
  `{lem-simple: gpt-oss:20b, lem-medium: gpt-oss:120b, lem-complex: qwen3.5:397b, lem-router: gpt-oss:20b}`.
- The tier smoke test in `scripts/weekly_model_check.sh` was **not** run: it needs the live LiteLLM
  proxy, and #921's acceptance requires it only if a model is adopted. Nothing was adopted, so no
  deployment changed and no restart is owed.
- Both runs are `in-runner-judge` mode with **no judge evidence** (the runner had no proxy to reach,
  so every judge expectation renders unscored). It changes no verdict here — each of the six already
  fails a *deterministic* graded expectation, and a judge can only add a reason to reject — but a run
  that intends to *promote* something still needs one.

## Not in this PR

- `.litellm/ollama_catalog_snapshot.json` is untouched: the snapshot refresh for these tags is
  PR #920 (`auto/ollama-catalog`), and the file is machine-owned.
- Filed **#923** while working this: a harness-wide outage (broken venv, revoked key) renders as a
  full 0% quality report and appends permanent leaderboard rows. Hit for real here — a worktree
  without `openai` installed produced a complete, plausible report off zero HTTP requests. That run
  was discarded before commit; the fix is out of scope for this issue.

## Follow-up filed

**#925** — `.litellm/config.yaml` runs the *unversioned* `deepseek-v4-flash` on two tiers, so those
tiers follow the catalog's moving tag, including onto the `:0731` build this PR declines.
`scripts/model_health_check.py`'s `diff_catalog()` compares tag NAMES, so a re-point files no
evaluation issue and would swap a live tier's model unbenchmarked. The id is left unversioned here
(it is the build measured at 90%, and a pin has upkeep on every path that keys the exact id
string); #925 adds the detection that makes that safe.

Closes #921
Follow-up: #925

### DIFF
--- a/.litellm/config.yaml
+++ b/.litellm/config.yaml
@@ -60,6 +60,13 @@ model_list:
   # deepseek-v4-flash (284B MoE / 13B active, 1M context, native structured output, fast
   # no-thinking mode) is a MEDIUM usage-level model — the same quota class gpt-oss:120b and the
   # departed minimax-m2.7 sit in — so adding it is flat on quota, not a step up.
+  #
+  # Keep this id UNVERSIONED. The catalog's `deepseek-v4-flash:0731` is a DIFFERENT build (167GB vs
+  # 140GB, published 2026-07-31), and it was benchmarked against this one on 2026-08-02 for #921
+  # (`bm-20260802-b84f19`): it scored BELOW the build below on the contract rate on every tier
+  # measured — 80% vs 90% here, 80% vs 90% on lem-complex, 40% vs 50% on lem-simple — at the same
+  # Medium usage level, so there is nothing a pin would buy. Declined; see
+  # docs/model-benchmarks/README.md. Newer tag ≠ better model is the whole point of the harness.
   - model_name: lem-medium
     litellm_params:
       model: openai/deepseek-v4-flash
@@ -95,6 +102,13 @@ model_list:
       api_key: os.environ/OPENAI_API_KEY
 
   # ── COMPLEX tier: long-form, high-quality content ────────────────────────────
+  # NOTE: openai/kimi-k3 was EVALUATED and DECLINED 2026-08-02 (issue #921). It is not a quality
+  # call and it was never benchmarked: the Ollama Cloud API answers HTTP 402 for it — "this model
+  # uses extra usage only (not included plan usage)". Its page prices per token ($3.00/$15.00 per
+  # 1M) instead of publishing a usage-level pip, so the benchmark reads its level as `unknown`,
+  # which the #842 standing spend policy holds by default. Every deployment in this file is served
+  # inside plan usage; kimi-k3 would move a tier onto metered extra-usage billing, which is an
+  # owner spend decision, not a roster refresh. Re-evaluate only if the plan changes.
   # NOTE: openai/kimi-k2:1t was RETIRED 2026-06-16 (returns 410) and was removed — it was
   # the primary here, so content generation fell through to the OpenAI/Anthropic fallbacks,
   # which fail when those keys aren't set. qwen3.5:397b (OLLAMA) is now primary.
@@ -116,6 +130,8 @@ model_list:
   # group — see the note at the top of this file); the long-form quality reading, and any adoption
   # of the HIGH usage-level minimax-m3 / glm-5.2 as the quality option, is a
   # scripts/benchmark_models.py decision on #842, not a config-time one.
+  # Unversioned for the same reason as the lem-medium deployment above — the `:0731` build measured
+  # worse here too (80% contract vs this build's 90%, #921).
   - model_name: lem-complex
     litellm_params:
       model: openai/deepseek-v4-flash
--- a/.litellm/config.yaml
+++ b/.litellm/config.yaml
@@ -61,12 +61,22 @@ model_list:
   # no-thinking mode) is a MEDIUM usage-level model — the same quota class gpt-oss:120b and the
   # departed minimax-m2.7 sit in — so adding it is flat on quota, not a step up.
   #
-  # Keep this id UNVERSIONED. The catalog's `deepseek-v4-flash:0731` is a DIFFERENT build (167GB vs
-  # 140GB, published 2026-07-31), and it was benchmarked against this one on 2026-08-02 for #921
-  # (`bm-20260802-b84f19`): it scored BELOW the build below on the contract rate on every tier
-  # measured — 80% vs 90% here, 80% vs 90% on lem-complex, 40% vs 50% on lem-simple — at the same
-  # Medium usage level, so there is nothing a pin would buy. Declined; see
-  # docs/model-benchmarks/README.md. Newer tag ≠ better model is the whole point of the harness.
+  # The catalog's `deepseek-v4-flash:0731` is a DIFFERENT build (167GB vs 140GB, published
+  # 2026-07-31), and it was benchmarked beside this one on 2026-08-02 for #921
+  # (`bm-20260802-b84f19`): rejected on all three tiers, no `recommend`. It did not beat the build
+  # below — 80% vs 90% contract here, 80% vs 90% on lem-complex, 40% vs 50% on lem-simple — at the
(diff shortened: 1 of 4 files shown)


gitchrisqueen:
🔎 Claude adversarial review — FIXED 3 findings (+1 follow-up filed)

Reviewed the diff, both committed reports, `scripts/benchmark_models.py`, `scripts/model_health_check.py` and the surrounding config. The **decision** (decline both tags) is right and the two reports are internally consistent — every scorecard rate reconciles against its per-expectation list, and the `Closes #921` is honest (both acceptance criteria met; nothing adopted, so the tier smoke test is genuinely N/A). What did not hold up is what the record *claims*, which is the whole deliverable of a docs PR.

- **The `:0731` verdict was stated as settled fact on one-case margins.** `.litellm/config.yaml` said it "scored BELOW" the incumbent build and the README said it "is *below the build LEM already ships* everywhere" — while the same PR states that ten cases per tier makes one case ten points and that a one-run margin under ~20 points is noise. Every margin in that comparison is exactly one case (80/90, 80/90, 40/50), and `bm-20260802-5fff18` scored `:0731` **100%** on `lem-medium` ninety minutes earlier. Fixed: both now say it *did not beat* the incumbent, and the decline rests where it actually rests — no `recommend` on any tier, burden on the candidate, a tie inside noise is not adoption.
- **"There is nothing a pin would buy" is backwards.** The id is UNVERSIONED, so both the `lem-medium` and `lem-complex` deployments follow the catalog's moving tag — including onto the `:0731` build if ollama re-points it. And `diff_catalog()` compares tag NAME sets, so a re-point files **no** evaluation issue: a live tier would change model unbenchmarked, which is the one outcome the harness exists to prevent. Kept unversioned (it is the build measured at 90%, and a pin has upkeep on every path that keys the exact id string), but the comment now states the trade instead of denying it, and the detection gap is filed as **#925** (`agent:ready`, `priority:low`) and linked in the PR body.
- **"every case is `judge:timeout`" is not what happened.** `judgeable_cases()` never sends a deterministically-failed case to the judge, which is exactly why both reports' `Timeouts` column equals each row's first-draft pass count rather than 10. Corrected, with a line telling the next reader how to read that column. Also "each of the six verdicts" → nine across the two committed runs (verified: all nine fail a deterministic graded expectation, so the missing judge changes nothing).

Not changed, checked and fine: the leaderboard rows match `leaderboard_rows()`/`update_leaderboard()` output and ordering; committing the contradictory earlier run is the right call and is explained; `.litellm/ollama_catalog_snapshot.json` correctly left to #920 — worth noting that dedup only searches **open** issues, so if #920 has not merged before the next weekly scan, these two tags get re-filed as a fresh evaluation issue.

`/home/lem/linkedin_engagement_manager/.venv/bin/python -m pytest tests/unit -q` → **9481 passed**. Config still parses; `lem-medium`/`lem-complex` deployment ids unchanged (comments only).

gitchrisqueen:

PR #926, opened 2026-08-02
fix(benchmarks): a harness-wide outage is refused, not published as 0% (closes #923)

### LINKED ISSUE #923: benchmark: a harness-wide outage renders as a full 0% quality report and lands in the rolling leaderboard

gitchrisqueen:
## What & why

`scripts/benchmark_models.py` treated an **infrastructure** failure as a **quality measurement** and
committed it to the permanent record. `ProviderClient.complete` turns every exception into a case
result — right for a provider 500 or a rate limit, and also exactly what a broken venv, a revoked
`OLLAMA_CLOUD_API_KEY` or a DNS outage looks like. The run then completed "successfully": a full
report, `reject` on every verdict, champions at `0% (0/10)`, and one leaderboard row per model
indistinguishable from a real bad run forever after. Hit for real while working #921 — a worktree
whose venv had no `openai` produced a complete, plausible report off a run that never made an HTTP
request, caught by eye rather than by the harness. The harness runs unattended, so an expired key
would have shipped a PR asserting that every model LEM runs scores zero.

`harness_outage()` reads the tell: **every case of every model errored, the tier's own champion
included**. A roster of candidates can genuinely be bad; the incumbent that serves production scoring
zero on every single case means nothing reached a provider at all. #910 settled this for one case (a
completion that was never measured is `no output`, not a zero-length answer) — this is the same
principle at run scope.

## What changed

- **`render_report` runs `assert_measured` beside the existing provenance guard**, so the refusal
  covers `--run` and `--render` alike: no report file, **no leaderboard rows**, exit `1` (the weekly
  cron already reads any rc outside `{0, 2}` as a real failure and opens no PR), and the underlying
  error named **once** — taken from the commonest case error, with a count of any others rather than
  a wall of repeats.
- **The honest half is untouched.** One model failing every case beside a champion that answered, or
  some cases erroring everywhere, still renders and still lands on the leaderboard exactly as today.
  Those runs now carry the split in the report **header** — `- **Unmeasured cases:** 3 of 60 …`,
  naming any model that answered nothing at all — rather than only under the per-case ❌ detail.
- The results JSON carries `harness_outage` so the artifact `--render` reads later says why it was
  refused; the guard recomputes from the scorecards, so an older results file is still caught.
- `docs/model-benchmarks/README.md` states the rule beside the existing *measurement variance* /
  *no output* notes; CLAUDE.md's #721 paragraph gained the one-line invariant.

## Testing

`poetry run pytest tests/unit -q` → **9496 passed**, of which 202 in
`tests/unit/test_benchmark_models.py` (15 new):

- the outage itself (candidate + champion, every case) and the **champion-side tell** — a 0%
  champion is what says it was the harness, not the roster;
- the underlying error named once, not per case; a second distinct error counted, not listed;
- both honest-half shapes (one model down, some cases down) still render **and** still produce
  leaderboard rows;
- a dry run of canned outputs is never an outage; a run with no provider configured never publishes;
- `write_report` leaves a pre-existing leaderboard file byte-identical and creates no report;
- the header line (count + silent model) and its absence on a clean run;
- the CLI `--render` path exits `1`, writes nothing, and prints the cause once.

Closes #923

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -230,8 +230,11 @@ every check is classed by what production does with its failure: `contract` (the
 it — the ABSOLUTE floor) vs `repairable` (a regeneration gate retries it — advisory, still
 champion-relative). A case whose `max_tokens` mirrors a call site (`budget_mirrors_production`) is
 never retried at a bigger budget; an EMPTY completion is re-measured at the same one, and a still-
-empty one is no output, never a 0-char answer. Only `recommend` becomes a swap; recommendations are
-RENDERED, never written (`.litellm/model_upgrades.yaml` is the retirement map). Full posture:
+empty one is no output, never a 0-char answer. Same principle at RUN scope (#923): a run where every
+case of every model errored — champion included — is a harness outage, not a scorecard of zeros, and
+is REFUSED (no report, no leaderboard rows, exit 1, the cause named once); a partial failure still
+publishes and carries an `Unmeasured cases` count in the report header. Only `recommend` becomes a
+swap; recommendations are RENDERED, never written (`.litellm/model_upgrades.yaml` is the retirement map). Full posture:
 `docs/model-benchmarks/README.md`.
 
 ### Content-quality telemetry (issue #630)
--- a/docs/model-benchmarks/README.md
+++ b/docs/model-benchmarks/README.md
@@ -136,6 +136,35 @@ headroom, so it applies to production-budget cases too. Two things follow:
   measured. No output fails the gate's "provider answered every case" expectation instead, which is
   what it actually was.
 
+### A harness outage is not a run of zeros (#923)
+
+The rule above at run scope. A provider error is a legitimate case result — a 500, a rate limit, a
+retired tag — and `ProviderClient.complete` turns any exception into one. That is also what a broken
+venv, a revoked `OLLAMA_CLOUD_API_KEY` or a DNS failure looks like, so a harness that could not call
+anything still *completed*: a full report, `reject` on every verdict, champions at `0% (0/10)`, and
+one leaderboard row per model that is indistinguishable from a real bad run forever after. It
+happened for real while working #921 — a worktree whose venv had no `openai` produced exactly that,
+off a run that never made an HTTP request, and it was caught by eye rather than by the harness. The
+harness runs unattended (`scripts/weekly_model_check.sh` opens the report as a PR), so an expired key
+would ship a PR asserting that every model LEM runs scores zero.
+
+The tell is **every case of every model, the tier's own champion included**. A roster of candidates
+can genuinely be bad; the incumbent that serves production scoring zero on every single case means
+nothing reached a provider at all. On that condition the run is **refused**, not published:
+
+- no per-run report is written and **no leaderboard rows** are appended — nothing measured, nothing
+  to record;
+- the script exits **1**, which the weekly cron already reads as a real failure (only `0` and `2`
+  open the PR), so the outage alerts instead of shipping;
+- the underlying error is named **once** (`refusing to render: harness outage: …`), taken from the
+  commonest case error, with a count of any others rather than a wall of repeats.
+
+The honest half is untouched: one model failing every case beside a champion that answered, or some
+cases timing out everywhere, is a real measurement and renders exactly as before. Those runs now
+carry the split in the report **header** — `**Unmeasured cases:** 3 of 60`, naming any model that
+answered nothing at all — rather than only under the per-case ❌ details, because how much of a run
+is a measurement at all decides whether the rest of it means anything.
+
 ## The gate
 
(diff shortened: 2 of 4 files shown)


gitchrisqueen:

Read the full diff, `merge_scorecard` / `ProviderClient.complete` / `render_report` / `main` around it, `scripts/weekly_model_check.sh`, and the committed leaderboard. Verified the load-bearing claims: `errors` really is one entry per errored case (so `errored == total` is the right test), `write_report` renders before `os.makedirs` (so an outage touches neither the report nor `README.md`), `write_report` is the ONLY leaderboard writer, both `main` paths return `1`, and the cron's `case "$BRC" in 0|2)` genuinely opens no PR on `1` — it alerts instead. Committed leaderboard carries no 0% outage rows, so nothing needed scrubbing. All four acceptance criteria are covered; no later phase is described in #923, so `Closes #923` is honest.

- **`Unmeasured cases` header named a silent model once per tier** (`render_report`): the list was per-scorecard, not per-model. The weekly cron runs 3–4 tiers against the same candidate, so a candidate that answered nothing anywhere was about to render as ``​`cand`, `cand`, `cand`, `cand` answered nothing at all`` — four dead models instead of one, in the header line this issue exists to add. Deduped + sorted, matching the outage record's own `champions` list. Test: a model silent across two tiers appears once.
- **The champion-less outage was unpinned and read as excluded** (`harness_outage`): the docstring, `docs/model-benchmarks/README.md` and CLAUDE.md all state the tell as "champion included", but the code — correctly — refuses on `errored == total` whether or not a champion was measured. A tier with no configured incumbent still measured nothing, and requiring a champion card there would let exactly that run publish its zeros. Behaviour unchanged; stated as *tell, not precondition* in the docstring and the doc, and locked with a test asserting the run is refused with an empty `champions` list and no champion clause in the message.

Noted, not changed: `ProviderClient.complete` reports a #910 empty completion as an `error`, so a run where every case of every model came back empty-but-successful is also refused as an "outage". That is the right call under #910's own principle (a completion that was never measured is no output), the real cause is carried verbatim in `Underlying error:`, and the shape needs a champion that answers nothing on every case — not worth widening the guard for.

`tests/unit/test_benchmark_models.py` 204 passed (+2), `test_litellm_roster_config.py` green.

gitchrisqueen:

PR #927, opened 2026-08-02
feat(models): the catalog scan sees a build swapped under an unchanged tag (closes #925)

### LINKED ISSUE #925: Catalog scan can't see a re-pointed unversioned Ollama tag — a build swap under an existing id ships unbenchmarked
## Context

`.litellm/config.yaml` runs two deployments (`lem-medium`, `lem-complex`) on the **unversioned** id
`openai/deepseek-v4-flash`. An unversioned Ollama id follows whatever the catalog's moving tag
points at, so the build serving production can change with no repo change at all.

The weekly catalog scan cannot see that happen. `diff_catalog()` in `scripts/model_health_check.py`
compares **tag NAME sets** only:

```python
def diff_catalog(snapshot: dict, current: dict) -> dict:
    old, new = set(snapshot or {}), set(current or {})
```

so a re-point of an existing name — same key, new `size` / `modified_at` — produces no `added`
entry and therefore **no `agent:ready` evaluation issue**. `snapshot_changed` (a whole-dict compare)
does fire, so the change lands as a couple of edited numbers inside the machine-owned
`.litellm/ollama_catalog_snapshot.json` refresh PR, which is exactly the diff nobody reads line by
line.

This is not hypothetical for the current roster. #921 measured `deepseek-v4-flash:0731` (167GB,
2026-07-31) beside the incumbent 140GB build and it did **not** beat it on any of the three content
tiers (`bm-20260802-b84f19`: 80% vs 90% contract on `lem-complex` and `lem-medium`, 40% vs 50% on
`lem-simple`). If ollama.com later moves the bare `deepseek-v4-flash` tag onto that build, both LEM
tiers adopt it silently and unbenchmarked — the one outcome the harness exists to prevent.

## Scope

- Detect a **changed build under an unchanged tag name** in the catalog scan: compare `size` /
  `modified_at` (or a digest, if `ollama.com/api/tags` exposes one) for tags that are present in
  both the snapshot and the live catalog.
- Report it only for tags that are actually **referenced by `.litellm/config.yaml`** — the catalog
  carries hundreds of tags LEM does not run, and a re-point of one of those is noise.
- Route it the way a new tag is routed: ONE consolidated `agent:ready` evaluation issue (reuse
  `plan_issue` / the existing dedup-marker mechanic; the marker should include the new
  size/timestamp so a *second* re-point is not deduped away as already-filed).
- Say in the issue body what the config deployment is, so the reader knows a live tier moved.
- Do NOT auto-swap or auto-pin anything — this is an evaluation trigger, same as the new-tag path.

## Acceptance

- [x] A tag present in both snapshot and catalog whose build changed produces an evaluation issue.
- [x] Only config-referenced tags do; an unreferenced tag's re-point stays silent.
- [x] A re-point of the same tag twice files/appends twice rather than deduping to the first.
- [x] Unit tests in `tests/unit/test_model_health_check.py` cover all three.
- [x] `poetry run pytest tests/unit -q` green.

Found while adversarially reviewing PR #924 (the #921 tag scan), which decided to leave
`deepseek-v4-flash` unversioned. That decision is fine on today's evidence; this issue is the
detection that makes it safe to keep.

gitchrisqueen:
## What

The weekly catalog scan can now see a **build swapped underneath an unchanged tag name**.

`.litellm/config.yaml` runs the *unversioned* `openai/deepseek-v4-flash` on both `lem-medium` and
`lem-complex`, so those tiers follow whatever ollama.com's moving tag points at. `diff_catalog()`
compares tag NAME sets, so a re-point of an existing name produced no `added` entry and therefore no
`agent:ready` evaluation issue — the change landed only as a couple of edited numbers inside the
machine-owned `.litellm/ollama_catalog_snapshot.json` refresh PR, which is the diff nobody reads line
by line. #921 measured `deepseek-v4-flash:0731` and it beat the incumbent on nothing; if the bare tag
ever moves onto that build, both tiers adopt it silently and unbenchmarked.

## How

- **`plan_repoints(deployments, snapshot, catalog)`** — tags present in **both** the snapshot and the
  live catalog whose `size` / `modified_at` moved. Restricted to tags `.litellm/config.yaml` actually
  runs on Ollama Cloud; the catalog carries hundreds LEM never serves and a re-point of one of those
  is noise. `diff_catalog` is left alone (name sets are its job) with a docstring pointing here.
- **A field only counts when BOTH sides carry a value.** A snapshot written without a field has no
  baseline for it — reading that as a build swap would file an issue for every configured tag at once
  the day the snapshot shape changes. A seeded (missing) snapshot reports nothing, same as new tags.
- **Routed exactly like a new tag**: ONE consolidated `agent:ready` issue via the existing
  `plan_issue` dedup mechanic, on its own title prefix (`REPOINT_TITLE_PREFIX`) since a re-point is
  not a new model. The body names the tiers that moved and the exact config `model:` line.
- **The dedup marker carries the new build's fingerprint** (`tag @ size/modified_at`), so a *second*
  re-point of the same tag appends to the open issue instead of deduping to the first.
- **Never an auto-swap or auto-pin** — an evaluation trigger, same as the new-tag path. The issue
  body also says the benchmark candidate is the *versioned* build id, because benchmarking the bare
  tag would run the same id as the champion and compare nothing.
- `_file_issues` reads specs through `_issue_spec()`, so a `--plan-file` written by an older copy of
  the tool files the kinds it does have rather than raising `KeyError`.
- `scripts/weekly_model_check.sh` counts re-points into the gate that runs `--file-issues` (and logs
  `re-pointed-tags=N`), so a re-point-only week still files. Exit code 2 covers it too.

Rendered against the live config with `deepseek-v4-flash` re-pointed to the 0731 build, the issue
reads: *serving `lem-complex`, `lem-medium` (config `model: openai/deepseek-v4-flash`) — size
`140GB` → `167GB`, modified_at `2026-04-24…` → `2026-07-31…`*.

## Testing

- `tests/unit/test_model_health_check.py`: `TestPlanRepoints` (size move, timestamp-only move,
  unreferenced tag stays silent, unchanged build, added/removed tags are not re-points, missing
  baseline, non-Ollama deployment, every tier named once), `TestRepointIssue` (marker fingerprint,
  the second re-point appends rather than dedupes, body names the live deployment), plus CLI
  end-to-end cases for report + `--file-issues` from a saved plan, an unreferenced re-point staying
  quiet, and the seeded-snapshot and committed-repo-state guards asserting no `REPOINTED`.
- `poetry run pytest tests/unit -q` — **9513 passed**.
- `bash -n scripts/weekly_model_check.sh` clean.

Closes #925

### DIFF
--- a/docs/model-benchmarks/README.md
+++ b/docs/model-benchmarks/README.md
@@ -9,7 +9,8 @@ this file (`<date>-<run_id>.md`).
 ## Why this exists
 
 `scripts/model_health_check.py` (#716) detects that a model is *about to change*: a published
-retirement, a new Ollama Cloud tag, a family we trail. It cannot say whether a candidate is any
+retirement, a new Ollama Cloud tag, a family we trail, or — since #925 — a tag we already run whose
+*build* moved underneath an unchanged name. It cannot say whether a candidate is any
 **good** — before this, a swap was decided on a spec sheet. The benchmark is the measurement half,
 and it always runs the current **champion** alongside the candidate, so a verdict is a comparison
 rather than an absolute claim about one model.
@@ -309,11 +310,14 @@ Four things this run is worth reading for beyond the two verdicts:
   one case, which is why that decline rests on "no `recommend`" rather than on the gap's size.
 - **Declining `:0731` is not the same as being safe from it.** `.litellm/config.yaml` runs the
   *unversioned* `deepseek-v4-flash` on two tiers, so those tiers follow whatever the catalog's
-  moving tag points at — and `scripts/model_health_check.py` diffs tag NAMES, so a re-point of that
-  name onto the 0731 build files no evaluation issue and swaps a live tier's model unbenchmarked.
-  The id was left unversioned anyway (it is the build measured at 90%, and a pin has upkeep of its
-  own on every path that keys the exact id string), but the detection gap is real and is tracked on
-  **#925**.
+  moving tag points at — and `scripts/model_health_check.py` diffed tag NAMES only, so a re-point of
+  that name onto the 0731 build filed no evaluation issue and swapped a live tier's model
+  unbenchmarked. The id was left unversioned anyway (it is the build measured at 90%, and a pin has
+  upkeep of its own on every path that keys the exact id string); **#925** closed the detection gap
+  instead — the weekly scan now also compares each CONFIGURED tag's `size`/`modified_at` against the
+  committed snapshot and files a re-point evaluation issue naming the tiers that moved. The dedup
+  marker carries the new build's fingerprint, so a *second* re-point is not deduped away as
+  already-filed. It stays a trigger, never an auto-pin.
 
 Both runs are `in-runner-judge` mode with **no judge evidence** — the runner had no LiteLLM proxy to
 reach, so the judge answered nothing and every judge expectation renders as unscored. Read the
--- a/scripts/model_health_check.py
+++ b/scripts/model_health_check.py
@@ -15,6 +15,10 @@ docs.ollama.com/cloud.md and ollama.com/api/tags and reports what is about to ch
     replacement to model_upgrades.yaml (the reactive half then swaps it before the sunset bites);
   * new/removed catalog tags vs the committed snapshot (.litellm/ollama_catalog_snapshot.json) ->
     ONE `agent:ready` evaluation issue for the new ones, snapshot refreshed in the same PR;
+  * a CONFIGURED tag whose BUILD moved under an unchanged name (issue #925) -> ONE `agent:ready`
+    evaluation issue. An unversioned id like `deepseek-v4-flash` follows whatever the catalog's
+    moving tag points at, so the build serving a live tier can change with no repo change at all —
+    and a tag-NAME diff cannot see that;
   * a configured model trailing a newer version of its OWN family (minimax-m2.7 while minimax-m3
     ships) -> ONE consolidated `agent:ready` upgrade issue carrying each candidate's usage level,
     because a version bump that jumps usage tiers costs more quota and is an evaluation, never an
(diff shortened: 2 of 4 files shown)


gitchrisqueen:

- **The filed issue steered the reader into a benchmark that compares nothing** (`build_repoint_issue_body`): the Scope said "the benchmark candidate is the VERSIONED build id, not this bare tag: measuring the bare tag would run the same id as the champion and compare nothing" — but the bare tag *is* the champion. `scripts/benchmark_models.py` reads champions off `.litellm/config.yaml` (`--champions` overrides, default = config), and after a re-point that config line already resolves to the NEW build. Running `--models deepseek-v4-flash:0731` therefore measures the new build against the new build: every `beats champion (contract)` / `(first-draft)` expectation ties for free and the gate can hand back a `recommend` that rests on nothing — the exact degenerate comparison the bullet was warning about, one level up. Fixed: the body now says both sides have to be pinned and gives the flags (`--models <new-versioned-id> --champions <tier>=<previous-versioned-id>`), with the previously measured build identified by the `old` size/`modified_at` already listed above it. New test `test_body_pins_the_champion_side_not_only_the_candidate`.

Checked and found clean:
- `plan_repoints` seams — added/removed tags are excluded (both-sides membership), non-Ollama and unreferenced tags excluded, `is_ollama` keys on the same `OLLAMA_CLOUD_URL` marker every other path uses, so it fires on the real box config.
- The missing-baseline guard is right for the live snapshot: `glm-5.2` and `minimax-m3` carry `size: 0` today, so only `modified_at` can fire for them — no per-tag false positive, and a seeded snapshot reports nothing.
- Dedup: markers carry the fingerprint, so a second re-point appends whether or not the snapshot-refresh PR merged in between; a re-detected *same* re-point is `none`, so no weekly spam. `REPOINT_TITLE_PREFIX` doesn't collide with the eval/upgrade prefixes.
- `_issue_spec` covers the `--plan-file` KeyError path for all three kinds; `_print_catalog_plan` and the exit-2 branch use `.get("repoints")`, so an older saved plan still files what it has.
- Shell: verified the 5-field `read` parse (new + old plan JSON), the `|| echo "0 0 0 0 0"` fallback and `bash -n`.
- Rendered the issue against the real `.litellm/config.yaml` + committed snapshot with `deepseek-v4-flash` moved to the 0731 build — names `lem-complex`, `lem-medium`, `model: openai/deepseek-v4-flash`, `140GB → 167GB`.
- Close is honest: every acceptance box on #925 is implemented, no later phase declared.
- `poetry run pytest tests/unit -q` — 9514 passed.

gitchrisqueen:

Issue #928, opened 2026-08-02
Live-ground the group-feed comment composer — Phase 2 (follow-up of #916)


gitchrisqueen:
## Why

#916 fixed the escalating `Selector miss: Comment composer` warning (408 occurrences in 18h, every
one on a `/groups/` URL) two ways: it widened the composer lookup from the post card to the widest
scope that still provably holds ONE post (`_single_post_scope`), and it downgraded a total miss to
DEBUG, because "no composer opened" means we skip the post and release the claim.

The downgrade is correct — but it means the group commenting lane is now **quiet** whether or not
the widening actually works. The widening was derived from the DOM contract in
`docs/sdui-selenium-notes.md` and from the PostHog log evidence, **not** from a live group feed, so
it needs a grounding pass exactly like the #816 feed run did. If it did not work, group commenting
posts nothing and burns one `lem-medium` comment generation per examined post, silently.

Telemetry that answers it without a browser: `feed_scan` events with `feed_sort = "n/a"` are the
group scans. Before #916, 3 days of them examined 2,515 posts and landed 1 comment.

## Scope

- Run `scripts/linkedin_live_validation.py` against a group feed (and the home feed, as the control)
  and record per-locator hit counts the way `_COMMENT_ACTION_LOCATORS` / `_REACTION_TRIGGER_LOCATORS`
  already do.
- Confirm whether `_post_composer_for_card`'s widened scope resolves the group composer. If it does
  not, record the real group-card DOM shape and fix the resolver against it.
- If a group card genuinely cannot be commented on inline, stop `auto_comment_in_groups` from
  spending an LLM call per post it can never post (`_engage_card` generates BEFORE
  `post_comment_inline`).

## Files

- `src/cqc_lem/app/run_automation.py` (`_post_composer_for_card`, `_single_post_scope`,
  `_COMMENT_ACTION_JS`, `auto_comment_in_groups`)
- `scripts/linkedin_live_validation.py`
- `docs/sdui-selenium-notes.md`

## Acceptance

- [ ] A live group-feed run records whether the composer resolves, with per-locator counts in the doc.
- [ ] Group commenting either lands comments, or stops spending LLM calls on posts it cannot comment on.
- [ ] `feed_scan` for group scans shows a non-zero comment rate, or the lane is explicitly disabled.

Follow-up of #916.

gitchrisqueen:
[...]

PR #929, opened 2026-08-02
fix(engagement): resolve a post's comment composer beside the card, not only inside it (closes #916)

### LINKED ISSUE #916: fix(errors): RecurringWarning: Selector miss: Comment composer (2x)
## Why

> Selector miss: Comment composer

- Occurrences (last 24h): **2** across 1 distinct actor(s)
- First seen: `2026-08-01T17:26:07.171000Z` · last seen: `2026-08-02T00:35:05.377000Z`
- SDK: `posthog-python`

[Open the issue in PostHog](https://us.posthog.com/project/475262/error_tracking/019fbe5c-9948-7870-9649-34a11d6579cf) — it has the stack trace, the grouped occurrences and the affected people.

## Scope

## Files
- `tests/unit/`

## Acceptance
- All required CI gates pass.

Auto-filed from PostHog Error Tracking. Dedup marker (do not remove): `posthog-issue-019fbe5c-9948-7870-9649-34a11d6579cf`

gitchrisqueen:
## What

`Selector miss: Comment composer` fired **408 times in 18 hours**, escalated to ERROR by the
recurrence rule, and filed a grouped `$exception`. Every one of those 408 carried a `/groups/` URL
(`attributes.url` on the PostHog log rows) — not one came from any other surface.

It was not noise. `_card_for_textbox` returns only the **nearest** ancestor carrying the post's
comment action, and where LinkedIn renders the comment section *beside* that node instead of inside
it, the card-scoped `find_first` could never see the composer the Comment click had just mounted.
The post was then skipped — after the `lem-medium` call that wrote the comment had already been paid
for — and ~35s of Chrome time burned per card (`WAIT_DEFAULT_TIMEOUT` x `MAX_WAIT_RETRY + 1`).

## How

`_post_composer_for_card` is now the ONE resolver for a post card's composer, shaped like the
existing `_reply_composer_for_comment`:

- a rendered `role=textbox` **inside** the card still wins (#876 unchanged);
- otherwise `_single_post_scope` widens to the widest ancestor that still covers **this post
  alone** — so the search area provably covers this post and no neighbour, which is the invariant
  #876 actually needs. Borrowing the post-above's still-mounted composer stays impossible. The
  bound counts per-post **markers** (`_POST_MARKER_SELECTORS`: the post-text node the feed
  enumerates cards from, plus `Hide post by`) against the card's own baseline, never comment
  ACTIONS — the composer being searched for brings its own submit button whose text is literally
  `Comment`, so an action count would see two the moment the comment section is a sibling and would
  never widen at all;
- a box starting **above** the card is rejected outright (the share box, or one left open on a post
  we already did) — the same hard above-filter the reply resolver uses;
- a box labelled `creating comment` beats an unlabelled one, so a reply box under someone else's
  comment can't win and answer a stranger instead of the author;
- no box of ours = skip the post. Still no page-wide fallback.

A miss is now **DEBUG inside the resolver**, once, the way the reply resolver already logs its own:
no composer means we skip the post and the caller releases the claim, which is an expected no-op —
and `utilities/CLAUDE.md` is explicit that warning on one costs a triage for working behaviour. It
also polls ~6s instead of ~35s per dead card.

## Testing

`tests/unit/app/test_comment_inline.py` — 48 tests, `tests/unit` — 9522 passed.

New coverage: a nested composer still wins (and the wider scope is never even asked for); the
composer mounted beside the card is found; a composer above the card is rejected; the post's own box
beats a reply box; a total miss is DEBUG and **never** `log_warning` (the regression this issue is);
a stale card gives up immediately with one message, not one per poll; a late-mounting composer is
still caught; the call site has no second, unscoped lookup.

## Notes

The widening is derived from the DOM contract in `docs/sdui-selenium-notes.md` and the PostHog log
evidence, not from a live group feed — and the DEBUG downgrade makes the lane quiet either way. That
verification is tracked, not left in prose.

Follow-up: #928 (live grounding pass; `needs-human`, it needs a real Selenium session).

Closes #916


gitchrisqueen:

- **The widening guard could never widen (silently), in exactly the render it was written for** — `_SINGLE_POST_SCOPE_JS` kept an ancestor only while it held **exactly one `isCommentAction` button**. But the composer this widens to *find* brings its own submit button, and that button's text is literally `Comment`: `_SUBMIT_NEAR_COMPOSER_JS` clicks precisely that (`innerText` in `['comment','post','reply']`, aria-label not `comment`/`reply`), and its `!x.disabled && offsetParent !== null` filter is in-repo evidence that such buttons sit in the DOM disabled/hidden before anything is typed. `isCommentAction` matches on `textContent` with no visibility or disabled test, so the first ancestor holding the card **and** the sibling comment section counted **two** → `break` → `scope = card` → the same miss as before. With the miss now DEBUG, that no-op would never have surfaced again.
  **Fixed** (`24ceaf9a`): the walk is bounded on per-post **markers** — `_POST_MARKER_SELECTORS` = the post-text node the feed enumerates cards from (`_FEED_POST_TEXT_SEL`) plus the card's own `Hide post by` control (an image-only post has no text node but still has that) — kept only while every count equals the **card's baseline**, not a hardcoded 1, so a reshare rendering two text nodes inside one card still widens. Crossing into a neighbouring post raises a count and stops the walk, so #876's "never borrow a neighbour's box" invariant is unchanged. JS behaviour verified against a stubbed DOM (widens past a sibling comment section, stops at a neighbour, handles the reshare baseline); docs + PR body corrected to describe the shipped bound.

Checked and found sound: no page-wide fallback anywhere on the path (`find_first` is no longer reachable from `post_comment_inline`); the hard above-filter and the `creating comment` label preference match `_reply_composer_for_comment`'s proven shape; a stale card exits on the first poll with one message; DEBUG-not-warning is right per `utilities/CLAUDE.md` (the caller releases the claim, so a miss is an expected no-op); poll budget (6s) stays well under the old ~35s and under `WAIT_DEFAULT_TIMEOUT`; the `Closes #916` is honest — every acceptance box is met and the live grounding that is genuinely out of reach here is filed and linked as #928.

Tests: `tests/unit/app/test_comment_inline.py` 49 passed (new: the bound must not key on a comment-action count); `tests/unit/app` 1903 passed.

gitchrisqueen:

PR #931, opened 2026-08-02
fix(errors): a stored automation pause is INFO, not a warning that files a defect (closes #917)

### LINKED ISSUE #917: fix(errors): RecurringWarning: Automation PAUSED for 1800s (reason: deploy) (1x)
## Why

> Automation PAUSED for 1800s (reason: deploy)

- First seen: `2026-08-01T11:34:10.102000Z` · last seen: `2026-08-01T11:34:10.102000Z`
- SDK: `posthog-python`

[Open the issue in PostHog](https://us.posthog.com/project/475262/error_tracking/019fbd1a-74c8-73a2-8154-06b60d1988a8) — it has the stack trace, the grouped occurrences and the affected people.

## Scope

## Files
- `tests/unit/`

## Acceptance
- All required CI gates pass.

Auto-filed from PostHog Error Tracking. Dedup marker (do not remove): `posthog-issue-019fbd1a-74c8-73a2-8154-06b60d1988a8`

gitchrisqueen:
## What & why

PostHog filed `RecurringWarning: Automation PAUSED for 1800s (reason: deploy)` — a grouped
`$exception` for behaviour that is working exactly as designed.

`pause_automation()` in `utilities/linkedin/rate_limit.py` emitted a `log_warning` every time it
successfully stored the global Selenium kill-switch. Deploy maintenance mode (`utilities/maintenance.py`,
`MAINTENANCE_REASON = "deploy"`) sets one on **every** release, and releases batch 4x daily — so the
recurrence escalation (3 occurrences in 24h) re-emitted the warning at ERROR and turned a routine
deploy step into a defect.

**Root cause is the level, not the pause.** Succeeding at what the caller asked for is never a
degraded path detected *here* — it is a deliberate state transition, which the level table puts at
INFO. Nothing is lost by dropping the warning, because every caller for which a pause *is* the defect
already says so where it detects it:

| Caller | Reason | Its own signal |
|---|---|---|
| `maintenance.begin_maintenance` | `deploy` | `log_info` "Maintenance mode ON for Ns — paused dispatch, cancelled N consumer(s)" |
| `run_scheduler.auto_suppression_tripwire` | `suppression:<uid>` | `log_critical` "Suppression tripwire fired…" (#629) |
| `POST /admin/pause-automation` | `admin` | the operator pressed the button and gets the response |
| 429 breaker | — | doesn't call this; warns in `mark_rate_limited` |

So the warning was a second defect filed for someone else's already-reported fault — the
"warn where you detect, not where you notice" rule from #878, read sideways for a state-setter.

### Changes
- `pause_automation` logs the stored pause at **INFO**. The Redis-failure path (a kill-switch that
  did **not** take) keeps its `log_warning` — that one is a genuine degraded path.
- It now logs the reason it *actually stored*: an empty `reason` falls back to `"manual"` for the
  Redis key, but the message used to print `(reason: )`.
- Documented the state-setter variant of the escalation rule in `docs/error-tracking.md` and
  `src/cqc_lem/utilities/CLAUDE.md`, so the next call site doesn't repeat it.

No behaviour change to the pause itself: same key, same TTL, same return value, same fail-open.

## Testing

`tests/unit/utilities/linkedin/test_rate_limit.py` — three new cases:
- a stored pause logs INFO and does **not** warn (the #917 regression guard);
- an empty reason logs the `manual` it stored, matching the Redis value;
- a Redis error on `set` still warns and does not log the INFO.

```
tests/unit/utilities/linkedin/test_rate_limit.py + test_maintenance.py … 93 passed
tests/unit/app/test_suppression_tripwire.py + api suppression/admin … 138 passed
tests/unit/utilities … 5837 passed
```

Closes #917

### DIFF
--- a/docs/error-tracking.md
+++ b/docs/error-tracking.md
@@ -109,6 +109,14 @@ The second half of that pattern is **warn where you detect, not where you notice
 failure one frame up files a second defect for one fault, so that caller now logs the outcome at
 DEBUG and each failing path inside `react_to_post_inline` owns its single warning (issue #878).
 
+The same rule reads sideways for a **state-setter**: succeeding at what you were asked to do is never
+a degraded path, however serious the state is. `pause_automation` warned whenever it stored the global
+Selenium kill-switch, and maintenance mode sets one on every release — 4x daily — so a routine deploy
+escalated to ERROR and filed `RecurringWarning: Automation PAUSED for 1800s (reason: deploy)`
+(issue #917). It logs INFO now; the callers for which a pause IS the defect already say so where they
+detect it (the suppression tripwire escalates CRITICAL, the 429 breaker warns in `mark_rate_limited`),
+and only failing to store the pause — a kill-switch that didn't take — still warns.
+
 | Env | Default | Purpose |
 |---|---|---|
 | `LOG_ESCALATION_ENABLED` | `true` | master switch; false → zero Redis calls |
--- a/src/cqc_lem/utilities/CLAUDE.md
+++ b/src/cqc_lem/utilities/CLAUDE.md
@@ -70,6 +70,13 @@ Two things follow for anyone writing a call site here:
    where you detect, not where you notice.** ONE unreadable card still warns twice, because the
    pre-click 'Reaction state' read warns too (issue #874, open); collapse that into the opener's
    signal as well, and never add a fourth.
+   The rule reads sideways for a **state-setter**: doing what you were asked is not a degraded path,
+   however serious the state is. `pause_automation` warned every time it stored the global Selenium
+   kill-switch, and maintenance mode sets one on EVERY release, so a routine deploy filed
+   `RecurringWarning: Automation PAUSED for 1800s (reason: deploy)` (issue #917). It logs INFO now —
+   the callers for which a pause IS the defect already say so where they detect it (suppression
+   escalates CRITICAL, the 429 breaker warns in `mark_rate_limited`) — and only a kill-switch that
+   FAILED to store still warns.
 2. **Keep the message a stable template.** The dedup key masks volatile tokens (URLs, emails, UUIDs,
    URNs, hex, `[...]`, quoted strings, numbers) and combines them with the call site, so
    `Selector miss: Feed sort control` and `Selector miss: Reaction state` stay two distinct problems
--- a/src/cqc_lem/utilities/linkedin/rate_limit.py
+++ b/src/cqc_lem/utilities/linkedin/rate_limit.py
@@ -13,7 +13,7 @@ import json
 import os
 from datetime import datetime, timezone
 
-from cqc_lem.utilities.logger import log_warning
+from cqc_lem.utilities.logger import log_info, log_warning
 
 _COOLDOWN_KEY = "linkedin:429_cooldown"
 _TRIP_COUNT_KEY = "linkedin:429_trip_count"   # consecutive trips → escalating cooldown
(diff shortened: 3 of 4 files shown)


gitchrisqueen:

Reviewed the full diff at `af39867c` plus the surrounding seams. Nothing I would block a human PR for; no code changes pushed.

**What I verified (not just read):**
- **Every `pause_automation` call site**, against the PR's claim that no signal is lost:
  - `maintenance.begin_maintenance` (`maintenance.py:170`) → its own `log_info` "Maintenance mode ON for Ns — paused dispatch, cancelled N queue consumer(s)". ✔
  - `auto_suppression_tripwire` **new trip** (`run_scheduler.py:571`) → `log_critical` immediately after. ✔
  - `auto_suppression_tripwire` **daily re-arm** (`run_scheduler.py:567`) — the path the PR body does *not* call out, and the one that previously produced a *second* recurring warning every day a trip stood. It is not silent: `track_suppression_check(user_id, verdict, paused=True, rearmed=True)` fires on that branch (`run_scheduler.py:568`), plus the stored trip record and the Account banner. Downgrading it is the correct read of "expected no-op → never a warning", not a lost signal.
  - `POST /admin/automation-pause` (`api/main.py:5751`) → synchronous 200 to the operator who pressed it, and `GET /admin/automation-status` reads the live state. ✔
  - 429 breaker does **not** route through here — `mark_rate_limited` sets its own `_COOLDOWN_KEY` and keeps its `log_warning`. ✔
- **Escalation actually stops.** `log_escalation.note/escalate` is wired only into `log_warning` (`escalated_from: "WARNING"`), so INFO mints no `$exception`. The fix removes the cause rather than masking it with `LOG_ESCALATE_EXCLUDE` — the right lever.
- **No downstream consumer of the old WARNING.** Grepped `docs/`, `scripts/` (incl. `posthog_provision.py`) and the SPA: nothing keys an alert, dashboard tile or assertion on `Automation PAUSED`. The only remaining references are the two doc notes this PR adds.
- **Existing tests still hold.** `tests/unit/utilities/test_maintenance.py` and `test_suppression_tripwire.py` patch `pause_automation` itself; `tests/integration/test_suppression_api.py:131,148` call it for real but assert on Redis state, not logs. No caplog/level assertions anywhere on this path.
- **The reason fix is real, not cosmetic.** `stored_reason = reason or "manual"` now makes the message match the value written to `linkedin:automation_paused`, which `automation_pause_reason()` / `is_suppression_pause()` read to decide whether a deploy may lift a pause — the log and the branch logic can no longer disagree.
- **Close is honest.** #917's three acceptance criteria (exception gone / unit test on the path / CI gates) are all covered by this diff; the issue declares no later phase and has no unchecked boxes, so `Closes #917` is correct.
- Convention compliance: structured logger with `action_type`, no `print`, no raw SQL, no new migration, type hints unchanged. `codecov/patch` green; `Unit Tests (Python 3.12)` green on this head.

**Considered and deliberately not raised:** the message still prints `int(seconds)` while the TTL is `max(1, int(seconds))`, so `pause_automation(0)` logs "PAUSED for 0s" for a 1-second pause. Every real caller floors its own value (`maintenance` `max(1, …)`, admin `or 3600`, `pause_seconds()`), the divergence is at most one second, and it predates this PR — a nit, not a defect.

Local `poetry run pytest` could not run in this worktree (no venv provisioned — `Command not found: pytest`); CI is the source of truth and Unit Tests passed on `af39867c`. Integration Tests + Run Tests were still pending at review time; the runner re-checks before merge.

gitchrisqueen:

PR #934, opened 2026-08-02
fix(auth): show whether the LinkedIn sign-in approval landed (closes #933)

### LINKED ISSUE #933: fix(auth): Account approval status not visible in SPA backend
## Why
User reports that their account is already approved but they cannot see a way to confirm this in the SPA backend.

Auto-filed from in-app feedback #31 by the feedback→auto-work loop (`docs/launch-and-marketing-plan.md` §B.2/B.3). Classifier: bug/high, risk `none`, confidence 0.80.

## Scope
- Address the reported bug in the `auth` area of the product.

## Files
- `src/cqc_lem/api/main.py`
- `src/cqc_lem/utilities/linkedin/token_refresh.py`
- `tests/unit/`

## Acceptance
- All required CI gates pass.

gitchrisqueen:
## What & why

The reporter said their account was **already approved** but there was no way to confirm it in the app.

The approval in question is LinkedIn's device-verification challenge — *"Did you just try to sign in?"* — which LEM hits when it signs in on the user's behalf. That challenge is approved **on LinkedIn**, so `send_login_approval_email` was the only place it ever appeared, and nothing in the product changed once the user tapped Yes. The Account page said *"A session is saved"* before the approval and exactly the same thing after: a saved session only proves a credential exists, never that it worked.

This makes the sign-in itself visible.

## How

- **`utilities/linkedin/login_status.py`** (new) — the per-user sign-in record, in Redis next to the 429 breaker (reusing `shared_redis_client()`): `approval_pending` → `signed_in` / `approval_timed_out`. `approval_cleared_at` is set **only** when the sign-in followed a pending approval — that is the precise fact the reporter could not see.
- **`_persist_session_cookies`** is where both of `login_to_linkedin`'s success paths meet, so it is the one place a sign-in is recorded — and it is recorded *before* the cookie write is judged, since the approval landed whether or not the cookies persisted.
- **The wait loop always closes its own pending record** (`_gave_up`) instead of letting it expire, so the card can never sit for minutes telling the user to approve a sign-in nobody is waiting on. The pending record also carries a much shorter TTL (15 min vs 30 days) for the case where the worker dies mid-challenge.
- **`GET /user/linkedin-signin-status`** — session-scoped; a foreign user's record is never returned.
- **`LinkedInSignInStatusCard.tsx`** on the Account setup section, polling only while an approval is outstanding.

**Fails open throughout.** Redis down → every write no-ops, the endpoint reports `unknown`, and the card renders *"no sign-in recorded yet — nothing is wrong."* `unknown` must never read as a broken connection. The `user_id` lookup is best-effort for the same reason: a failed lookup costs a status line, never a login.

## Interpretation note

The auto-filed feedback was one sentence, so the reading is stated plainly: *"approved"* = the LinkedIn device sign-in approval the user is emailed about, which is the only user-facing "approval" in the `auth` area and the only one with no surface in the SPA. If the reporter meant something else, say so and this can be re-pointed cheaply — the store and endpoint are small.

## Testing

- 13 unit tests for the store — lifecycle, TTL env handling, per-user keys, fail-open on a down / erroring / corrupt Redis.
- 8 unit tests for the login-flow wiring — the three transitions, plus a lost cookie write (still records the sign-in) and an unresolvable user (silent skip).
- 7 integration tests driving the real endpoint through the real store (only Redis faked), including the 401, the cross-user check and the Redis-down case.
- 6 vitest cases for the card, including that `unknown` never reads as a failure.
- Full unit suite green locally: **9547 passed**.

Closes #933


gitchrisqueen:

- **Pending sign-in record left stale when the approval landed but the login then failed** (`helper.py` / `login_status.py`): `_wait_for_manual_approval` reported a confirmed approval only by returning `True` and letting the flow fall through to `_persist_session_cookies`. Every path that dies in between — the Feed-title `wait.until` timing out, the post-submit 429 check, `_page_is_transport_error`, or `_is_logged_in` being false at the end — skips `mark_signed_in` entirely, so the record stays `approval_pending` for up to the 15-minute pending TTL and the card keeps saying *"Waiting for you to approve … this turns green on its own"* to a user who already tapped Yes. That is the reported bug recurring inside the fix for it, and the PR body's "the wait loop always closes its own pending record" only held for the give-up branch. Fixed by recording the sign-in where the approval is detected (`_approved()`, mirroring `_gave_up()`), and making `mark_signed_in` carry `approval_requested_at`/`approval_cleared_at` across the second write the same login makes at the cookie persist — bounded by the pending window (`_is_recent`), so a routine sign-in weeks later never re-claims an old approval.

Checked and found clean: the close is honest (#933's acceptance criteria are generic and fully met, no later phase left untracked); per-user Redis keys + session-scoped endpoint, no cross-user read; fail-open holds on a down/erroring/corrupt Redis and on an unresolvable `user_id`; `log_debug` (not `log_warning`) on the expected no-ops, so no defect is filed for working behaviour; `unknown` never renders as a failure.

Tests: 4 new store cases (double write preserves the approval, an old approval is not re-claimed, unparseable timestamp) + 1 new login-flow case (approval recorded even when the login dies before the cookie persist), and the existing double-write assertion updated. Local: 9551 unit passed, 7 integration passed for the endpoint.

gitchrisqueen:

PR #935, opened 2026-08-02
feat(engagement): preview and edit the weekly group post before it publishes (closes #932)

### LINKED ISSUE #932: feat(scheduling): Add preview and edit capability for group post content
## Why
User reports that when scheduling a post to a group, they see toggles but cannot see the actual content of the post nor edit/revise it before posting.

Auto-filed from in-app feedback #30 by the feedback→auto-work loop (`docs/launch-and-marketing-plan.md` §B.2/B.3). Classifier: enhancement/medium, risk `none`, confidence 0.90.

## Scope
- Address the reported enhancement in the `scheduling` area of the product.

## Files
- `tests/unit/`

## Acceptance
- All required CI gates pass.

gitchrisqueen:
## What & why

A user reported that scheduling a post to a group showed them **toggles and nothing else** — they could not see the text that would be posted, let alone revise it. That was accurate: `auto_post_to_group` generated the post and published it inside the same Selenium run, so the text existed only in the moment before it went live under their name.

The weekly group post is now **two beats with a review window between them**:

| Beat | When | Does |
|---|---|---|
| `group-post-drafts` → `auto_draft_group_post` (new) | Sun 15:00 | The ONE place a group post's text is written. Opens **no browser** — voice comes from the cached profile (`load_profile_for_user`) — so a draft costs one LLM call and no Chrome session slot. Group choice is still the least-recently-tried post-enabled one (#769/#858), so the rotation is unchanged. |
| `group-posts` → `auto_post_to_group` | Tue 15:00 (unchanged) | Publishes that draft and **generates nothing**. A run with no READY draft publishes **nothing** — falling back to a fresh generation would ship exactly the un-previewed post this replaces. |

**Silence still ships it.** The draft's resting status is `ready`, not a pending approval, so cadence and volume are identical for a user who never opens the app. The preview exists for the user who *does* want to look.

### Where the user sees it
`GroupsCard` (Account → LinkedIn Groups — the same screen that only had toggles) now shows the queued post: the group it's going to, its **actual text in an editable box**, a character count against LinkedIn's 3000 cap, **Save post**, and **Skip this week**. Backed by `GET`/`PUT /user/group-post-draft`, both scoped to the caller's OWN open draft — the request never names a draft id, so one session can't reach another user's post.

### Edge cases the design pins down
- **One open draft per user** — the waiting draft may already carry the user's edits, so the draft beat carries it forward rather than replacing it with a generation they never asked for.
- **A dead session leaves the draft open**, so next week's run publishes the text they already approved.
- **The draft dies with its group's turn**: an unpostable (admin-only/announcement) group marks it `failed` alongside the #858 rotation stamp, and a group whose *Post* switch was flipped off between the two beats marks it `skipped` rather than publishing into a group the user opted out of.
- Emptying the text is not how you cancel (422 / disabled Save) — **Skip this week** is.

### Schema
`V20260802130926__add_group_post_drafts.sql` — new additive table `group_post_drafts` (`ready` / `skipped` / `published` / `failed`), no changes to existing tables.

## Testing

- `tests/unit/app/test_groups.py` — draft DB helpers (create/read/edit/status, ship time stamped only by `published`), the drafting task (writes without a browser, never replaces a waiting draft, quiet skip with no cached profile), and the publish task (publishes the reviewed text and never generates; **no draft / skipped draft / another user's draft / a pre-#932 call with no `draft_id` all publish nothing**).
- `tests/unit/app/test_selenium_queue_routing.py` — `auto_draft_group_post` must stay off the Selenium lanes.
- `tests/integration/test_group_post_draft_api.py` (new, 13 cases) — GET/PUT incl. 401/404/422/500 paths and the draft-resolved-from-session-not-request guard.
- `src/cqc_lem/ui/.../GroupsCard.test.tsx` — preview renders the real text, hides when nothing is queued, save/skip payloads, and the empty/over-cap guards.
- `poetry run pytest tests/unit -q` → **9546 passed** locally; the new integration file passes too. Vitest could not run in this worktree (UI dev deps aren't installed here) — CI covers it.

Docs: `docs/engagement-automation.md` gains the full posture; CLAUDE.md gains the one-line invariant; the now-stale "group posts have no review queue" line in `docs/content-core.md` and the Groups entry in `docs/SETTINGS_IA_RESEARCH.md` are corrected.

Closes #932


gitchrisqueen:

- **A DB read error could cancel every user's reviewed group post.** `get_post_enabled_group_ids` returned `[]` on `mysql.connector.Error`, and `[]` is exactly the value `auto_group_posts` reads as "no group takes posts" → `update_group_post_draft(..., SKIPPED)`. One transient fault during the weekly beat would have silently binned every user's approved draft (edits and all) for the week. Fixed: the read answers `None` for **unknown** (never `[]`), and the beat holds the draft open on unknown so it ships at the next slot — only a user, or a switch they actually flipped off, cancels a post they reviewed. Tests for both the `None` read and the hold.
- **A rewrite could be silently dropped and the ORIGINAL text published.** The draft textarea was outside `SettingsSaveContext`'s registry, so *Save All* saved only the group toggles and reported success, and the unsaved-changes guard (beforeunload + link interception) never fired on a pending edit — the exact "text I didn't approve went out" failure #932 exists to end. Fixed: it registers as its own save section (`'group-post'`, dirty only when the edit is savable), the card's own Save button now reports through `sectionSaveCallbacks('group-post')`, and the 3000 cap is one constant instead of three literals. Two new `GroupsCard.test.tsx` cases (Save All writes the rewrite; an unchanged textarea is not dirty).
- **Nothing pinned the review window.** The draft beat landing *before* the publish beat is the whole feature, and no test covered the new `group-post-drafts` entry (the repo pins every other beat this way). Added `TestGroupPostBeats` asserting both tasks and `draft day < publish day`.
- **New code used the legacy `myprint` shim** in three of the five new `db.py` helpers while the other two used `log_error` in the same PR — CLAUDE.md says avoid it in new code. Now `log_error(..., exc=err, user_id=…)` throughout.
- **Stale invariant comment** in `ai_helper.generate_group_post` ("a group post publishes straight from here — it never reaches the content-plan review queue"). It does now; corrected to say the queue is *unattended*, which is why the lint still only advises there.

Also checked and found sound: the close is honest (#932 has no later phase and its acceptance criteria are met); `post_enabled` is `NOT NULL DEFAULT 1` and `get_next_group_for_post` filters on the same flag, so the draft/publish beats can't disagree about a NULL; the publish path never regenerates and never touches `record_group_post` without a shipped post; `_strip_non_bmp` covers the user-edited text; both endpoints resolve the draft from the session, never from the request; the migration is additive with a timestamp version.

`poetry run pytest tests/unit -q` → **9549 passed**; the new integration file → 13 passed. Vitest can't run in this worktree (no UI deps) — CI covers it.

gitchrisqueen:

PR #949, opened 2026-08-03
feat(agent-pipeline): phase-guard holds route to MODE=phasefix instead of the owner

gitchrisqueen:
## Why

Every phase-guard hold currently parks the PR with `needs-human` + `agent:blocked` and assigns the owner — to perform two purely mechanical actions (file+link the follow-up issue, or drop the closing keyword) that the RUNBOOK's phased-work rule already teaches agents to do in MODE=start step 6. Deciding to finish an issue's stated requirements is not a human decision, and the owner should not be triaging multi-phase closes unless a genuine decision exists.

## What

- `phase_guard_ok` failure path now labels the PR **`agent:phasefix`** (new label, created) and posts a marker comment saying an agent will handle it — no owner assignment.
- New **phasefix lane** (before revise — a phasefix hold blocks an otherwise-green merge): dispatches **MODE=phasefix**, where the agent files + links the follow-up (checking for an existing one first) or drops the close, then hands the PR back to `agent:working`. If the remainder needs a `risk:*` sign-off, the hold rides on the **follow-up issue**, never the PR.
- Owner is assigned only after **`MAX_PHASEFIX_ATTEMPTS=2`** failed passes (durable counter in `state/phasefix-<PR>.count`, cleared when the guard passes), with a distinct "phasefix exhausted" comment.
- RUNBOOK gains the MODE=phasefix section; the phased-work enforcement paragraph and the playbook's label table/enforcement text updated to match.

Boxes-only warnings, DRY_RUN behavior, and the guard's fail-open posture are unchanged.

## Testing

`bash -n` passes. Shell/docs only — no app code paths; the live pipeline copy is synced separately per the existing hot-patch convention.

### DIFF
--- a/docs/AGENT_WORKFLOW_PLAYBOOK.md
+++ b/docs/AGENT_WORKFLOW_PLAYBOOK.md
@@ -30,6 +30,7 @@ it. **An issue without `agent:ready` does not exist to the pipeline.**
 | `agent:revise` | (PR) Owner requested changes / answered a Decision Comment | Runner |
 | `agent:blocked` | Held after escalation; runner skips it | Runner/agent |
 | `agent:depfix` | (PR) Dependabot PR with failing CI, priority fix lane | CI router |
+| `agent:phasefix` | (PR) Phase-guard hold: agent files/links the follow-up for a multi-phase close | Runner |
 | `needs-human` | Waiting on a human decision or human-only action | Agent (on escalation) or human |
 
 ### Priority (queue order within the ready pool)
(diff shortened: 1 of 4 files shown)


gitchrisqueen:
🔎 Claude adversarial review — FIXED 5 findings (pushed `fffae57c`)

- **The phasefix budget could never fire (unbounded Claude spend).** `MAX_PHASEFIX_ATTEMPTS` was counted inside `phase_guard_ok`, which only runs on an `agent:working` PR. A MODE=phasefix run that died — or escalated — without relabeling left `agent:phasefix` on the PR, so the guard never saw it again, the counter never moved, and the lane re-dispatched a 45-minute Claude run every 5-minute tick forever. Fixed: the counter now counts **dispatches**, bumped in the lane where the spend happens, and the lane enforces the cap itself and parks to the owner when it's spent (park path factored into `phase_park_to_human`, shared with the guard).
- **The lane re-dispatched on top of a human hold.** The `agent:phasefix` query had none of the `needs-human`/`agent:blocked` filtering every other lane has, and RUNBOOK step 5 told the agent to escalate without dropping the lane label — so an escalated PR would get another phasefix run every tick while it sat parked for the owner. Fixed both ends: jq filter on the lane, and step 5 now removes `agent:phasefix` on escalate.
- **Race: a second slot could park a PR mid-run.** The candidate was judged before `claim_branch`, so with the counter at max one slot could park a PR to the owner while another slot's phasefix agent was still working and about to hand it back. The claim now comes first.
- **A stale hold burned a run.** After option (b) (drop `Closes #N`) the PR closes nothing, so the guard has nothing to hold — but the lane still dispatched with `ISSUE=`. It now hands the PR straight back to `agent:working`.
- **Hardening:** the empty-worktree check the START lane documents (`cd ""` → agent runs in $HOME) was missing from the new lane; `$BASE/state` is now in the top-level `mkdir -p` instead of relying on `capacity.sh` loading; a garbage counter file reads as 0 rather than as exhausted (it would have parked a PR to the owner on the first hold).

Also checked, no finding: the guard's fail-open posture, boxes-only warning path, DRY_RUN behavior, comment de-duplication (the "phasefix exhausted" grep is distinct from the marker grep), the `agent:phasefix` label exists in the repo, lane ordering vs. the merge fast path, and review-freshness (a phasefix pass pushes no commit, so the existing review marker still holds). This PR closes no issue, so the phase guard has nothing to check on itself.

Tests: `tests/unit/test_agent_pipeline_phasefix.py` (10 passing) — runs the counter helpers lifted verbatim from `tick.sh` and asserts the lane's guards; `bash -n` clean.

⚠️ **Note for the owner:** `/home/lem/agent-pipeline/tick.sh` was already hot-patched to this PR's pre-fix code, so the live runner currently has the unbounded lane. It needs the same sync after merge — I deliberately did not edit the live script while a tick was executing it.

gitchrisqueen:

PR #955, opened 2026-08-03
fix(billing): daily caps stop resetting — the brand phase seeds, it no longer re-asserts (closes #952)

### LINKED ISSUE #952: fix(billing): Daily caps reset to lower values after being set
## Why
I keep setting my daily caps to the balanced recommendations and they keep resetting to less for some reason. Investigate and find out why and what is causing an overuse and fix it

Auto-filed from in-app feedback #35 by the feedback→auto-work loop (`docs/launch-and-marketing-plan.md` §B.2/B.3). Classifier: bug/high, risk `none`, confidence 0.90.

[Watch the session replay](https://us.posthog.com/project/475262/replay/019fc536-ece7-7dd0-a328-ff040b0de970) — the browser session this report was filed from.

## Scope
- Address the reported bug in the `billing` area of the product.

## Files
- `src/cqc_lem/api/main.py`
- `src/cqc_lem/utilities/db.py`
- `tests/unit/`

## Acceptance
- All required CI gates pass.

gitchrisqueen:
## What was happening

The reporter kept setting their daily caps to the Settings hub's **Balanced** recommendation and found them lowered again later. Nothing on the save path resets them — the nightly beat does.

- `sync-brand-account` runs daily at **00:45 UTC** (`my_celery.py`) → `auto_sync_brand_account` → `sync_brand_preferences`.
- Since #736 the brand account is **user 1 by convention** — which is also the owner's ordinary LEM account.
- `brand_preference_overrides` re-asserted the `LAUNCH_PHASE` policy over that row as a ceiling. Prod runs the default `LAUNCH_PHASE=P0` (`.env.example:440`), whose policy is **8 comments / 5 DMs / 5 invites**, `pre_review` + `suggest`.
- The SPA's **Balanced** preset is **15 / 10 / 8**, `auto_approve` + `auto_queue` — deliberately the *P1* numbers (`docs/SETTINGS_IA_RESEARCH.md` §7).

So the product recommended P1 and then, before morning, put it back to P0 — silently, since the only trace was an INFO line that prod's `POSTHOG_LOG_LEVEL=WARNING` never forwards. That is the "resets to lower values" the reporter saw. It only ever affects the brand user (user 1); no other account goes through this task.

## The fix

The phase now **seeds** rather than re-asserts:

| Account state | What the sync writes |
|---|---|
| No saved `engagement_preferences` row | The phase's caps + connect posture, exactly as before |
| A saved row | Nothing for caps/posture — except a cap above `BRAND_CAP_CEILINGS` (20 / 20 / 10), which is still pulled back |

Saved settings are the sign-off; an env var nobody has touched is not. The hard guarantee that survives is the one that matters for safety — the brand can never run hotter than a paying user's out-of-the-box defaults — and the per-day caps, 429 breaker (`rate_limit.py`) and human pacing are unchanged.

Two failure modes are closed with it:

- `db.engagement_preferences_are_configured` is **three-valued** — `None` means the row could not be READ, which is not "never configured" (#639). An unreadable row **skips the sync** instead of seeding the phase over the owner's real caps.
- A cap the sync cannot parse is left alone rather than rewritten. Its only source is `get_engagement_preferences`, which hands back code defaults after a failed read, so rewriting from it is how a fault would silently widen outbound.

The task's return line now names only the caps it actually wrote (it previously read the three cap keys unconditionally, which would `KeyError` now that a run can apply none).

## Testing

- `tests/unit/app/test_brand_account.py` — the reported case (Balanced under P0 writes nothing), a saved stricter cap, a `0` cap, the connect posture, the over-ceiling clamp still applying + still being logged, an unreadable cap, an unreadable row skipping the sync, and both new summary shapes.
- `tests/unit/utilities/test_db_engagement_prefs.py` — the three states of `engagement_preferences_are_configured`.
- Full unit lane green locally: `9804 passed`.

## Note for after deploy

The account's caps are currently sitting at the P0 values this task last wrote. Re-pick **Balanced** in Settings once after the release — from then on it stays.

`release:now`: user-visible, `priority:high` bug on the owner's own account, and every night that passes reverts the setting again.

Closes #952

### DIFF
--- a/docs/launch-and-marketing-plan.md
+++ b/docs/launch-and-marketing-plan.md
@@ -283,15 +283,24 @@ self-marketing off. What the brand does:
 self-marketing can never exceed ToS-safe volume. Outbound cadence for the brand is a `risk:product-decision`
 config (daily connect/DM caps) the owner signs off once.
 
-**The collision guard (issue #736).** Because the brand user is *also* the owner's ordinary LEM account, the
-nightly `sync_brand_preferences` applies the phase policy as a **ceiling, not an assignment**: it can only pull
-`max_comments_per_day` / `max_dms_per_day` / `max_invites_per_day` DOWN to the phase's number, and only tighten
-`connection_request_mode` / `connection_targeting_mode` (strictness order `pre_review` > `auto_approve`, `off` >
-`suggest` > `auto_queue`). A value the owner tuned STRICTER by hand survives the sync — honouring it can never
-widen outbound, and stomping it every night is the one real risk this convention introduces. The seeded content
-fields (`focus_topics`, `business_goals`) stay fill-only-when-empty as before, everything else on the row is
-untouched (only the policy fields are sent — issue #639), and any field the sync does change is named
-before → after in one INFO line, so an edit to the owner's own settings is never silent.
+**The collision guard (issues #736, #952).** Because the brand user is *also* the owner's ordinary LEM account,
+the nightly `sync_brand_preferences` **seeds, it does not re-assert**. Which half applies is decided by
+`db.engagement_preferences_are_configured` — three-valued, because "could not read the row" is not "never
+configured" and an unreadable row skips the sync outright:
+
+- **No saved engagement row** → the phase sets `max_comments_per_day` / `max_dms_per_day` /
+  `max_invites_per_day` and `connection_request_mode` / `connection_targeting_mode` outright.
+- **A saved row** → those are the owner's own Settings choices and the phase is not applied to them at all.
+  Only `BRAND_CAP_CEILINGS` still binds: a cap above the shipped per-user default (20 / 20 / 10) is pulled back,
+  so the brand can still never run hotter than a paying user out of the box.
+
+The re-assertion this replaces was the reported bug: the Settings hub recommends the **Balanced** preset
+(15 / 10 / 8 — the P1 numbers) while prod runs `LAUNCH_PHASE=P0`, so the caps the owner picked in the UI were
+back at 8 / 5 / 5 by morning with nothing on screen to say why. Saved settings are the sign-off; an env var
+nobody has touched is not. The seeded content fields (`focus_topics`, `business_goals`) stay fill-only-when-empty
+as before, everything else on the row is untouched (only the policy fields are sent — issue #639), and any field
+the sync does change is named before → after in one INFO line, so an edit to the owner's own settings is never
+silent.
 
 ### C.3 Other automated channels (each agent-run)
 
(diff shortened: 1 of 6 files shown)


gitchrisqueen:

- **Parallel existence helper in `db.py`**: the new `engagement_preferences_are_configured` re-implemented `has_engagement_preferences` (which has lived at db.py:11215 since the onboarding work) — same question, two queries, and opposite error postures (`None` vs `False`). It also read all 41 engagement columns + JSON/bool-decoded them just to test row existence, on a nightly beat. Fixed: the three-valued function is now the ONE query (`SELECT 1 … LIMIT 1`) and `has_engagement_preferences` is its two-valued view (`… is True`), so both existing call sites (`GET /user/engagement-preferences`, `onboarding.py`) keep their exact behaviour and the two can never drift.
- **`auto_sync_brand_account` reported "nothing changed" for a run that did change the row**: the new summary keys off `CAP_FIELDS` only, so a configured account whose caps were left alone but whose empty `focus_topics`/`business_goals` were seeded still wrote to the owner's row and reported "its own caps are within policy, nothing changed". That is the same phantom-edit problem this PR fixes, pointing the other way. Fixed: three states now — caps written / content-only (names what it seeded) / genuinely nothing.

Both covered by new tests. Full unit lane green locally: `9807 passed`.

Verified and NOT flagged:
- The diagnosis holds — `sync_brand_preferences` is the only writer of `max_*_per_day` outside the SPA save path, it only ever touches `brand_user_id()` (user 1), and the Balanced preset (15/10/8, `presets.ts`) is exactly P1 against a prod `LAUNCH_PHASE=P0` (8/5/5).
- `sync_brand_preferences` can now return `{}` (falsy, not a failure) — the sole caller correctly tests `is None`.
- The three caps are `INT NOT NULL DEFAULT` in V34/V20260724103715, so the "left alone because unreadable" path cannot leave a NULL cap that `run_automation` would then read as 20.
- `MODE_STRICTNESS`/`_capped`/`_strictest` have no remaining references anywhere.
- Close is honest: #952 has no phased scope and every acceptance criterion is in the diff.

One judgement call worth the owner's eye at merge (not a blocker, and documented in the PR body + `docs/launch-and-marketing-plan.md`): once user 1 has a saved row, advancing `LAUNCH_PHASE` no longer ramps the brand's caps **or** its connect posture — P0's `pre_review` gate is no longer forced on the brand. `BRAND_CAP_CEILINGS` (20/20/10) still binds the caps, and every posture the preset can set is one a paying user gets out of the box, so nothing runs hotter than the shipped product.

gitchrisqueen:

Issue #957, opened 2026-08-03
fix(security): restore a CSRF layer for the four query-parameter mutating /api routes (follow-up of #950)


gitchrisqueen:
## Why

#950 retired the shared `/api` bearer token from the SPA build. That was right — it was one of the
server's `API_ACCESS_TOKENS` inlined into a public bundle, so every visitor held it — but it had a
second, accidental job that went with it.

A cross-site HTML form POST **cannot set an `Authorization` header**, even when the attacker knows
the value, and setting one via `fetch()` triggers a preflight that has nothing to succeed against
(no CORS middleware is installed). So the bearer was worthless as access control and real as a
**CSRF** layer.

Four mutating routes take query parameters and no body, so the "JSON body needs a preflight" layer
does not cover them either (`docs/identity-and-sessions.md` § CSRF):

- `POST /create_weekly_content/`
- `POST /invite_to_li_company_page/`
- `POST /aws_test_get_my_profile/`
- `POST /automate_reply_commenting`

Since #950 **`SameSite=Lax` is the whole defence for those four** — one layer, not two. `Lax` holds
on its own (the cookie is simply not attached to a cross-site POST) and each route only ever queues
work for the CALLER's own account, so the realistic worst case is a forged job the user could have
started themselves. That is why this is a follow-up and not a blocker. But one layer is one layer,
and a new query-parameter mutating route inherits it.

## Scope

Add the standard replacement — a **non-secret custom request header** the SPA sends and a forged
cross-site form cannot (e.g. `X-LEM-Client: spa` set in `ui/src/api/client.ts`'s interceptor):

- Require it on gated `/api` requests whose credential is the **session cookie**. A request
  authenticating on the bearer is exempt: a non-browser caller has no CSRF exposure, and an old SPA
  bundle cached across the release still sends one, so the exemption also makes the rollout
  breakage-free.
- Nothing about it is a secret. It is a header a cross-origin form cannot set — that is the entire
  mechanism, and it must be documented that way so it is never mistaken for a token.
- Consider whether to scope it to state-changing methods only, or to every gated request.

## Acceptance

- [x] The SPA sends a non-secret custom header on every `/api` request, from the one axios client.
- [x] A cookie-authenticated mutating `/api` request without that header is refused (403), with a
      test per each of the four query-parameter routes above.
- [x] A bearer-authenticated request without the header still succeeds (non-browser callers and
      pre-#950 cached bundles keep working) — with a test.
[...]

PR #958, opened 2026-08-03
fix(security): retire the SPA's shared /api bearer token — the session is the identity (closes #950)

### LINKED ISSUE #950: fix(security): the shared `/api` bearer token ships in the SPA build — retire it now that the session is the identity (follow-up of #914)
## Why

Split out of **#914**, which closed the hole where `/api` routes read the acting user from an
`email` / `user_id` / `post_id` **request parameter**. Every route resolves its caller from the
session now, so that token is no longer *sufficient* for anything — which is the important half,
and why this is a follow-up rather than a blocker.

What remains is that it still exists and is still trusted as a gate. `API_ACCESS_TOKENS` is checked
by the middleware in `src/cqc_lem/api/main.py`, and the SPA ships it in its build
(`VITE_API_TOKEN`, `src/cqc_lem/ui/src/api/client.ts`) — a build-time constant in a public bundle,
so it is held by everyone who has ever loaded the page. A secret that every visitor holds is not a
secret; carrying it as one means:

- it reads as a layer in threat models that it is not (it was named as the only thing in front of
  the #914 routes, and was worth nothing there);
- it cannot be rotated without a rebuild + redeploy of the SPA;
- `docs/identity-and-sessions.md`'s CSRF section counts it as one of the layers in front of the four
  query-parameter mutating routes, which is only honest while it is unguessable — and it isn't.

Quoted from #914's scope, which deliberately did not touch it:

> The only thing in front of them is the shared bearer token (`API_ACCESS_TOKENS`), which the SPA
> ships in its build (`VITE_API_TOKEN`, `ui/src/api/client.ts`) and is therefore known to anyone who
> loads the page.

## Scope

1. **Enumerate what the token is actually still doing.** It is not the identity anywhere after #914.
   Establish whether it is load-bearing for anything besides the SPA — the browser extension,
   `scripts/`, Celery callbacks, the tutorial capture harness, the LinkedIn OAuth return trip, any
   Postman/README example. Each of those either has a session, has an admin path
   (`_require_api_and_admin` / `X-Admin-Secret`), or needs one.
2. **Stop shipping it to the browser.** The SPA is cookie-authenticated; a token in its bundle buys
   nothing it does not already have.
3. **Decide what the middleware becomes.** Options, in the PR: keep the bearer as a *non-browser*
   credential only (rotatable, never in a build artifact); replace it with per-caller tokens; or
   drop the check entirely and let the session + admin secret be the whole gate. This is the part
   worth an owner decision, so bring a recommendation.
4. **Correct the docs that count it as a layer** — `docs/identity-and-sessions.md` (CSRF section)
   and `docs/AUTH_SECURITY_DESIGN.md`.

### Also in scope (same "still reachable, no longer used" shape)

- **`db.update_user(email=…)`** is now unreachable from the API — #914's only caller stopped passing
  it — but it still moves an account's address with no `user_email_history` row, no PIN to the new
  address and no session revoke, i.e. around everything `POST /user/email/change/init|verify` exists
  to enforce. A latent footgun, not a live one. Delete the parameter or route it through
  `change_user_email`.

## Acceptance

- [ ] No LEM-issued shared secret is present in the built SPA bundle (`grep` the `dist/` output in
      CI, so it stays true).
- [ ] Every non-browser caller of `/api` is enumerated and authenticates on a session, an admin
      path, or a documented non-browser credential.
- [ ] `docs/identity-and-sessions.md` + `docs/AUTH_SECURITY_DESIGN.md` no longer count the SPA
      bearer token as a control, and the CSRF layer count is restated honestly.
- [ ] `db.update_user` can no longer move an account email outside `change_user_email`, with a test.
- [ ] Tests: the middleware's new contract, and a regression test that the four query-parameter
      mutating routes named in #914 still refuse an unauthenticated caller.

Follow-up of: #914. Related: #745 (2b sessions), #905 (2c.1 scopes).

gitchrisqueen:
## What & why

`API_ACCESS_TOKENS` was still checked by the `/api` middleware and still shipped to the browser: the
SPA read it out of `VITE_API_TOKEN`, and Vite inlines a `VITE_*` at build time, so one of the
server's bearer tokens was **a constant in a public bundle**. Held by everyone who had ever loaded
the page, unrotatable without a rebuild + redeploy, and counted as a layer in threat models it was
worth nothing in. #914 had already made the session the identity on every `/api` route, so the token
was checked and no longer sufficient for anything — which is what made this a follow-up rather than
a blocker.

### 1. The SPA ships no LEM-issued secret

- `ui/src/api/client.ts` sends no `Authorization` header.
- `compose/local/Dockerfile` has no `VITE_API_TOKEN` ARG; `build-and-push.yml` no longer passes it.
- **UI Build now builds with CANARY values** for the LEM-issued `VITE_*` names and greps `dist/` for
  them. Read one back into the bundle and CI fails. (`VITE_POSTHOG_KEY` is deliberately not
  canaried — a write-only third-party ingest key is *meant* to be public.)
- `UI_API_TOKEN` (the GitHub secret) is retired; docs say to delete it.

### 2. What the middleware becomes — a credential gate, not authorisation

It asks only *"did this caller bring **a** credential"*: a valid **non-browser** bearer (scripts,
Postman, `/api/admin/*`) **or** a session credential the route itself judges (the `lem_session`
cookie, or the `X-Session-Token` header the SPA sends on the cookie-less fallback).

Presence, not validity — the middleware runs before routing and has no database, and
`require_session_user_id()` in the handler already fails closed. **A forged cookie clears the edge
filter and is then refused by the route**: the same 401, one step further in. The six `/api/admin/*`
routes that gate through `_require_api_and_admin` still demand the bearer *and* `X-Admin-Secret` —
it re-checks the bearer in the handler. The other twelve admin routes run on `X-Admin-Secret`
alone or on an admin session, and for those this PR *did* drop a layer; see the review-round
note below.

Rollout is breakage-free in both directions: an old SPA bundle cached across the release still sends
the bearer (still valid), a new one sends the cookie.

### 3. `db.update_user` can no longer move an address

`email=` is gone from the signature and `email = %s` is out of `_ALLOWED_USER_CLAUSES`. #914 removed
its last caller, but it still did a bare `UPDATE users SET email` — no `user_email_history` row, no
PIN to the new address, no session revoke, i.e. around every guarantee
`POST /user/email/change/init|verify` exists to make. A latent footgun one keyword argument from
being fired.

### 4. Docs — including the part that is *worse*, not better

`docs/identity-and-sessions.md` gains a section on the bearer's new contract plus a table
enumerating **every** non-browser caller of `/api` and what it authenticates on (SPA, tutorial
capture harness, browser extension, `scripts/generate_media_variants.sh`, Postman, Stripe webhook,
the LinkedIn OAuth return trip, Celery). `docs/AUTH_SECURITY_DESIGN.md` states plainly that this
token is not a control in that design and never was.

**The CSRF count is restated honestly, and it went DOWN.** A cross-site form POST cannot set an
`Authorization` header *even when the attacker knows the value*, so the bearer was worthless as
access control and **real as a CSRF layer**. The four query-parameter mutating routes from #914 are
now behind `SameSite=Lax` alone — one layer, not two. `Lax` holds on its own (the cookie is simply
not attached to a cross-site POST) and each route only queues work for the caller's own account, so
the realistic worst case is a job the user could have started themselves. The standard replacement —
a non-secret custom request header — is filed and linked rather than smuggled into this diff.

**Follow-up: #957** (restore a CSRF layer for those four routes).

## Owner decision (applied)

Answered `ok` on the Decision Comment → **1A** (bearer = a non-browser credential; a session
credential also clears the gate — what this PR implements) and **2A** (ship now; the CSRF
replacement lands on **#957**). No shape change; the sections above already describe 1A.

### What the review round added on top

- **The safety claim is now a test, not a paragraph.** `tests/unit/api/test_api_route_identity.py`
  walks the live route table and fails the build on any gated `/api` route that reaches neither the
  session resolver nor the admin secret. Both gate sets are DERIVED by closing over the call graph
  from `get_session_user_id` / `_require_admin`, so a new wrapper (`_require_user_admin`,
  `_owned_edition`, the next one) counts automatically. 151 gated routes checked;
  `/api/carousel-templates` is the single allowlisted entry (static template metadata — no caller to
  authorise) and the allowlist is itself checked for staleness.
- **The canary could pass without running.** A missing `dist/` makes `grep -rl` exit 2, which an
  `if` reads as "found nothing" — the step printed a tick and went green. It now asserts `dist/`
  exists, distinguishes exit 1 from exit ≥2, and carries a **positive control**: two `VITE_*` values
  the SPA reads at module scope must be *present* in `dist/`, because a negative grep proves nothing
  unless it could have found something. A **source grep** on `src/` catches the reintroducing diff a
  build earlier.
- **The cookie's attributes are pinned by test** now that it is the browser's only credential —
  `httponly` / `secure` / `samesite=lax` / `path=/`, plus the fallback when
  `SESSION_COOKIE_SAMESITE` is unset or unusable (a login must not 500). And the four
  query-parameter routes are asserted **POST-only**: `Lax` *does* send the cookie on a top-level
  GET, so a state-changing GET would be forgeable by a bare link.
- **Integration coverage for the real flow** (`tests/integration/test_api_credential_gate.py`): a
  live session row, its token in `lem_session`, a gated route reached with **no `Authorization`
  header at all** → 200; a forged cookie → 401 from a real lookup, not a mocked one; a bearer holder
  → 401; a revoked session → 401 immediately.
- **A whitespace-only cookie cleared the gate.** `lem_session="  "` parses to `"  "`, which is
  truthy — the cookie is stripped like the header now, with the case tested.
- Smaller: `Callable`/`Awaitable`/`-> Response` on the middleware, `Optional[str]` on the
  `update_user` signature this PR touched, and the in-code comment corrected — the gate keeps only
  traffic that never tries off the handlers, since one arbitrary cookie byte clears it.

### What the adversarial review round added

- **`/api/admin/*` is three gates, not one, and the PR said one.** Six routes use
  `_require_api_and_admin` (bearer **and** `X-Admin-Secret` — it re-checks the bearer itself, so
  they are unchanged). Nine use `_require_admin` (`X-Admin-Secret` alone — including
  `/admin/youtube-token`, which installs an OAuth refresh token, and `/admin/user/location`). Three
  use `_require_user_admin` (an admin session; they are SPA pages and never could hold a bearer).
  For those twelve the bearer came from the **middleware**, so this PR is what removed it — while
  the docs went on asserting the layer was still there. It costs nothing today (every bearer value
  is public, so a caller with `ADMIN_SECRET` has one too) and stops being free the moment **#965**
  rotates them. `identity-and-sessions.md` now carries the per-route table;
  `AUTH_SECURITY_DESIGN.md` and `TESTING_ENGAGEMENT_API.md` are corrected to match. A new standing
  test pins what survives: every `/api/admin/*` route reaches one of the three — the route-identity
  test would have accepted one that merely resolved a session, which every signed-in user has.
- **A test that could not fail.** `test_allowlist_entries_are_real_gated_routes` asserted
  `_api_token_required(path) or not _API_ACCESS_TOKEN_SET`; the unit env has no
  `API_ACCESS_TOKENS`, so the right-hand side was always true and the "is it actually gated" half
  never ran — moving `/api/carousel-templates` into `_PUBLIC_API_PREFIXES` would have gone
  unnoticed. It asks `_is_public_api_path` directly now, which is env-independent.
- **One CSRF layer means one env var can delete it.** `SESSION_COOKIE_SAMESITE=none` attaches the
  cookie to cross-site POSTs, which for the four query-parameter routes is now the *whole* defence
  rather than a redundant one. The env table warned about `strict` and said nothing about `none`.
  Documented in the table and in the CSRF section until #957 lands.

### The part of the rollout that is not free — **#965**

"Breakage-free in both directions" is true only because a known-public secret keeps working. Every
`API_ACCESS_TOKENS` value ever built into a bundle is held by everyone who loaded the page, and
post-merge it still clears the edge filter and still pairs with `X-Admin-Secret` on `/api/admin/*`.
This PR retires the **shipping mechanism**, not the leaked secret. They are kept alive only for the
cached-bundle window (a tab open across the release — hours, not days). **#965** tracks the
rotation + the `UI_API_TOKEN` deletion; the docs now say this rather than implying the token is
already harmless.

`update_user(email=…)`: grep evidence for the removal — the only caller anywhere in `src/`,
`scripts/` or `tests/` is `main.py:2292`, `update_user(user_id, blog_url=…, sitemap_url=…)`. No
Celery task, script or migration passes `email=`.

**Follow-up: #957** (restore a CSRF layer for those four routes) · **#965** (rotate the
now-public token values).

## Testing

`poetry run pytest tests/unit -q` → **9836 passed** locally.

- `tests/unit/api/test_api_token_gate.py` — the gate's new contract: a session cookie or
  `X-Session-Token` clears it without a bearer; an *empty* cookie, a whitespace header, or an
  unrelated cookie does not; a valid bearer still does; the gate stays off when
  `API_ACCESS_TOKENS` is unset.
- Same file, new class — the #914 regression, re-run because the gate got **looser**: each of
  `POST /create_weekly_content/`, `/invite_to_li_company_page/`, `/aws_test_get_my_profile/`,
  `/automate_reply_commenting` refuses (a) no credential, (b) a forged cookie, (c) a bare bearer
  holder. 401 every time, and the task behind it never runs.
- `tests/unit/utilities/test_db_coverage_profiles.py` — `update_user(email=…)` raises without
  opening a connection, and `email = %s` is not an allowed clause.
- UI Build (CI) is the standing proof for the bundle criterion. Not runnable locally — the
  toolchain needs Node 22 and this box has 18.

## Acceptance criteria

- [x] No LEM-issued shared secret in the built SPA bundle, greppable in CI so it stays true
- [x] Every non-browser caller of `/api` enumerated and authenticating on a session, an admin path,
      or a documented non-browser credential
- [x] `identity-and-sessions.md` + `AUTH_SECURITY_DESIGN.md` no longer count the SPA bearer as a
      control; the CSRF layer count restated honestly
- [x] `db.update_user` can no longer move an account email outside `change_user_email`, with a test
- [x] Tests: the middleware's new contract + the four query-parameter routes still refuse an
      unauthenticated caller

Closes #950


gitchrisqueen:

Held (`needs-human`, risk: **security**). The issue named the middleware's future as the part worth
your call, and retiring the bearer also removes a CSRF layer it was accidentally providing — so
there are exactly two questions here. Everything else (the bundle, the docs, `update_user`, the
tests) is done and needs nothing from you.

### 1. What does the `/api` middleware become?

The token is no longer the identity anywhere (#914), and it is no longer in the browser (this PR).
The question is what is left of the check.

- **A. Bearer = a non-browser credential; a session credential also clears the gate** — what this PR
  implements. `API_ACCESS_TOKENS` stays set in prod for `scripts/`, Postman and `/api/admin/*`
  (which also needs `X-Admin-Secret`), and is now rotatable in `/opt/lem/.env` alone with no
  rebuild. The gate is an edge filter that keeps credential-less traffic off the handlers;
  `require_session_user_id()` in the route is the actual authorisation and fails closed.
  ✅ *recommended*
- **B. Drop the middleware entirely** — session + admin secret become the whole gate. Slightly
  simpler and arguably more honest (the gate proves nothing about *who* you are), but it deletes the
  one credential `scripts/generate_media_variants.sh` and the Postman collection use, and removes
  the outer refusal for unauthenticated scanners. More churn, less defence in depth.
- **C. Per-caller tokens** — issue one token per non-browser caller (script, Postman, harness) so a
  leak is attributable and revocable individually. Genuinely better, but it is a new credential
  store, a rotation story and an issuance path for **three** callers. Disproportionate now; worth
  filing if you want it.

### 2. The CSRF layer this removes — ship now or hold?

A cross-site form POST cannot set an `Authorization` header **even knowing the value**, so the
bearer was worthless as access control and *real* as a CSRF layer. The four query-parameter mutating
routes from #914 (`create_weekly_content`, `invite_to_li_company_page`, `aws_test_get_my_profile`,
`automate_reply_commenting`) are now behind `SameSite=Lax` alone. `Lax` holds on its own — the
cookie is not attached to a cross-site POST at all — and each route only queues work for the
caller's own account, so a forged one buys a job the user could have started themselves.

- **A. Merge now; the replacement lands on #957** — already filed (`agent:ready`, `risk:security`):
  a non-secret custom request header the SPA sends and a forged form cannot. The docs state the
  count as one layer in the meantime rather than implying two.  ✅ *recommended*
- **B. Hold this PR until #957 lands, ship them together** — no window at all where the count is
  one, at the cost of leaving the token in the public bundle for another cycle.
- **C. Don't retire the bearer** — keep shipping it in the SPA build. Closes #950 as won't-fix.

**My recommendation: `1A 2A`.** The token's value was never as a gate and its cost is real (a
secret every visitor holds, unrotatable without a redeploy, miscounted in two threat-model docs);
`SameSite=Lax` genuinely covers the CSRF case on its own, so #957 is hardening a defence that
already holds rather than patching a hole this PR opens.

---

**One operational step after merge, whichever you pick:** delete the **`UI_API_TOKEN`** repo secret
(Settings → Secrets and variables → Actions). Nothing reads it any more — `ui-build.yml` and
`build-and-push.yml` no longer pass it and the Dockerfile has no `VITE_API_TOKEN` ARG. Leave
`API_ACCESS_TOKENS` in `/opt/lem/.env` as it is; it now has no build artifact to match, so you can
rotate it there whenever you like.

gitchrisqueen:
ok

gitchrisqueen:
Addressed your review — `ok` read as **1A 2A**, and applied.

**1A** (bearer = a non-browser credential, a session credential also clears the gate) is what the PR already implemented, so nothing changed shape. **2A** — shipping now, with the CSRF replacement on **#957**, which is filed, linked in the PR body and labelled `agent:ready` + `risk:security`.

Since your `ok` arrived alongside a **Changes Required** from the automated reviewer, I worked its blocking findings too rather than hand you a PR that had one outstanding. The pushed commit is `35f2b99`:

| Finding | What I did |
|---|---|
| No mechanical proof every gated route enforces session/admin auth | `tests/unit/api/test_api_route_identity.py` — walks the live route table, 151 gated routes, fails the build on one that reaches neither gate. Both gate sets are **derived** by closing over the call graph from `get_session_user_id` / `_require_admin`, so a new wrapper counts automatically. One allowlisted route (`/api/carousel-templates` — static template metadata), and the allowlist is checked for staleness. Verified it actually fails by removing that entry. |
| The canary passes silently with no `dist/` | Real bug — `grep -rl` exits 2 on a missing directory and the `if` read that as clean. Now asserts `dist/` exists, distinguishes exit 1 from ≥2, and adds a **positive control** (two `VITE_*` the SPA reads must be *present* in `dist/`). Also a **source grep** on `src/`, which catches the reintroducing diff a build earlier. |
| The previously-public token is still valid | Stated plainly in the docs instead of left implicit, and **#965** files the rotation + the `UI_API_TOKEN` deletion. "Breakage-free in both directions" is only true because a known-public secret keeps working — for the cached-bundle window, not forever. |
| CSRF / cookie attributes unpinned | `TestSessionCookieAttributes` pins `httponly` / `secure` / `samesite=lax` / `path=/` on what `_set_session_cookie` issues, plus the fallback when `SESSION_COOKIE_SAMESITE` is unset or unusable. And the four query-parameter routes are asserted **POST-only** — `Lax` does send the cookie on a top-level GET. |
| Whitespace cookie clears the gate | Real: `lem_session="  "` parses to `"  "`, which is truthy. Cookie stripped like the header, case added. |
| Missing integration coverage | `tests/integration/test_api_credential_gate.py` — real session row → `lem_session` cookie → gated route with **no `Authorization` header** → 200; forged cookie → 401 from a real lookup; bearer holder → 401; revoked session → 401 immediately. |
| `SESSION_COOKIE_NAME` import | Already present (`main.py:120`) — no change needed. |
| Implicit `Optional`, untyped middleware | Both fixed. |
| `update_user(email=…)` out of scope | Kept (you said everything but the two questions was settled), with the grep evidence asked for: the **only** caller anywhere in `src/`, `scripts/` or `tests/` is `main.py:2292`, passing `blog_url` / `sitemap_url`. Nothing passes `email=`. |
| Lingering `VITE_API_TOKEN` declarations | Checked — no `env.d.ts` / `vite-env.d.ts` exists, and the only remaining mentions are the docs explaining the retirement, the Dockerfile comment saying there is no ARG, and the CI canary value itself. |

**Deliberately not done**, so it isn't read as silence:

- **Rename `api_token_middleware` → `api_credential_middleware`** — accurate, but pure churn across a security diff and it worsens the revert story. Say the word and I'll do it separately.
- **Log gate rejections** — an edge 401 is routine (scanners, expired sessions), so a `log_warning` there would re-emit at ERROR and file a grouped `$exception` for working behaviour, which is the exact anti-pattern in `utilities/CLAUDE.md`. Not worth a DEBUG line nobody reads either.
- **Per-IP rate limiting on `/api`** and **`__Host-` cookie prefix** — both real, both bigger than this PR. `__Host-` in particular is incompatible with the plain-http fallback origin the `X-Session-Token` path exists for, so it needs that retirement first. Happy to file either if you want them tracked.

`poetry run pytest tests/unit -q` → **9836 passed** locally. The integration test skips without MySQL here; the Integration lane in CI is where it runs.

**One thing still yours after merge**, unchanged from the Decision Comment: delete the `UI_API_TOKEN` repo secret. #965 carries that plus the `API_ACCESS_TOKENS` rotation, and it is intentionally **not** `agent:ready` — it touches `/opt/lem/.env` and repo secrets, which the pipeline is forbidden from editing.

gitchrisqueen:

The core of the PR holds up: the SPA ships no LEM secret, the CI canary has a positive control on both ends, `_require_api_and_admin` re-checks the bearer itself (so the middleware loosening does not open the admin-test routes), the route-identity walk really does cover 151 gated routes, and the integration test genuinely runs against MySQL in CI (verified in the run log, not assumed from a PASS). Every acceptance criterion on #950 is met and the two follow-ups (#957, #965) are filed and linked, so `Closes #950` is honest. Findings were in the accounting, which is what this PR is *for*.

- **`/api/admin/*` is three gates, not one — the PR and two docs claimed all of them take bearer + `X-Admin-Secret`.** Only the six `_require_api_and_admin` routes do. Nine run on `_require_admin` (`X-Admin-Secret` alone — including `/admin/youtube-token`, which installs an OAuth refresh token, and `/admin/user/location`); three on `_require_user_admin` (an admin session; they are SPA pages and never could hold a bearer). For those twelve the bearer came from the **middleware**, so this PR is what removed it, while the docs went on asserting the layer. Free today (every bearer value is public, so anyone holding `ADMIN_SECRET` holds one too) — not free once **#965** rotates them. Fixed: per-route table in `identity-and-sessions.md`, corrections in `AUTH_SECURITY_DESIGN.md` + `TESTING_ENGAGEMENT_API.md`, the PR body, and a new standing test that every `/api/admin/*` route reaches one of the three (the existing route-identity test would accept one that merely resolved a session — which every signed-in user has).
- **`test_allowlist_entries_are_real_gated_routes` could not fail.** It asserted `_api_token_required(path) or not _API_ACCESS_TOKEN_SET`; the unit env sets no `API_ACCESS_TOKENS`, so the right-hand side was always true and the "this allowlisted path is actually GATED" half never ran — moving `/api/carousel-templates` into `_PUBLIC_API_PREFIXES` would have gone unnoticed. Fixed: asks `_is_public_api_path` directly, which is env-independent.
- **One CSRF layer means one env var can delete it.** With the bearer gone, `SameSite=Lax` is the whole defence for the four query-parameter mutating routes — and `_samesite()` accepts `none`, which attaches the cookie to cross-site POSTs. The env table warned about `strict` and said nothing about `none`. Fixed: documented in the env table and in the CSRF section (with why LEM never needs it) until #957 lands.

Tests: `tests/unit/api` 1025 passed locally, including the new admin-gate invariant. Docs-only + test-only changes otherwise — no runtime behaviour altered by this round.

gitchrisqueen:

PR #959, opened 2026-08-03
fix(security): require X-LEM-Client on cookie-authenticated /api writes — restores the CSRF layer #950 removed (closes #957)

### LINKED ISSUE #957: fix(security): restore a CSRF layer for the four query-parameter mutating /api routes (follow-up of #950)

gitchrisqueen:
## What & why

#950 retired the shared `/api` bearer token from the SPA bundle. That was right — it was one of the
server's `API_ACCESS_TOKENS` inlined into a public bundle, so every visitor held it — but it had a
second, accidental job. A cross-site HTML form **cannot set an `Authorization` header**, even when
the attacker knows the value, and setting one from `fetch()` triggers a preflight that has nothing
to succeed against (no CORS middleware is installed). So the bearer was worthless as access control
and real as a **CSRF** layer.

Four mutating routes take query parameters and no body, so the "a JSON body needs a preflight" layer
never covered them either — `POST /create_weekly_content/`, `POST /invite_to_li_company_page/`,
`POST /aws_test_get_my_profile/`, `POST /automate_reply_commenting`. That left `SameSite=Lax`
holding them alone.

This adds the standard replacement: a **non-secret custom request header**, `X-LEM-Client`.

**Scope note:** the gate applies to **every** state-changing `/api` request that authenticates on
the session cookie — `POST`/`PUT`/`PATCH`/`DELETE`, not just those four. That is deliberate (owner
decision `1A`): covering only the four would leave the next query-only route uncovered, which is the
exact failure this issue exists to stop. The four are what made the gap visible, not the boundary.

## What changed

- **SPA** (`ui/src/api/client.ts`) — the one axios request interceptor sets `X-LEM-Client: spa` on
  every request. All SPA HTTP already goes through this client; audited and there are no raw
  `fetch()` callers, no `navigator.sendBeacon`, no `<form action=…method=post>` (every `<form>` in
  the UI is a React `onSubmit` handler that calls this client), no `EventSource`/WebSocket.
- **Server** (`api/main.py`) — `_require_client_header()` refuses a **state-changing** `/api`
  request that authenticated on the session **cookie** and did not send the header:
  **403 `client_header_required`** (never 401 — the SPA's interceptor reads any 401 as a dead
  session and signs the user out). It is called from the **one resolver**
  (`get_session_user_id`, cookie branch) *before* the scope check and before anything writes, so a
  forged request reaches no audit row, no enrolment promotion and no handler — the same argument as
  #905's scope narrowing: the credential this defends against is the one the browser attaches by
  itself, so the check belongs where that credential is read, not at ~150 call sites.
- **SPA 403 handling** — a bundle cached from before this shipped sends no header and gets 403 on
  every write. That is a **stale bundle, not a dead session**, so the response interceptor routes it
  into the existing stale-chunk guard (`recoverFromChunkError`, #743): ONE reload lands on a bundle
  that sends the header, and a second failure inside the cooldown surfaces "a new version was
  released — please refresh" instead of looping. It never signs the user out.
- **Docs** — `docs/identity-and-sessions.md` § CSRF restates the layer count with the new layer
  counted, adds `### X-LEM-Client is not a secret`, records the two load-bearing assumptions below,
  and says why `/auth/logout` sits outside this layer; `CLAUDE.md` gains the one-line invariant;
  `docs/AUTH_SECURITY_DESIGN.md` notes the design's "double-submit CSRF token" shipped as this
  header instead (a double-submit token needs a JS-readable cookie to echo, whose only guarantee is
  the one a custom header already gives).

### It is not a secret, on purpose

The value is a constant in a public bundle and is meant to be. The mechanism is that a cross-origin
form cannot set a header **at all**, whatever the value would have been — so the server checks
**presence** and never the value. Comparing it would buy nothing against an attacker who can read
the bundle, and would invite the next reader to rotate it like a token and put it in `.env`. Both
the server comment and the SPA comment say so.

### The two assumptions it rests on — both now pinned by a test

1. **The SPA is same-origin with the API, by construction rather than by deployment.** The axios
   `baseURL` is the **relative** `/api`, so every request is same-origin whatever the host (Vite dev
   server, docker-compose, the prod nginx edge) and a custom header on a same-origin request is
   never preflighted. An absolute `baseURL` would put the header behind a CORS preflight the server
   answers nothing to. Pinned by `the baseURL is relative` in `client.test.ts`.
2. **No CORS middleware is installed.** CORS with credentials would let a genuine cross-origin
   caller ask permission to send this header and reinstate the hole. Pinned by
   `test_no_cors_middleware_is_installed`, which fails on a one-line `app.add_middleware(...)` in
   some unrelated PR.

A third, smaller one: the request and the session cookie are stamped onto their ContextVars by the
**same** middleware block, so a live HTTP request can never carry the cookie without the request —
which is what makes "no request in scope → no-op" safe rather than a silent bypass (`_request_object`
defaults to `None`, so there is no `LookupError` either). Pinned by
`test_the_request_and_cookie_contextvars_are_set_together`. That no-op logs **DEBUG**, not a
warning: it is the expected shape for every Celery/direct caller, and warning on an expected no-op
files a defect for working code.

### The "ONE resolver" claim, made executable

`test_the_session_cookie_is_read_in_exactly_one_place` asserts the session cookie is read only by
`session_cookie_middleware` — a handler that read it on its own path would authenticate a caller
without ever passing this check. `POST /user/sessions/revoke` and `POST /auth/logout` both resolve
through `get_session_user_id`; logout additionally swallows a failed resolve on purpose (a logout
must always complete), and what covers **it** is the JSON-body layer — `LogoutRequest` is a body
model, and a cross-site form cannot produce `application/json`. That is recorded in the doc.

### Two deliberate exemptions

- **Reads.** CSRF is a forged *write*; with no CORS the attacker cannot read the response, and
  requiring a header on GET would break the browser's own credentialed navigations (a plain
  `<a href>` download, an `` src).
- **A bearer-authenticated caller.** Scripts, Postman and the admin tooling are not browsers. This
  exemption is **permanent** — it exists for non-browser callers, not as a rollout shim to retire
  once caches turn over, though it happens to cover that too: an SPA bundle cached from before #950
  still sends a bearer and no header. A request carrying an explicit `session_token` is likewise
  untouched — the attacker would have to know it, and it is httpOnly.

## Ordering vs #958 — resolved

Owner decision `2A` was "merge this first, then #958". #958 merged first anyway, and this branch is
rebased onto it (`4e99cc9`). The consequence of the swap is the good one: the SPA bearer is already
retired on `main`, so this layer is **live the moment it ships** rather than a no-op waiting on
#958. The two assumptions that moved underneath it — the `/api` gate now accepts a session
credential, and it reads the cookie for a presence check — are integrated and re-pinned.

## Testing

`tests/unit/api/test_csrf_client_header.py` (51 cases, real resolver — `get_session_user_id` is not
patched):

- a cookie-authenticated write refused **403** per **each of the four** query-parameter routes, each
  asserting the Celery task / `mark_queued` behind it was never called;
- the same four writes succeeding **with** the header;
- **every unsafe method** — `POST`/`PUT`/`PATCH`/`DELETE` asserted at the method set directly, plus
  a real `PUT /dm` and `DELETE /dm` through the routing + resolver path, so a refactor down to
  `{"POST"}` cannot pass green; `GET`/`HEAD`/`OPTIONS` pass straight through;
- the refusal landing **ahead of the scope check** — a scoped-away cookie session with no header is
  refused `client_header_required` and writes no audit row;
- a **bearer-authenticated** caller succeeding **without** it, per route; a guessed bearer refused
  by each layer in its own exact-code test (**401** at the credential gate when one is configured,
  **403** here when none is — which also pins the fail-closed case end to end);
- the two **`multipart/form-data`** writes refused per route — `POST /user/newsletter-draft/cover`
  and `POST /avatar/training` (which spends an avatar credit). Query parameters were never the only
  shape the JSON-body layer misses: a cross-origin caller can produce multipart with no preflight
  too, so these are the concrete reason the gate is scoped to every state-changing
  cookie-authenticated request rather than to a route list;
- reads untouched, explicit-token writes untouched, an anonymous write still **401** (not 403);
- any header value passes and the **name is case-insensitive** (presence is the mechanism) while an
  empty value does not;
- no HTTP request in scope (Celery, direct calls) is a no-op;
- the three structural assumptions above.

`ui/src/api/client.test.ts` (9 cases) proves the header rides on every method out of the one axios
client with and without a session token held, that the `baseURL` is relative, and that a 403
`client_header_required` reloads once → then messages → and never clears `lem_session`/`lem_email`,
while any other 403 (`step_up_required`) is left to the caller's error UI.
`tests/unit/api/test_auth_hardening.py` gains the header on its one cookie-authenticated write case.

Green locally: **9852 passed** (unit lane) and **400 passed / 50 files** (SPA lane, run on node 22
rather than this box's node 18), `tsc -b` and `eslint` clean on the changed files.

Closes #957


gitchrisqueen:

Held (`needs-human`, risk: **security**). This changes what the server will accept on every
cookie-authenticated write, so a wrong call here is a self-inflicted outage rather than a bug — and
its value depends on when it lands relative to #958.

### 1. How wide should the `X-LEM-Client` requirement be?

The issue asked this explicitly ("Consider whether to scope it to state-changing methods only, or to
every gated request"). As pushed, it is **A**.

- **A. State-changing methods only** (`POST`/`PUT`/`PATCH`/`DELETE`), cookie-authenticated only — CSRF
  is a forged write, and with no CORS the attacker cannot read a response anyway. Leaves the
  browser's own credentialed GETs working: the `<a href>` extension download, `` srcs, any
  future direct link.  ✅ *recommended*
- **B. Every gated `/api` request, reads included** — marginally stricter, and breaks any credentialed
  GET a browser makes without going through axios. Nothing in the SPA does that today, so the cost
  is a future foot-gun rather than a present outage.
- **C. Only the four query-parameter routes** — smallest blast radius, but the layer then does not
  cover the next query-only route somebody adds, which is the exact failure this issue exists to
  stop.

### 2. Merge order relative to #958 (issue #950)?

#958 is still open, so `main` still ships the bearer token. That matters because a bearer-
authenticated request is exempt from this header — so **until #958 merges, this layer is a no-op in
production**, and after #958 merges the four routes are behind `SameSite=Lax` alone until this lands.

- **A. Merge this one first, then #958** — the four routes are never down to one layer, and an SPA
  bundle cached across either release keeps working (it sends a bearer, which is exempt). Both PRs
  touch `client.ts`'s interceptor and the § CSRF doc section, so #958 needs a small rebase.  ✅ *recommended*
- **B. Merge #958 first, then this** — #950 ships sooner; the window where the four query-parameter
  routes sit behind `SameSite=Lax` alone stays open for as long as this PR waits, and a bundle
  cached inside that window sends neither credential and would get 403s on writes until it reloads.
- **C. Hold this until #958 is reviewed and merge them together** — no window at all, at the cost of
  both sitting parked on one review.

**My recommendation: `1A 2A`.** A is the layer that actually generalises to the next query-only route
without touching browser navigations, and shipping it before #958 means the four routes are never
down to a single layer and no cached bundle is ever caught between the two releases.

<sub>Full rationale, the exemption list and the "this header is not a secret" contract are in the PR
body and `docs/identity-and-sessions.md` § CSRF.</sub>

gitchrisqueen:
ok

gitchrisqueen:
Addressed your review — thanks. You replied **`ok`**, so I took both recommendations: **1A** (state-changing methods only, cookie-authenticated only) and **2A** (this merges before #958). Both are the shape that was already pushed, so nothing about the design changed; what follows is the review pass on top of it, driven by the automated review's blocking items.

Pushed in 88506cca.

### The two assumptions you'd want proven

**1. Same-origin.** Not a deployment property — the axios `baseURL` is the **relative** `/api`, so every SPA request is same-origin whatever the host (Vite dev server, docker-compose, the prod nginx edge), and a custom header on a same-origin request is never preflighted. There is no deployment path where it isn't. Two tests now hold it:

- `the baseURL is relative` (`client.test.ts`) — an absolute `baseURL` would put the header behind a CORS preflight nothing answers.
- `test_no_cors_middleware_is_installed` — CORS with credentials would let a genuine cross-origin caller *ask permission* to send this header and reinstate the hole. It fails on a one-line `app.add_middleware(...)` in some unrelated PR.

**2. The ONE resolver.** Made executable rather than asserted in prose: `test_the_session_cookie_is_read_in_exactly_one_place` fails if any handler reads the cookie on its own path. The two you called out are both covered — `POST /user/sessions/revoke` resolves through `get_session_user_id`, and so does `POST /auth/logout`, which additionally swallows a failed resolve on purpose (a logout must always complete). What covers **logout** is the JSON-body layer: `LogoutRequest` is a body model, and a cross-site form can only send `text/plain` / `x-www-form-urlencoded` / `multipart`. That's now recorded in `docs/identity-and-sessions.md`. No admin dependency resolves the cookie separately.

### The rest

- **`_request_object`** already declares `default=None`, so there's no `LookupError`. The stronger point on the fail-open: the request and cookie ContextVars are set by the **same** middleware block, so a live HTTP request can't carry the cookie without the request — `test_the_request_and_cookie_contextvars_are_set_together` pins that. Reaching the `None` branch therefore means a non-HTTP caller. It now logs, at **DEBUG** rather than warning: a warning there would fire on every Celery/direct call, and per the escalation contract warning on an expected no-op files a defect for working code.
- **PUT/PATCH/DELETE** — covered at the method set directly (so a refactor to `{"POST"}` can't pass green) *and* through real `PUT /dm` and `DELETE /dm` routes; `GET`/`HEAD`/`OPTIONS` asserted to pass through.
- **Ordering vs the scope check** — pinned: a scoped-away cookie session with no header is refused `client_header_required`, and `record_auth_event` is asserted never called.
- **Blast radius** — you're right, and the title was wrong. Retitled to name the real scope ("cookie-authenticated /api writes"), with a scope note in the body. Audit done: no `navigator.sendBeacon`, no raw `fetch()`, no `EventSource`/WebSocket, and every `<form>` in the SPA is a React `onSubmit` calling this client — no `action`/`method="post"` anywhere.
- **The loose `in (401, 403)`** — split into two exact-code tests, one per layer. The 403 half doubles as the end-to-end fail-closed assertion you asked for in (14): no configured bearer ⇒ an `Authorization` header buys no exemption.
- **SPA 403 handling** — added, and it reuses the existing stale-chunk guard (`recoverFromChunkError`, #743) rather than growing a second reload path: one reload lands on a bundle that sends the header, a second failure inside the cooldown surfaces "a new version was released — please refresh". Never a sign-out — that's the 401 branch, which is the whole reason the refusal is a 403.
- **Test hygiene** — `_FakeRequest` moved to the top, fully typed, and backed by Starlette's own case-insensitive `Headers`; a route-level test sends `x-lem-client` lowercase. Fixtures and test methods annotated. `try/finally` around the adapter stub in `client.test.ts`.
- **Docs** confirmed landed (they were truncated in your view): § CSRF records both narrowings, both assumptions, the logout carve-out, and the SPA reload behaviour.

**Deliberately not done, and why:** (13) no expiry on the bearer exemption. It isn't a rollout shim — it exists permanently for non-browser callers (scripts, Postman, admin tooling), and it only *also* happens to cover pre-#950 cached bundles. Putting a sunset on it would eventually break the callers it's actually for; the doc and the code comment now say that explicitly instead. (15) not logging the header value — no use for it today, and the "presence is the entire check" contract is easier to keep if nothing ever reads it.

**Testing:** 9852 passed in the unit lane; the SPA lane ran this time (fetched node 22 — this box is on 18) at 400 passed / 50 files, with `tsc -b` and `eslint` clean. Running it caught a real bug in my own new tests, which is why it was worth doing: a custom axios adapter has to settle the status itself, so the 403 cases were silently resolving instead of rejecting.

gitchrisqueen:

Reviewed the diff, the resolver it hooks into, both SPA interceptors, and the issue's acceptance criteria. The design holds up: the gate is in the ONE resolver on the cookie branch, ahead of the scope check and any write; a bearer/explicit-token caller is exempt; refusal is 403 so a stale bundle reloads instead of being signed out. I confirmed independently that the SPA has no `fetch`/`sendBeacon`/`<form method=post>` outside the axios client, that the extension authenticates on an explicit `session_token` (so it is untouched), that no handler reads the cookie for itself, and that the only mutating `GET`s are the OAuth callback and an idempotent training-status sync. Ran `tests/unit/api` locally: 1076 passed.

- **CLAUDE.md was over its hard cap → the `CLAUDE.md ≤ 40,000 chars` check was RED** (40,308 / 40,000; `main` sits at 39,878, so there were only 122 chars of headroom and the new entry spent 430). Rewrote the CSRF entry as the map line it should be — header name, `client_header_required`, "NOT a secret" — with the posture staying in `docs/identity-and-sessions.md`, plus three lossless trims in the same bullet. Now 39,981 with every invariant kept.
- **Query parameters were never the only shape the JSON-body layer misses, and nothing said so.** A cross-origin caller can produce `multipart/form-data` with no preflight either (`<form enctype=…>`, or a `no-cors` `fetch` with a `FormData` body), and two mutating routes take exactly that: `POST /user/newsletter-draft/cover` and `POST /avatar/training` — the latter spends an avatar credit and starts a LoRA training run. The gate **does** cover them, precisely because it was scoped to every state-changing cookie-authenticated request; but the doc enumerated "the four" as the complete no-preflight set, and that same doc says it is the thing to revisit first when another form-encoded route appears. A later narrowing back to a route list would have uncovered both silently. Named in the design comment and the doc, and pinned by a refused case per route (`TestTheMultipartWrites`, asserting neither handler was reached).
- **The design comment cited a `client.test.ts` test name that does not exist** (`test_the_api_client_baseurl_is_relative`; the vitest case is "is same-origin by construction — the baseURL is relative"). A load-bearing assumption pointing at nothing is worse than no pointer. Fixed.
- **The PR body's "Ordering vs #958" was stale and wrong in the security-relevant direction** — #958 merged first, so the body's claim that "this layer is a no-op in production until #958 lands" no longer held; it is live on merge. Section rewritten, and the testing notes updated for the multipart cases.

Close is honest: all four acceptance criteria are covered by the diff, and #957 declares no later phase, so `Closes #957` stands.

gitchrisqueen:

PR #960, opened 2026-08-03
fix(ui): label the Engagement Roster per-week field (closes #956)

### LINKED ISSUE #956: Engagement Roster: per-week number field is unlabeled in the SPA (bare 'N /wk')
## Problem

In Account → Engagement Roster (`src/cqc_lem/ui/src/pages/account/EngagementRosterCard.tsx`), each roster row renders: profile URL, name, type — then a bare `<input type=number>` suffixed only by `/wk` (line ~144-148). The field is `max_comments_per_week` and it has an `aria-label`, but there is **no visible label or help text**, so a sighted user has no idea what the number controls. Reported directly by the owner: *"it shows the url, name, type, then another number field followed by '/w'. I don't know what the number field is — it's unclear to the user."*

## Fix

- Add a visible label (e.g. a column header row for the roster list, or an inline label "Max comments/wk") — not just the suffix.
- A `title`/tooltip on the field explaining: "LEM will comment on this account's recent posts at most this many times per week (0 pauses the account)."
- Keep the existing `aria-label` in sync with the visible label.
- The card's intro paragraph (line ~118) already explains the concept — the fix is connecting that sentence to the actual field.

## Acceptance

- A user seeing the roster for the first time can tell what the number does without reading code.
- Component test updated (there is existing coverage for this card's rendering).

🤖 Filed by Claude Code on owner request (2026-08-03).

gitchrisqueen:
## What & why

In Account → Engagement Roster, each row rendered profile URL, name, type — then a bare
`<input type=number>` whose only adornment was the suffix `/wk`. The field is
`max_comments_per_week` and it had an `aria-label`, but nothing visible said so, so a sighted user
had no way to know what the number controlled (owner-reported: *"I don't know what the number field
is — it's unclear to the user."*).

## Changes — `EngagementRosterCard.tsx`

- **Visible column headers** above the rows on wide screens: Profile URL / Name / Type /
  **Max comments/wk** / On. The grid was rebalanced (5-3-2-1-1 → 4-3-2-2-1) so the number column is
  wide enough to carry a real label.
- **Mobile keeps its own label.** The rows stack below `sm`, where a header row would not line up, so
  the field renders the full `Max comments/wk` label inline there and the short `/wk` suffix only on
  wide screens where the header already names it.
- **Tooltip** on the input and on its header: *"LEM will comment on this account's recent posts at
  most this many times per week (0 pauses the account)."*
- **`aria-label` and the visible label are now the same words**, both from one `PER_WEEK_LABEL`
  constant, so a screen-reader user and a sighted user are told the same thing.
- **Intro paragraph names the field** instead of alluding to it ("the set number of times per week"),
  and states that 0 pauses an account without removing it — connecting the explanation the card
  already had to the actual control.

## Testing

New `EngagementRosterCard.test.tsx` (the card had **no** existing component test — the issue assumed
one; `SettingsHub.test.tsx` only stubs the card out). Covers: the number column has a visible header
label, the input carries the explanatory `title`, the accessible name matches the visible label and
still reads the target's value, the prose states that 0 pauses an account, and editing the ceiling
still `PUT`s the roster.

Not run locally: this worktree has no `node_modules` and only Node 18 (vitest 4 needs ≥20), so the
vitest lane is CI-verified. The component was type-checked against the sibling checkout's
`node_modules` with no errors of its own.

Closes #956

### DIFF
--- a/src/cqc_lem/app/run_automation.py
+++ b/src/cqc_lem/app/run_automation.py
@@ -61,7 +61,7 @@ from cqc_lem.utilities.db import get_user_password_pair_by_id, get_user_id, inse
     get_engager_candidates, get_profile_facts, count_invites_sent_today, ConnectionRequestStatus, \
     CatchupTouchStatus, insert_catchup_touch, has_catchup_touch, get_catchup_touch, \
     update_catchup_touch_status, count_catchup_touches_sent_today, max_catchup_touches_allowed, \
-    get_engagement_targets, record_target_engagement, ENGAGEMENT_TARGET_WEEKLY_DEFAULT, \
+    get_engagement_targets, record_target_engagement, resolve_weekly_cap, \
     record_follower_stat, get_linkedin_profile_url_by_user_id, \
     get_comment_outcome_targets, record_comment_outcome, \
     count_user_comments_on_post_url, get_post_age_minutes, get_story_bank_entries, \
@@ -1362,7 +1362,7 @@ def select_roster_targets(targets: list, limit: int) -> list:
     for t in targets or []:
         if not t.get("active", True):
             continue
-        cap = int(t.get("max_comments_per_week") or ENGAGEMENT_TARGET_WEEKLY_DEFAULT)
+        cap = resolve_weekly_cap(t.get("max_comments_per_week"))
         if int(t.get("comments_this_week") or 0) >= cap:
             continue
         eligible.append(t)
@@ -1639,7 +1639,7 @@ def comment_on_roster_posts(driver, wait, my_profile: LinkedInProfile, user_id: 
             continue
         time.sleep(random.uniform(2, 4))
         stats["targets_visited"] += 1
-        weekly_left = max(0, int(target.get("max_comments_per_week") or ENGAGEMENT_TARGET_WEEKLY_DEFAULT)
+        weekly_left = max(0, resolve_weekly_cap(target.get("max_comments_per_week"))
                           - int(target.get("comments_this_week") or 0))
         allowed_here = min(_ROSTER_MAX_POSTS_PER_AUTHOR, weekly_left, max_posts - stats["posted"])
         posted_here = 0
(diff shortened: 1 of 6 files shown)


gitchrisqueen:

- **The new copy was false: a per-week cap of `0` did NOT pause the account.** The PR's whole point is telling the operator what the number does, and both the new tooltip and the new prose promise *"0 pauses the account"*. The backend read the cap as `value or ENGAGEMENT_TARGET_WEEKLY_DEFAULT` in three places — `db._clean_target_row` (db.py:5161), `run_automation.select_roster_targets` (:1365) and `comment_on_roster_posts` (:1642) — and in Python `0 or 2` is `2`, so an explicit 0 was read as *unset* and the author got the default two comments a week instead. Worse, because `_clean_target_row` is also what feeds the SPA, the roster refetched after save and showed **2**, silently reverting the number the operator had typed. Fixed by adding `db.resolve_weekly_cap()` as the ONE place the cap is resolved (`None` → default, explicit `0` survives) and routing all three sites through it. Tests: `test_a_zero_cap_reads_back_as_zero` / `test_a_missing_cap_falls_back_to_the_default` (db lane) and `test_a_zero_cap_pauses_the_author` (roster lane) — both would have failed before the change. Note `upsert_engagement_targets` and the API's `_clamp_weekly_cap` were already 0-safe, so the value was being stored correctly all along; only the reads were wrong.

Checked and found clean: the `4-3-2-2-1` grid still sums to 12 and matches the header row; the old `aria-label="Max comments per week"` string has no other referent anywhere in the repo (grepped `.ts/.tsx/.py/.md`) so nothing else broke; `PER_WEEK_LABEL` is unique as an accessible name (the wrapping `<label>`'s own text normalises to `Max comments/wk/wk`, so `getByLabelText` can't double-match); the card registers under the same `engagement-roster` save-section id; `useRegisterSaveSection` is a no-op outside a provider so the test's bare harness is valid; `sectionSaveCallbacks` reaches the mocked `capture`. ESLint is clean on both changed UI files (the one error in `EngagementRosterCard.tsx` is the pre-existing `react-hooks/set-state-in-effect` at line 66, untouched by this PR), and `tsc -p tsconfig.app.json` reports nothing on either file. The vitest lane could not run here (no `vitest`/`@testing-library` in any local `node_modules`, and Node 18 vs vitest 4) — CI is the source of truth for it. Full `tests/unit` lane: **9810 passed**.

Close is honest: issue #956 has no phased scope and every bullet under **Fix** / **Acceptance** is covered by the diff.

gitchrisqueen:

Issue #771, opened 2026-07-28
R3 spike #406 / #747 follow-up: live-validate article-editor selectors for newsletter publishing


gitchrisqueen:
## Why this exists

Issue #747 closed the silent-no-op half of the R3 spike (#406): newsletter publishing now emits `log_error` with the failed step when `_fill_and_publish_article` returns `None`. The spike's second recommendation — folding the article-editor selectors into a live validation sweep — was intentionally left out of #747 because it needs a real LinkedIn session.

This issue tracks that remaining work so it is not dropped the way #406's recommendations were.

## Scope

- Live-validate the four article-editor steps that `_fill_and_publish_article` reports: `article_title`, `article_body`, `article_next`, `article_publish`.
- Build the selector-resolution ladder using the same TEXT/href/aria-label principles as `utilities/linkedin/message_thread.py` (#731) so it survives LinkedIn SDUI churn. Do NOT key on `feed-shared-*` / `comments-comment-*` class names.
- Capture fallback routes for each step (e.g., title field via `aria-label` or `placeholder`, publish button via text + `disabled` state, next button as either primary or secondary action).
- Add `--article-editor-url` support to `scripts/linkedin_live_validation.py` that reports each step's `reply_state`-style verdict: `OK` / `MISSING` / `UNKNOWN`.
- Wire the validated selector map back into `_fill_and_publish_article` so a missing step returns a precise `failed_step` and, on failure, the task logs the step name (already implemented in #747 / PR #764).

## Acceptance

- [ ] A live LinkedIn session can open the article editor and report each of the four steps present and actionable.
- [ ] The selector map has at least two independent routes per step where feasible.
- [ ] Unit tests cover the fallback-order logic with mocked DOM (no live session required).
- [ ] PR references this issue and #747.

## Related

- #406 — original spike
- #747 — alert on `None` path (closed by PR #764)
- #731 — message-thread resolution ladder (use the same pattern)
- #478 — SDUI comment-thread selector map

gitchrisqueen:
PR #789 is ready and parked as `needs-human`. A Decision Comment with live-validation options is on the PR: https://github.com/christopherqueenconsulting/linkedin_engagement_manager/pull/789#issuecomment-5113888444

gitchrisqueen:
[...]

PR #981, opened 2026-08-03
feat(newsletter): implement blog-align — align_with_blog now reaches the generator (closes #967)

### LINKED ISSUE #967: feat(newsletter): implement blog-align — align_with_blog toggle currently changes nothing (blog_content always None)
src/cqc_lem/app/run_automation.py:2171-2172:
`# TODO(blog-align): when align_with_blog, fetch recent blog/sitemap content to pass as blog_content. For now the edition is generated from topic + profile.`

`align_with_blog` is a real user-facing toggle (NewsletterCard.tsx:203) defaulting TRUE (api/main.py:1033, db.py:8204) — users flipping it are silently ignored; `blog_content` is always None.

Work: when the toggle is on, fetch recent blog/sitemap content and pass it to `generate_newsletter_edition`. Related: #771 covers the article-editor selectors, a different half of newsletters.

Todo-sweep dedup key(s), pre-seeded so the weekly sweep never double-files this work: `todo-sweep-1a6c7b50d9d8`

gitchrisqueen:
## What & why

`align_with_blog` is a user-facing toggle in the SPA (`NewsletterCard.tsx`) that **defaults ON**, and it changed nothing: every path into `generate_newsletter_edition` passed `blog_content=None`, guarded by a `TODO(blog-align)`. Users flipping it were silently ignored.

This implements it.

### The one place the toggle becomes text

New `src/cqc_lem/utilities/blog_source.py` — `resolve_blog_source(user_id, settings)`:

- **Blog URL first, sitemap as the fallback.** Both go through the SAME fetchers the `blog_summary` / `website_content` post types already use (`get_main_blog_url_content`, `fetch_sitemap_urls` / `filter_relevant_urls` / `extract_page_content` in `run_content_plan`) — one scraper for the app, not a parallel helper.
- **HTML is stripped to prose before it's handed over.** WordPress returns rendered HTML and the scrape fallback returns whole pages; without this, the generator's 4000-char source budget is mostly markup.
- **Best-effort, never blocking.** Toggle off, nothing configured, unreachable, or empty → `None`, and the edition is written from topic + profile exactly as before. It never raises.
- **Escalation contract:** the toggle defaults ON, so "on but no blog/sitemap set" is an *expected no-op* → DEBUG. Only a **configured** source that reads back empty warns.

### Wired into all three generation paths

| Path | Behaviour |
|---|---|
| `run_scheduler._topup_newsletter_drafts_for_user` | resolved **per edition** — each call picks a different recent article, so a batch of queued drafts doesn't rehash one. The first empty resolve in a run latches off so we don't pay for N failed fetches. |
| `run_scheduler.regenerate_newsletter_edition` | resolved once, before the LLM attribution block |
| `run_automation.auto_publish_newsletter_edition` | replaces the `TODO(blog-align)` |

## Testing

- New `tests/unit/utilities/test_blog_source.py` (14 tests): markup stripping, bytes/paragraph-list inputs, toggle-off touches no network, blog→sitemap fallback on both failure and empty, sitemap skipping unreadable pages, carry-budget truncation, and the DEBUG-vs-WARNING split between "nothing configured" and "configured but unreadable".
- `tests/unit/app/test_newsletter_publish.py`: per-edition resolution, the empty-resolve latch, toggle-off, regenerate wiring, and the direct publish task.
- Full unit lane green locally: **10002 passed**.

## Scope

Everything the issue asks for is in this PR — no later phase. `#771` (article-editor selectors) is a separate, already-filed half of newsletters and is untouched here.

Closes #967


gitchrisqueen:

Reviewed the full diff plus the seams it touches (`get_main_blog_url_content` / `fetch_sitemap_urls` / `filter_relevant_urls` / `extract_page_content` in `run_content_plan`, `generate_newsletter_edition`, and the three call sites). Fixed in 92d5a00f; full unit lane green (10004 passed).

- **Sitemap fallback handed every edition in a run the SAME page** — `_from_sitemap` walked `urls[:3]` from a deterministic `filter_relevant_urls` result, so a user on the sitemap path with `max_queued_drafts=3` got three drafts repurposing one identical article, at 3× the fetch cost. That is exactly the rehash the per-edition design (and the CLAUDE.md invariant this PR adds) claims to prevent; the blog path is fine because `get_main_blog_url_content` does `random.choice`. The sitemap path now draws its candidates at random the same way, with a seeded test that fails against the old ordering.
- **`resolve_blog_source` could raise, despite documenting "never raises"** — `get_user_blog_url` / `get_user_sitemap_url` were unguarded and `get_db_connection()` is documented to raise `mysql.connector.Error`. In `regenerate_newsletter_edition` the resolve sits OUTSIDE the try, and in `_topup_newsletter_drafts_for_user` outside any guard, so a transient pool error would fail a whole regeneration / abort that user's top-up over best-effort enrichment. Now guarded, with a test.
- **One condition, two warnings** (`utilities/CLAUDE.md`: "one condition gets ONE warning", precedent #877) — a failed fetch warned where it was detected and then again as "no source material could be read", so one incident fingerprints two grouped `$exception`s and can file two GitHub issues. The old test asserted `warn.call_count == 2`, codifying it. The caller now warns only for the case nothing else covers: a source that fetched fine and read back empty. The DEBUG-vs-WARNING split for "toggle on, nothing configured" is unchanged and correct.

Also added the missing type hint on `_plain_text`'s parameter.

Checked and found clean: `_plain_text` does not leak `<script>`/`<style>` bodies into the writer's 4000-char budget (bs4 ≥4.9 excludes them from `get_text()`; locked at 4.15.0 — verified by execution, not assumption); the `blog_align` latch correctly stops refetching after the first empty resolve without disabling the toggle-off path; `blog_content` is passed by keyword everywhere, so the test doubles' differing parameter order is harmless; the close is honest — issue #967 has no acceptance checkboxes and no later phase, and #771 is separately filed and untouched.

gitchrisqueen:

PR #982, opened 2026-08-03
feat(dm): implement recommendation + collaboration appreciation-DM sources (closes #968)

### LINKED ISSUE #968: feat(dm): implement recommendation + collaborator sources for appreciation DMs (currently permanent empty-dict stubs)
src/cqc_lem/app/run_automation.py:4542-4557 — two of the three appreciation-DM triggers are stubs that can never fire:
- `get_recent_recommendations`: "Scraping the LinkedIn Recommendations Received section is not yet implemented; returns an empty dict until a scraper is added here."
- `get_recent_collaborators`: "No structured LinkedIn data source exists for this yet; returns an empty dict"

The appreciation-DM feature is advertised (CLAUDE.md: "appreciation (connection / recommendation / collaboration)") and runs on a live beat (`send-appreciation-dms`), but only new-connection events ever trigger; DM templates for the other two types exist and are dead. Needs a live-grounding probe of the Recommendations Received section before the scraper ships.

Todo-sweep dedup key(s), pre-seeded so the weekly sweep never double-files this work: `todo-sweep-3f7cf17793ef`

gitchrisqueen:
## What & why

`automate_appreciation_dms_for_user` advertises three triggers, but two of them
(`get_recent_recommendations`, `get_recent_collaborators`) were permanent `return {}` stubs. Only
new-connection events ever fired, and the `recommendation_received` / `collaboration` DM templates
were dead code. This implements both sources.

**Two sources, both STANDING lists — that is what drives every design choice here.**
- `get_recent_recommendations` reads the user's own profile → `/details/recommendations/` (Received
  tab). A recommendation never leaves that page.
- `get_recent_collaborators` reads the mentions notification feed (`/notifications/?filter=mentions`)
  — the nearest structured signal LinkedIn actually exposes for "we did something together": someone
  put this user's name in their own post or comment. A card that does not SAY it was a mention is
  skipped. (LinkedIn has no first-class "collaboration" event; see the decision comment.)

Selectors follow the SDUI rules: ordered `data-view-name` → semantic-with-`/in/`-link → legacy
fallback chains, never a hashed class alone; every miss is a DEBUG no-op.

**Undated is SKIPPED, never thanked.** An undated card could be from 2018, so
`_parse_recommendation_date` / `_parse_relative_age_days` return `None` and the card is dropped.
Only what falls inside `APPRECIATION_LOOKBACK_DAYS` (default 30) counts. The one thing that *does*
warn is cards rendering while NONE of them date — that is format drift, and in production it reads
as "no recent recommendations" forever, i.e. a silently dead trigger.

**`appreciation_touches` is the claim that makes this safe.** The beat re-queues itself every ~60s
inside its window, so a standing list without a durable claim is a DM a minute, not a thank-you.
One row per `(user, person, event_type)`; the unique key is the guarantee.
`_dispatch_appreciation_dms` checks `has_appreciation_touch` **before** writing the message (a
repeat costs no LLM call) and `claim_appreciation_touch` **after** it and before the send (a missing
template never burns a person's one shot at being thanked). A ledger we cannot read reads as "don't
send" — a missed thank-you is recoverable by a human, one sent twenty times is not.
`connection_accepted` flows through the same dispatcher and now gets the same protection.

**OFF until live-grounded.** `APPRECIATION_SOURCES_ENABLED` (default `false`, read at the call
site) gates both scrapers. A scraper that finds nothing is a quiet no-op; one that finds the WRONG
cards DMs real people. `scripts/linkedin_live_validation.py --appreciation-sources` grounds both
surfaces read-only — it messages nobody and claims no ledger row — and reports, per card, exactly
what production would do with it.

## Files

- `compose/local/database/migrations/V20260803073500__add_appreciation_touches.sql` — additive, new
  table only, no existing DDL touched.
- `src/cqc_lem/utilities/db.py` — `claim_appreciation_touch` (INSERT IGNORE, fails closed) +
  `has_appreciation_touch`.
- `src/cqc_lem/app/run_automation.py` — the two sources, the locator chains, the date parsers, the
  dispatcher claim.
- `scripts/linkedin_live_validation.py` — `--appreciation-sources` probe.
- `docs/engagement-automation.md`, `CLAUDE.md`, `.env.example`, live-validation skill.

## Testing

- `tests/unit/app/test_appreciation_sources.py` (new, 37 cases): flag defaults off, lookback
  parsing, both date parsers incl. undated→`None`, window filtering, own-profile-link exclusion,
  non-mention/company-actor rejection, drift warning fires exactly once, and the dispatcher dedup
  (already-thanked costs no LLM call, an ungranted claim never sends, a missing template does not
  burn the claim).
- `tests/unit/utilities/test_db_appreciation_touches.py` (new): claim granted once, duplicate
  refused without raising, DB error fails closed.
- `tests/unit/test_linkedin_live_validation.py`: probe verdicts + a full probe run.
- `tests/unit/app/test_network_activation.py`: updated for the new dispatcher ledger calls.
- Full unit lane green locally: **10050 passed**.

**Follow-up: #1002** — the Recommendations Received chain matched 0 cards on the live run and
`APPRECIATION_SOURCES_ENABLED` is still `false` in prod. Both need a real LinkedIn session and an
env change, so they are tracked there rather than left in prose inside a closed issue.

Closes #968


gitchrisqueen:

Held (`needs-human`, risk: **live-linkedin** + a new migration). This PR makes LEM read two live
LinkedIn surfaces it has never read before and turn what it finds into **real DMs to real people**.
Both scrapers ship **OFF** (`APPRECIATION_SOURCES_ENABLED=false`) precisely so merging cannot start
that — the flip is yours, after a grounding run. The pipeline is not allowed to run the live probe.

### 1. Is "someone mentioned/tagged you" the right trigger for the `collaboration` DM?
LinkedIn exposes **no** collaboration event. The mentions notification feed
(`/notifications/?filter=mentions`) is the nearest structured thing it does expose. Recommendations
Received, by contrast, is unambiguous and needs no judgement call.
- **A. Keep mentions as the collaboration source** — someone who publicly featured you gets one
  thank-you, once ever, inside a 30-day window. Ships both triggers.  ✅ *recommended*
- **B. Ship recommendations only** — I revert `get_recent_collaborators` to a stub and file a
  follow-up for an explicit *manual* collaboration trigger (you mark the person in the SPA). The
  `collaboration` template stays dead until then.
- **C. A different source you name** — tell me what counts as a collaboration and I'll implement it.

### 2. If 1A: should the stock `collaboration` DM template change?
Today it reads *"it was a pleasure collaborating with you! Your contributions made a real
difference…"* — written for a project, not for being tagged in a post.
- **A. Reword the default to fit a mention** — e.g. *"thanks for the mention — genuinely
  appreciated. What are you working on at the moment?"* Only changes the **default**; anyone who
  customised their template keeps theirs.  ✅ *recommended*
- **B. Leave it as-is** — the wording lands slightly off for a mention, but nothing changes for
  existing users at all.

### 3. How do you want this to reach production?
- **A. Merge with the flag OFF, then ground, then flip** — migration applies, code is inert. You run
  `sudo docker exec -i celery_worker_selenium python - --user-id 1 --appreciation-sources <
  scripts/linkedin_live_validation.py`, paste the JSON here, I fix any selector drift, and only then
  do you set `APPRECIATION_SOURCES_ENABLED=true` in `/opt/lem/.env`.  ✅ *recommended*
- **B. Merge and flip ON in the same deploy** — appreciation DMs start immediately on whatever the
  scrapers find. Fastest, but the first live read is also the first live send.
- **C. Hold the PR until you've run the probe** — nothing merges, including the migration, until the
  selectors are confirmed against your real profile.

**My recommendation: `1A 2A 3A`.** Mentions is the only structured collaboration signal LinkedIn
has, so shipping it behind an off-by-default flag costs nothing and can be reverted in one env var;
the reworded default keeps the message honest for what actually triggered it; and 3A separates
"schema + inert code lands" from "LEM starts DMing people", which is the split that matters when the
scraper has never seen the real DOM.

<details>
<summary>What is already guaranteed regardless of your answer</summary>

- **Nobody is thanked twice.** New `appreciation_touches` table, unique on
  `(user, person, event_type)`. The beat re-queues itself every ~60s, so this claim is the whole
  safety property. Checked before the message is written, claimed before the send; an unreadable
  ledger reads as "don't send". `connection_accepted` now gets the same protection.
- **Nobody outside 30 days is thanked**, and an **undated card is skipped**, never treated as
  today's — otherwise a first run would DM everyone who ever wrote a recommendation.
- **The probe is read-only**: it messages nobody and claims no ledger row.
- The migration is additive (one new table, no existing DDL touched).
</details>

gitchrisqueen:
Ok

gitchrisqueen:
## 🔬 Live grounding run — `--appreciation-sources` (2026-08-03, owner-authorized, read-only)

Owner had the probe run from an interactive session on the VPS (branch code staged into
`celery_worker_selenium` via `PYTHONPATH`, script from this branch). Nothing was messaged; the
ledger was only read.

```json
{
  "user_id": 1,
  "appreciation_sources": {
    "own_profile_url": "https://www.linkedin.com/in/christopherqueen",
    "recommendations_received": {
      "url": "https://www.linkedin.com/in/christopherqueen/details/recommendations/",
      "current_url": "https://www.linkedin.com/in/christopherqueen/details/recommendations/",
      "lookback_days": 30,
      "cards": 0,
      "dated": 0,
      "people": [],
      "rows": [],
      "verdict": "no cards resolved — either the surface is genuinely empty or the card locator chain has rotated; production sends nothing either way"
    },
    "mentions": {
      "url": "https://www.linkedin.com/notifications/?filter=mentions",
      "current_url": "https://www.linkedin.com/notifications/?filter=mentions",
      "lookback_days": 30,
      "cards": 1,
      "dated": 1,
      "people": [
        {
          "profile_url": "https://www.linkedin.com/in/utkarsh%2Dtiwari%2D98164814b",
          "name": "",
          "age_days": 0.0,
          "in_window": true,
          "already_thanked": false,
          "text": "Status is online\nUnread notification.\nUtkarsh Tiwari mentioned you in a comment in STARTUP SPECIALISTS NETWORK GROUP Entrepreneurs Business Startups Social Digi"
        }
      ],
      "verdict": "1 card(s), 1 dated, 1 inside the 30-day window — production would thank those"
    }
  }
}
```
(`rows` omitted above — identical to `people` for this run.)

### Reading — three things to fix/consider before the env flip

1. **Mentions surface resolves** (card, date, actor URL) — the chain is live. ✅
2. **Actor `name` came back empty** on the one live card. The DM template presumably wants a first
   name; if `name` stays empty the fallback path (or a skip) should be deliberate, not accidental.
   The card text plainly contains "Utkarsh Tiwari", so the name locator missed while the URL
   locator hit.
3. **`profile_url` is percent-encoded** (`utkarsh%2Dtiwari%2D98164814b` — `%2D` is `-`). Everything
   that keys on profile_url (the `appreciation_touches` ledger, `has_appreciation_touch` dedup,
   DM history) will treat the encoded and decoded forms as different people. Normalize
   (URL-decode) before storing/comparing, or the once-ever guarantee quietly breaks.
4. **Recommendations: 0 cards.** Awaiting owner confirmation whether the profile actually has
   recommendations received (asked separately). If it does, that locator chain needs a re-ground;
   if not, the empty read is correct. Either way production sends nothing, so this doesn't block
   the merge — it blocks the flip.

Also noted: `appreciation_touches` doesn't exist in prod yet (migration ships with this PR) —
`has_appreciation_touch` degraded gracefully to `False` as designed.

Owner already replied **`Ok`** (all recommendations), so proceed per the recommended options with
the above as live evidence; items 2–3 are worth fixing on this branch while it's open.

gitchrisqueen:
Addressed your review — `Ok` applied as **1A 2A 3A**, plus the two fixes the live grounding run turned up.

**1A — mentions stays the collaboration source.** No code change needed; `get_recent_collaborators` already reads `/notifications/?filter=mentions` and skips any card that doesn't SAY it was a mention.

**2A — reworded the stock `collaboration` template** (`db.py::_DM_DEFAULT_TEMPLATES`):
> Hi {first_name}, thanks for the mention — genuinely appreciated. What are you working on at the moment?

It's the code **default** only — `get_dm_template` reads `dm_templates` first, so anyone who customised theirs keeps it.

**3A — unchanged and still the plan.** `APPRECIATION_SOURCES_ENABLED` stays `false`; the migration lands inert and the flip in `/opt/lem/.env` is yours.

**Grounding item 3 — percent-encoded profile URL (the one that mattered).** You're right that this breaks the once-ever guarantee: `appreciation_touches`, `has_appreciation_touch` and the DM history all key on `profile_url`, so `/in/jane%2Ddoe%2D42` and `/in/jane-doe-42` were two different people. `_normalize_profile_url` now percent-decodes the path as well as stripping query/fragment/trailing slash. That's the **shared** normalizer, so the catch-up and roster ledgers pick the fix up too — worth knowing that any row already stored in an encoded spelling now keys differently, i.e. at most one extra touch for those people, once. New table, so `appreciation_touches` is unaffected.

**Grounding item 2 — empty actor name.** The name locator missed while the URL locator hit, so `_mention_actor_name` recovers it from the card's own sentence (`"Utkarsh Tiwari mentioned you in a comment in …"`). It's bounded to the ≤5 punctuation-free words immediately before the verb, so the notification chrome (`"Unread notification."`) can never be read as a name. When nothing name-like survives, the DM still opens with **"Hi there"** — now a deliberate, documented fallback rather than an accident: a generically addressed thank-you beats not thanking someone who publicly featured you. The probe applies the same fallback, so a blank `name` in a future report means production really would say "there", not that the probe read the card differently.

**Grounding item 4 — recommendations, 0 cards: NOT fixed, deliberately.** Nothing to change until you confirm whether that profile actually has recommendations received. If it does, the card locator chain needs a re-ground and I'd rather do that against a real card than guess at a chain. It doesn't block the merge (production sends nothing either way) — it blocks the flip, so worth a second `--appreciation-sources` run after this lands.

Tests: `test_actor_name_falls_back_to_the_card_sentence`, `test_notification_chrome_is_never_read_as_a_name`, `test_link_text_still_wins_over_the_sentence`, `test_percent_encoded_slug_is_one_person_not_two`, `test_collaboration_default_thanks_them_for_the_mention`, plus a probe test asserting the row reports the name production would use. Full unit lane green locally: **10056 passed**. `docs/engagement-automation.md` updated with all three.

gitchrisqueen:

- **Appreciation DMs bypassed the per-day DM cap, and the new sources are what made that matter** (`run_automation.py`): `send_private_dm` enforces no cap — unlike `send_scheduled_dm` and the outreach funnel, which both gate on `remaining_actions(ACTION_DM, max_dms_per_day, count_dms_sent_today, caps=…)`. That was survivable while the only trigger was `connection_accepted` (bounded by how many invitations you accepted). Both new sources read STANDING lists, so the first pass after the flag is flipped hands the dispatcher up to a month of un-thanked people and sends every one of them in one burst. The ledger stopped one person being thanked twice; nothing stopped thirty people being thanked in ninety seconds. **Fixed:** `_appreciation_dm_budget` computes the same shared allowance + #626 account envelope ONCE per pass and threads it through all three triggers, so they can't each spend the cap. What the budget can't afford is left **unclaimed** (so a later pass thanks them, not nobody), and a spent budget skips both scrapes rather than loading two pages for a list it can't act on. Tests: `TestDailyDmBudget`, `TestTheBeatSpendsOneBudget`.
- **A number in the card's quoted post could be read as the mention's age** (`_RELATIVE_AGE_RE`): the age matched on a bare `\b`, which `$5m ARR` clears exactly like `2h` does — and since the parse takes the first match, a card whose quoted comment mentioned money or a version could make a two-year-old mention look like it was posted minutes ago, and get the person thanked for it. **Fixed:** the age now has to be a standalone token (start-of-text or whitespace/bullet/bracket before the digits). Prose that still parses ("10 years of experience") can only push the age OUT of the window, which skips — the safe direction on a surface that DMs real people. Tests: `test_a_number_inside_the_quoted_post_is_not_the_timestamp`, `test_a_glued_timestamp_still_parses`.
- **The close left scope behind** (`Closes #968`): the issue says *"Needs a live-grounding probe of the Recommendations Received section before the scraper ships"*, and the grounding run on this PR returned **0 cards** for exactly that section, with the flip explicitly deferred. That remainder existed only in prose inside an issue this PR closes — the #548 failure mode. **Fixed:** filed **#1002** (*Phase 2 — ground the Recommendations Received chain and flip `APPRECIATION_SOURCES_ENABLED`*) with the probe command, real acceptance criteria and a Decision Comment; linked in the PR body and commented on #968. `Closes #968` kept.

Checked and found nothing to change: the ledger claim ordering (checked before the LLM call, claimed before the send, unreadable ledger = don't send) is right and `connection_accepted` correctly inherits it; the migration is additive with a timestamp version and its unique key fits InnoDB's prefix limit; `_normalize_profile_url`'s percent-decode is the correct fix for the grounded `%2D` slug and its one-extra-touch effect on the catch-up/roster ledgers is already called out; undated → `None` → skip is applied consistently and the all-undated warning is a real drift signal, not a no-op warning; the probe is genuinely read-only and drives the production locator chains rather than a copy.

Full unit lane green locally: **10069 passed**.

gitchrisqueen:

gitchrisqueen:

gitchrisqueen:

gitchrisqueen:
Post-merge follow-up on the owner's **2B** answer (recommendations DO exist): a second read-only grounding pass confirmed the bare `/details/recommendations/` URL **does** default to the Received tab and the two real recommendations render there — but every `_RECOMMENDATION_TAB_LOCATORS` / `_RECOMMENDATION_CARD_LOCATORS` rung is unmatchable on the live SDUI render (no `li`, no `time`, no `role=tab`, no `data-view-name` on the page; the one `div[role=list]` is the footer). The mentions-side fixes from the grounding run (sentence-name, percent-decode) are confirmed merged. Filed **#1007** with the full DOM evidence — it's the flip-blocker for the recommendations half of `APPRECIATION_SOURCES_ENABLED`; note `?detailScreenTabIndex=2` turned out to be the **Pending-requests** tab, correctly never thank-worthy.

PR #997, opened 2026-08-03
fix(ai): ride out a LiteLLM proxy that is not accepting connections (closes #986)

### LINKED ISSUE #986: fix(errors): APIConnectionError: Connection error. (1x)
## Why
PostHog error tracking grouped this exception into an issue that is still active: **APIConnectionError**

> Connection error.

- First seen: `2026-08-02T09:01:04.964000Z` · last seen: `2026-08-02T09:01:04.964000Z`
- SDK: `posthog-python`

[Open the issue in PostHog](https://us.posthog.com/project/475262/error_tracking/019fc1b4-951f-7541-b002-ebff9ad8c9be) — it has the stack trace, the grouped occurrences and the affected people.

## Scope

## Files
- `tests/unit/`

## Acceptance
- All required CI gates pass.

Auto-filed from PostHog Error Tracking. Dedup marker (do not remove): `posthog-issue-019fc1b4-951f-7541-b002-ebff9ad8c9be`

gitchrisqueen:
## What & why

PostHog grouped an `APIConnectionError: Connection error.` on `auto_comment_in_groups`. The stack trace's only in-app frame is `ai_helper._call_llm:133`, and the chained causes are `httpx.ConnectError` / `httpcore.ConnectError` — **`[Errno 111] Connection refused`** talking to the LiteLLM proxy.

That proxy is a container LEM restarts on its own schedule (deploys, image pulls, the host's nightly unattended-upgrade reboot). While it comes back up every call is refused, and the OpenAI SDK spends its own two retries inside ~1.5s — far short of a container start. So one blip lost the generation in flight, aborted the run, and filed a defect for working infrastructure.

This is not an expected no-op (a lost generation is real degradation), so per the escalation contract the error stays — the app just stops being unable to survive a restart.

## The change

`AttributedOpenAI` already exists as the ONE place every endpoint funnels through, so the wait goes there:

- **`post()` retries a connection that was NEVER ESTABLISHED.** Chat, embeddings, images and speech are all POSTs, so this covers every LLM surface — including the ~12 call sites that bypass `_call_llm` entirely (embeddings for comment dedup / feedback clustering, `image_gen`, `tools.py`, `newsletter_cover`, …).
- **Wrapped at `post()` rather than `request()`** because the SDK's own retry re-enters `request()` — nesting the two budgets would multiply them (3 × 3 × 3 attempts and nested sleeps).
- **Only `APIConnectionError` whose `__cause__` is `httpx.ConnectError`.** Nothing was sent, so there is no provider spend to duplicate. A timeout (which may already have been served and billed), a 4xx, or a 5xx is the proxy *answering* and fails exactly as before.
- **Bounded**: `LLM_CONNECT_RETRY_ATTEMPTS` (default 3) × `LLM_CONNECT_RETRY_BACKOFF_SECONDS` (default 8, exponential) ≈ 24s ridden out; `1` turns the wait off. Read at call time, unparseable values fall back to the default. Documented in `.env.example`.
- **The per-retry log is DEBUG, not a warning.** A proxy restart happens on every deploy — warning on it would escalate and file a defect for expected behaviour (the #917 `pause_automation` lesson). An exhausted budget still raises, and `_call_llm` still logs *that* at ERROR, so a real outage still reaches error tracking.

## Testing

New `tests/unit/utilities/ai/test_client_connect_retry.py` (9 tests) drives **real requests through the SDK** over a transport that refuses connections the way a starting container does — same approach as `test_client_attribution.py`, so an SDK upgrade that moves the hook fails CI instead of silently restoring the old behaviour. It pins: a refused connection is ridden out; the backoff schedule is `8s, 16s`; a proxy that stays down still fails inside the budget; `attempts=1` disables it; an unparseable setting falls back to the default; the retry logs DEBUG once; a **timeout** and a **500** are each attempted exactly once; embeddings are covered too; and the shared singleton carries the guard.

```
poetry run pytest tests/unit -q   →  10018 passed
```

Dedup marker (do not remove): `posthog-issue-019fc1b4-951f-7541-b002-ebff9ad8c9be`

Closes #986

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -98,6 +98,10 @@ LITELLM_MASTER_KEY=your_litellm_master_key
 LITELLM_BASE_URL=http://litellm:4000
 # Uncomment to use LiteLLM outside of Docker
 # LITELLM_BASE_URL=http://localhost:4000
+# How long an LLM call waits out a proxy that is not accepting connections (deploy / restart).
+# Attempts include the first try, so 3 x 8s = ~24s of refused connections ridden out; 1 disables it.
+# LLM_CONNECT_RETRY_ATTEMPTS=3
+# LLM_CONNECT_RETRY_BACKOFF_SECONDS=8
 
 # --- OpenAI ---
 OPENAI_API_KEY=your_openai_api_key
--- a/.env.example
+++ b/.env.example
@@ -99,7 +99,9 @@ LITELLM_BASE_URL=http://litellm:4000
 # Uncomment to use LiteLLM outside of Docker
 # LITELLM_BASE_URL=http://localhost:4000
 # How long an LLM call waits out a proxy that is not accepting connections (deploy / restart).
-# Attempts include the first try, so 3 x 8s = ~24s of refused connections ridden out; 1 disables it.
+# Attempts include the first try and each wait DOUBLES (capped at 60s), so the default rides out
+# 8s + 16s = ~24s of refused connections; 1 disables the wait. Raising attempts grows the total
+# faster than linearly — 5 attempts is 8+16+32+60 = ~116s, not ~40s.
 # LLM_CONNECT_RETRY_ATTEMPTS=3
 # LLM_CONNECT_RETRY_BACKOFF_SECONDS=8
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -78,6 +78,14 @@ All LLM calls go through LiteLLM proxy via `utilities/ai/client.py`:
 response = client.chat.completions.create(model="lem-simple", messages=[...])
 ```
 
+`AttributedOpenAI` is the ONE client — it stamps attribution + trace ids on every endpoint, and it
+**rides out a proxy that is not accepting connections** (#986): the proxy is a container LEM restarts
+on deploys/reboots, and the SDK's own retries are spent in ~1.5s, so one blip used to lose the
+generation and file a defect for it. ONLY a connection that was never established is retried
+(`LLM_CONNECT_RETRY_ATTEMPTS` / `LLM_CONNECT_RETRY_BACKOFF_SECONDS`, ~24s by default) — nothing was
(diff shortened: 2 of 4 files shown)


gitchrisqueen:

- **Unbounded exponential wait (`client.post`)**: the backoff doubled with no ceiling while `LLM_CONNECT_RETRY_ATTEMPTS` is operator-tunable, so `=10` would sleep >1h inside a Celery task that sets no `time_limit`, holding the worker slot — against the module's own "bounded" claim. Added `_MAX_CONNECT_RETRY_DELAY = 60.0` and clamped the per-retry delay; the default schedule (8s, 16s) is byte-identical, and a new test pins `[8, 16, 32, 60, 60]` at `attempts=6`.
- **`.env.example` mis-sized the budget**: `3 x 8s = ~24s` is right only by coincidence at the default (real schedule is 8+16). An operator setting 5 attempts would expect ~40s and get ~116s. Rewrote it to state the doubling and the cap.

Verified and found sound (no change needed): `post()` really is the SDK funnel for chat/embeddings/images/speech and `super()` resolves to `SyncAPIClient.post`; `APIConnectionError.__cause__` survives the SDK's own internal retries (`raise ... from err` in `_request`), so the guard still fires with production `max_retries=2`; `APITimeoutError` subclasses `APIConnectionError` but its cause is `httpx.TimeoutException`, so timeouts are correctly excluded; wrapping `post()` rather than `request()` does avoid nesting the two budgets; `_attach_attribution`/`_attach_trace` are idempotent across re-posts (no metadata accumulation on the caller's body); no `async def` route calls an LLM helper, so the added `time.sleep` never blocks the FastAPI event loop; DEBUG (not WARNING) is the right level per the #917 escalation contract. Close is honest — #986 has no later phase and both acceptance criteria are met.

`poetry run pytest tests/unit -q` → **10019 passed**.

gitchrisqueen:

gitchrisqueen:

gitchrisqueen:

PR #1008, opened 2026-08-03
chore(ci): catch CLAUDE.md cap drift at the source (closes #1000)

### LINKED ISSUE #1000: chore(ci): catch CLAUDE.md cap drift at the source — the guard's paths filter hides it until an unrelated PR pays for it
## Why

`.github/workflows/claude-md-size.yml` runs **only** on PRs whose diff touches `CLAUDE.md`
(`paths:` filter). Nothing checks the cap on a push to `main`. So `main` can drift over 40,000
chars silently, and the bill lands on the next unrelated PR that happens to edit the file.

This has now happened twice:

- **#885** (closed): `main` was 41,763 chars; PR #884 inherited the red check and did the one-pass trim.
- **This time:** `main` reached **41,396 chars** on its own. PR #980 (#966, permalink commenting)
  added ~460 chars of SDUI invariant and its `CLAUDE.md ≤ 40,000 chars` check went red for a
  condition it mostly did not cause. #980 had to carry ~1,700 chars of trimming across bullets it
  never touched to go green.

#885's remedy was a one-time trim. The recurrence is the evidence that a trim alone doesn't hold —
CLAUDE.md grows on nearly every agent PR, so the cap gets re-crossed and the cost keeps landing on
whoever edits next rather than on whoever crossed it.

## Scope

- Run the guard where the drift happens: add `push: branches: [main]` to the workflow (keep the
  `paths:` filter for the PR trigger if the perf saving matters — a `main` push has no such filter
  to worry about).
- Make the overage visible before it is inherited. Options, pick one: fail the `main` run (loud,
  but a red `main` badge for a docs condition), or emit a warning + auto-file an issue at, say,
  38,000 chars so there is runway to trim deliberately. Prefer the early-warning shape — the cap
  is a harness limit, not a correctness bug.
- Settle the required-check question #885 left half-open: the workflow comment calls itself a
  required gate; confirm whether `CLAUDE.md ≤ 40,000 chars` is in branch protection's required
  contexts and make the comment match reality either way.
- Bring `main` back under the cap if it has drifted again by the time this is picked up.

Not in scope: another content trim pass. The point of this issue is that the *detection* is in the
wrong place.

## Files

- `.github/workflows/claude-md-size.yml`
- `scripts/check_claude_md_size.py` (only if a warn threshold is added)

## Acceptance

- [ ] A push to `main` that leaves `CLAUDE.md` over the cap is reported by CI, not discovered by
      the next PR that edits the file.
- [ ] A PR that edits `CLAUDE.md` while `main` is already over the cap can tell from the check
      output that it inherited the overage rather than caused it.
- [ ] The workflow's required-check claim matches branch protection.
- [ ] `python3 scripts/check_claude_md_size.py` exits 0 on `main`.

Follow-up of #885 (that issue trimmed the file; this one stops the drift being invisible).
Surfaced while landing #980 / #966.

gitchrisqueen:
## Why

`.github/workflows/claude-md-size.yml` only ran on PRs whose diff touched `CLAUDE.md`, so `main`
could drift over the 40,000-char cap on its own (a squash/rebase merge with no matching diff) and
the cost landed on whoever edited the file next (#885, then again just before this PR — `main` was
at 42,980 chars).

## What changed

- **`push: branches: [main]` trigger** (new `drift` job, no `paths:` filter on purpose — this is the
  one place that has to catch drift regardless of what a merge's diff looked like). It runs
  `check_claude_md_size.py --warn-at` (soft mode, default threshold 38,000 chars) and — only when the
  size crosses that threshold — files or updates a tracking issue (`gh issue list` client-side
  filtered by title prefix, then `create`/`comment`; the drift job **never fails the build**, since a
  docs-cap regression on `main` shouldn't redden the branch — that's the early-warning shape the issue
  asked to prefer over a hard fail).
- **PR job now takes `--baseline-ref`**: it fetches the PR's base branch and prints a note saying
  whether an over-cap PR *inherited* an already-over-cap `main` or *pushed* it over itself, so the
  check output distinguishes cause from inheritance.
- **Required-check comment confirmed against reality** (2026-08-03): `CLAUDE.md ≤ 40,000 chars` is
  still NOT in branch protection's required status checks — the existing comment was already
  accurate, dated it.
- **`CLAUDE.md` trimmed back under the cap** (42,980 → 38,602 chars): tightened wording across the
  two largest sections (Engagement automation, Anti-bot/session infra) plus a few others, moving
  nothing out of the file — the detail already lives in the linked `docs/*.md` files this trims
  toward.
- `CONTRIBUTING.md` updated to describe the three-layer guard (PR check / push drift watch / script).

`scripts/check_claude_md_size.py`'s default no-args invocation is unchanged (still prints size, exits
1 over cap) — the new `--warn-at` and `--baseline-ref` flags are additive.

## Testing

- `poetry run pytest tests/unit -q` — 10,183 passed (new `tests/unit/test_check_claude_md_size.py`,
  17 tests, covers strict/soft modes, baseline inherited-vs-caused logic, `$GITHUB_OUTPUT` writing,
  and the CLI).
- `poetry run ruff check tests/unit/test_check_claude_md_size.py` — clean (`scripts/` itself isn't
  ruff-gated in CI, matching the existing `scripts/model_health_check.py`).
- `python3 scripts/check_claude_md_size.py` — `ok: CLAUDE.md is 38,602 / 40,000 chars` (was failing
  at 42,980 before this PR).
- Manually exercised every script code path (strict/soft/warn-zone/over-cap/`$GITHUB_OUTPUT`/missing
  ref) and validated the new workflow YAML parses.

Closes #1000


gitchrisqueen:

- **CLAUDE.md trim silently dropped content, not just prose** (not a wording-economy pass): while condensing 42,980→38,602 chars, two spots lost actual invariants rather than just tightening phrasing — the identity-and-sessions bullet dropped the explicit gotcha "importing `db.get_session_user_id` around the wrapper is a bug", and the strong-auth bullet dropped that the second-factor attempt count is durable in `auth_challenges.attempts` / `SECOND_FACTOR_ATTEMPT_WINDOW_MINUTES` and carries into the next handle (so restarting the flow doesn't reset the budget). Both are exactly the "symbols, constants, invariants" CLAUDE.md's own header commits to keeping, and the PR body's claim that the trim moved "nothing out of the file" wasn't quite true for these two. Restored both (`4af46f69`); file is now 38,815 / 40,000 chars, still comfortably under cap.

Everything else checked out:
- `--warn-at`/`--baseline-ref` logic, `$GITHUB_OUTPUT` writing, and the drift-issue dedup query all match their tests; workflow YAML is valid; job-level `permissions` on the new `drift` job (contents:read, issues:write) are correct for `gh issue create`/`comment`.
- Verified against live branch protection: `CLAUDE.md ≤ 40,000 chars` is confirmed NOT in `required_status_checks.contexts` — the updated CONTRIBUTING.md/workflow comment is accurate.
- `python3 scripts/check_claude_md_size.py` exits 0 on this branch (38,815/40,000) — acceptance criterion 4 holds.
- All 4 acceptance criteria on #1000 are met; `Closes #1000` is honest, no untracked follow-up scope.

gitchrisqueen:
🧩 phase-guard — heads up: issue #1000 still has **boxes: 4 unchecked** acceptance box(es). Merging anyway (unticked boxes are common here), but if any of that scope did NOT ship, file a follow-up issue with `agent:ready` and link it, or drop the `Closes #1000` from this PR.

gitchrisqueen:

gitchrisqueen:

gitchrisqueen:

Issue #36, opened 2026-06-17
[M4] Build Schedule Content page


gitchrisqueen:
**New file:** `src/cqc_lem/ui/src/pages/ScheduleContent.tsx`
**Replaces:** `src/cqc_lem/streamlit/pages/2_Schedule_Content.py`

## Features
- Textarea with live character counter (max 3000, LinkedIn limit)
- Post type selector: TEXT / IMAGE / VIDEO / CAROUSEL
- Drag-and-drop media upload component
- Date + time picker with timezone auto-detection
- AI best-time suggestion button (`GET /ai/best_time?day_of_week=...`)
- Submit calls `POST /schedule_post/`
- Live LinkedIn post preview (see Issue #37) updates as content is typed

gitchrisqueen:
Resolved in PR #85: ScheduleContent.tsx is complete with live LinkedInPostPreview as you type.

PR #1011, opened 2026-08-03
feat(analytics): 12-hour time + timezone label in Best Times to Post (closes #1003)

### LINKED ISSUE #1003: feat(analytics): Use 12-hour time format and user's timezone in Best Times to Post
## Why
Best Times to Post currently shows times in 24h format. User requests 12h format and timezone based on user preference or browser detection.

Auto-filed from in-app feedback #36 by the feedback→auto-work loop (`docs/launch-and-marketing-plan.md` §B.2/B.3). Classifier: enhancement/low, risk `none`, confidence 0.95.

[Watch the session replay](https://us.posthog.com/project/475262/replay/019fc580-376e-702c-9388-18d2a89eab6a) — the browser session this report was filed from.

## Scope
- Address the reported enhancement in the `analytics` area of the product.

## Files
- `tests/unit/`

## Acceptance
- All required CI gates pass.

gitchrisqueen:
## What & why

In-app feedback (#36 → issue #1003): **Your Best Times to Post** showed a bare 24-hour clock face —
`Wednesday @ 16:00` — with nothing saying *which* zone that hour was on.

The hours themselves were already right. `/user/post-stats` calls
`recommend_post_times(rows, tz=get_user_timezone(user_id))`, so each `hour` is already the user's own
wall clock, and the response already carries the `timezone` it bucketed into. The SPA rendered
neither fact: it padded the raw integer and dropped the zone. This was also the one display left in
the SPA breaking `docs/timezone-contract.md` §4 ("times are always shown 12-hour").

## Changes

- **`utils/datetime.ts`** — two shared helpers for a bare **hour-of-day** (which has no instant to
  hand `Intl`, so it can't go through `formatInTimezone`):
  - `formatHour12(hour)` → `0 → "12:00 AM"`, `13 → "1:00 PM"`; out-of-range hours normalize rather
    than render nonsense.
  - `timezoneAbbreviation(tz, at?)` → `"EDT"` / `"GMT+5:30"`, DST-correct at the given instant,
    falling back to the IANA id when the engine has no short name.
- **`Dashboard.tsx`** — renders `Wednesday @ 4:00 PM` and labels the list once with the zone the
  **server** bucketed into (`postStats.timezone`). `useUserTimezone()` (configured Login Location,
  browser detection as its own last resort) stands in only for an older payload without the field —
  labelling a server-bucketed hour with the browser's guess would mislabel a traveling user.
- Folded the two hand-rolled AM/PM copies onto the shared helper (`NewsletterCard`, `ComposePost`).
  `ComposePost`'s copy would have rendered hour 0 as `"0:00 AM"` — latent, since it only ever picks
  7/8/10, but it's gone now rather than left as a third near-copy.
- **`docs/timezone-contract.md`** — records the bare-hour helpers beside the three conversion
  helpers, so the next hour-of-day surface doesn't hand-roll a fourth copy.

No backend change: the API already returned everything needed. (The issue's "Files" guess at
`observability.py` was wrong — the endpoint is `api/main.get_post_stats_endpoint`.)

## Testing

`src/cqc_lem/ui/src/utils/datetime.test.ts` — every hour 0–23 asserted against a 12-hour-only
pattern, the named boundaries (0/11/12/13/23), out-of-range normalization, the DST-correct zone
label across a boundary (`EDT` in July vs `EST` in January), and the bad-zone fallback.

Run locally under Node 22 (matching CI):
- `npx vitest run` → **424 passed / 51 files**
- `npx tsc -b` → clean
- `npm run build` → clean

Python patch coverage is unaffected — `codecov.yml` ignores `src/cqc_lem/ui/`; the SPA is gated by
`CI / UI Build` (vitest + build).

Closes #1003

### DIFF
--- a/docs/timezone-contract.md
+++ b/docs/timezone-contract.md
@@ -89,6 +89,14 @@ from logs rather than only from a post that went out at the wrong hour.
 | `toZonedInputValue(iso, tz)` | UTC ISO → `<input type="datetime-local">` value in `tz` |
 | `zonedInputToUtcIso(value, tz)` | `datetime-local` value read as `tz` → UTC ISO with `Z` |
 
+Some surfaces render a bare **hour-of-day** with no date attached — the Dashboard's "Your Best Times
+to Post", the newsletter publish-hour picker, the compose-page suggestion. There is no instant to
+hand `Intl`, so those get the other two helpers instead: `formatHour12(hour)` (12-hour, never a
+`14:00` clock face) and `timezoneAbbreviation(tz)` (the `EDT`/`GMT+5:30` label that says WHICH clock
+the hour is on). Never hand-roll either — three near-copies of the AM/PM arithmetic is how the Best
+Times list ended up 24-hour and unlabelled (issue #1003). A bucketed hour is labelled with the zone
+the SERVER bucketed it into (`/user/post-stats` returns `timezone`), not the browser guess.
+
 A `datetime-local` input carries **no timezone**. Its value must be the same wall clock
 `formatInTimezone` renders beside it, or the editor and its own label disagree about one post. On
 the way back out, `new Date(value).toISOString()` is always wrong — it interprets the value in the
--- a/src/cqc_lem/ui/src/pages/Dashboard.timezone.test.tsx
+++ b/src/cqc_lem/ui/src/pages/Dashboard.timezone.test.tsx
@@ -1,0 +1,111 @@
+import { describe, expect, it, vi, afterEach, beforeEach } from 'vitest'
+import type { ReactNode } from 'react'
+import { cleanup, render, screen, waitFor } from '@testing-library/react'
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
+import { MemoryRouter } from 'react-router-dom'
+import Dashboard from './Dashboard'
+
+// Issue #1003: "Your Best Times to Post" rendered a bare 24-hour clock face with no zone, so a
+// recommendation read as neither the user's clock nor anyone else's. The helper arithmetic is
+// covered in utils/datetime.test.ts; what is asserted HERE is the wiring the helpers hang off —
+// that the list renders 12-hour, and that it is labelled with the zone the SERVER bucketed the
+// hours into rather than the browser's guess.
+const get = vi.fn()
+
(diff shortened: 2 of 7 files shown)


gitchrisqueen:

Reviewed the full diff plus the seams it depends on: `/user/post-stats` (`api/main.get_post_stats_endpoint`) does return `timezone`, `db.get_user_timezone` always yields a validated non-empty IANA id (default `America/New_York`), and `post_stats.recommend_post_times(tz=…)` really does bucket into that zone's wall clock — so the new label states a true fact rather than relabelling UTC hours. `formatHour12` matches the two hand-rolled copies it replaces for every hour they could produce (and fixes ComposePost's latent `0:00 AM`), and a grep confirms no fourth copy was left behind. `Closes #1003` is honest — every acceptance criterion is met and no later phase is left untracked.

- **Test gap on the changed behavior**: the helper arithmetic was well covered in `utils/datetime.test.ts`, but the wiring the issue was about had none — nothing failed if `formatHour12(r.hour)` reverted to `String(r.hour).padStart(2, '0')`, and nothing pinned the label to `postStats.timezone` instead of the browser guess (the whole point of preferring the server-bucketed zone). Added `src/cqc_lem/ui/src/pages/Dashboard.timezone.test.tsx`, following the repo's existing `*.timezone.test.tsx` pattern: asserts the list renders 12-hour and never a 24-hour face, that the label follows the SERVER zone (`America/New_York` while the hook says `Asia/Tokyo`), that it falls back to the hook only when the payload carries no `timezone`, and that the empty-recommendations branch claims no zone at all.

Verified under Node 22 (matching CI): 52 files / 428 vitest tests pass, `npm run build` clean, eslint clean on the new file.

Not findings, recorded so they aren't re-litigated: `formatHour12(NaN)` renders `NaN:00 PM`, but no call site can produce a non-finite hour (API int, `Number()` off a 0–23 `<select>`, hardcoded 7/8/10); and `timezoneAbbreviation` falling back to the IANA id would render `Times are in X (X)`, reachable only via an invalid `users.timezone`, which `PUT /user/timezone` validates against `available_timezones()`.

gitchrisqueen:

gitchrisqueen:

gitchrisqueen:

Issue #37, opened 2026-06-17
[M4] Build Post Review & Management page


gitchrisqueen:
**New file:** `src/cqc_lem/ui/src/pages/ReviewSchedule.tsx`
**Replaces:** `src/cqc_lem/streamlit/pages/3_Review_Schedule.py`

## Features
- Responsive data table (TanStack Table) with columns: date, type, status, content preview
- Row click expands inline editor: content textarea, status dropdown, scheduled_time picker
- Save calls `POST /update_post/`
- Status filter tabs: All / Pending / Approved / Scheduled / Posted
- 'Generate Weekly Content' button calls `POST /create_weekly_content/`
- LinkedIn post preview panel (Issue #37) shown alongside editor

gitchrisqueen:
Resolved in PR #85: ReviewSchedule.tsx is complete with status tabs, inline editor, and LinkedInPostPreview panel.

PR #1016, opened 2026-08-03
fix(analytics): format avg engagement as a percentage in Best Times to Post (closes #1004)

### LINKED ISSUE #1004: feat(analytics): Format avg engagement as a percentage in Best Times to Post card
## Why
The "avg engagement" in the "Your Best Times to Post" card displays as a decimal number. User suggests converting this to a proper percentage with set precision.

Auto-filed from in-app feedback #37 by the feedback→auto-work loop (`docs/launch-and-marketing-plan.md` §B.2/B.3). Classifier: enhancement/low, risk `none`, confidence 1.00.

[Watch the session replay](https://us.posthog.com/project/475262/replay/019fc580-376e-702c-9388-18d2a89eab6a) — the browser session this report was filed from.

## Scope
- Address the reported enhancement in the `analytics` area of the product.

## Files
- `tests/unit/`

## Acceptance
- All required CI gates pass.

gitchrisqueen:
## What & why

The "Your Best Times to Post" card printed `avg_engagement` raw, so a recommendation read as
`avg engagement 0.0612` — a bare decimal with no unit (in-app feedback #37).

That number does **not** have a fixed scale. `post_stats._group_metrics` scores in **rate mode** —
a per-impression fraction — when every row in the comparison set carries impressions, and in
**count mode** — a weighted per-post count — otherwise, and it already stamps which one on the
payload as `metric` (`engagement_rate` / `engagement`). The card ignored it.

So the fix is to read `metric`, not to blanket-multiply by 100:

- **rate** → `avg engagement rate 6.12%` (percentage, fixed 2 decimals)
- **count** → `avg engagement 12.5` (a count is not a percentage — `40` would have rendered as
  `4000%`)
- **no `metric`** (a payload predating the field) → treated as a count, never as a percentage

`formatEngagementMetric` is added beside `formatRate` in `components/charts/palette.ts` — the same
helper the format/hook leaderboards already use for the identical two-scale problem — so the card
and the leaderboards can't drift apart on precision. Non-finite input degrades to `—` rather than
printing `NaN%`.

No backend change: `recommend_post_times` already returns `metric` in both modes (covered by
`tests/unit/utilities/test_post_stats.py`).

## Testing

- `src/components/charts/palette.test.ts` — new: the percentage applies to the fraction only, a
  count keeps 1-decimal precision, an absent/unknown metric falls back to a count, non-finite → `—`.
- `src/pages/Dashboard.bestTimes.test.tsx` — new: the card wiring, one case per metric mode plus
  the metric-less payload.
- Full SPA suite green locally: `npm test` → 54 files / 436 tests passed; `npm run build` clean.

Closes #1004

### DIFF
--- a/src/cqc_lem/ui/src/components/charts/palette.test.ts
+++ b/src/cqc_lem/ui/src/components/charts/palette.test.ts
@@ -1,0 +1,39 @@
+import { describe, expect, it } from 'vitest'
+import { ENGAGEMENT_RATE_METRIC, formatEngagementMetric, formatRate } from './palette'
+
+// Issue #1004: "avg engagement" rendered the raw number, so an impression-normalized rate read as
+// a bare decimal ("avg engagement 0.0612"). The scale is NOT fixed — post_stats emits a fraction
+// in rate mode and a weighted count otherwise — so what is asserted here is that the percentage
+// is applied to the fraction ONLY.
+describe('formatEngagementMetric', () => {
+  it('renders a rate as a percentage at fixed precision', () => {
+    expect(formatEngagementMetric(0.0612, ENGAGEMENT_RATE_METRIC)).toBe('6.12%')
+    expect(formatEngagementMetric(0.06, ENGAGEMENT_RATE_METRIC)).toBe('6.00%')
+    expect(formatEngagementMetric(0, ENGAGEMENT_RATE_METRIC)).toBe('0.00%')
+    expect(formatEngagementMetric(0.000004, ENGAGEMENT_RATE_METRIC)).toBe('0.00%')
+  })
+
+  it('never percent-ifies a per-post count', () => {
+    expect(formatEngagementMetric(40, 'engagement')).toBe('40')
+    expect(formatEngagementMetric(12.5, 'engagement')).toBe('12.5')
+    // A count is rounded to 1 decimal server-side; hold that precision if one ever slips through.
+    expect(formatEngagementMetric(12.34567, 'engagement')).toBe('12.3')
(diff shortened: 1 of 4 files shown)


gitchrisqueen:
🔎 Claude adversarial review — PASS

- Scale: verified `engagement_rate()` is `engagement_score / impressions` (a fraction), so `formatRate`'s ×100 is correct, and `METRIC_RATE = "engagement_rate"` matches the SPA constant exactly.
- Contract: `_group_metrics` stamps `metric`, `recommend_post_times` pops only `samples` before spreading, and `/user/post-stats` passes `detail` through — the field really does reach the card. Both modes are pinned server-side (`test_post_stats.py:105`/`:117`), so the card's new dependency can't silently regress.
- Count path: confirmed a weighted per-post count is never percent-ified (the 40 → "4000%" trap), and an absent/unknown `metric` falls back to a count, so a payload predating the field degrades safely rather than inventing a percentage.
- Drift: `Leaderboard.tsx` is the only other `avg_engagement` consumer and already reads `metric` + `formatRate`, so rate precision is genuinely shared between the two surfaces; no other SPA or Python surface renders this value.
- Close honesty: #1004 has no phased scope and no unchecked remainder — all three acceptance criteria are met, so `Closes #1004` is honest.
- Tests: `UI Build` runs `npm test`, so both new vitest files actually execute in CI (green); codecov patch/project pass.

No findings — no code changes pushed.

gitchrisqueen:

gitchrisqueen:

PR #1018, opened 2026-08-03
fix(celery): fill task defaults into the celery-once dedup key (closes #989)

### LINKED ISSUE #989: fix(errors): KeyError: 'sweep_slot' (1x)
## Why
PostHog error tracking grouped this exception into an issue that is still active: **KeyError**

> 'sweep_slot'

- First seen: `2026-08-03T01:21:05.872000Z` · last seen: `2026-08-03T01:21:05.872000Z`
- API route: `/api/linkedin/verification-pin/inbound`
- SDK: `posthog-python`

[Open the issue in PostHog](https://us.posthog.com/project/475262/error_tracking/019fc535-e47e-7a02-b70c-14e284957e6a) — it has the stack trace, the grouped occurrences and the affected people.

## Scope

## Files
- `tests/unit/`

## Acceptance
- All required CI gates pass.

Auto-filed from PostHog Error Tracking. Dedup marker (do not remove): `posthog-issue-019fc535-e47e-7a02-b70c-14e284957e6a`

gitchrisqueen:
## What

`celery-once` builds its lock key from `Signature.bind(...).arguments` — which **never applies defaults** — and then indexes that mapping with every name in `once={'keys': [...]}`:

```python
restrict_kwargs = {key: kwargs[key] for key in restrict_to}   # celery_once/helpers.py
```

So enqueuing a task without explicitly passing a *defaulted* key parameter raises `KeyError: '<key>'` **inside `apply_async`**, before the message is ever sent — which surfaces as a 500 on the caller. That is the reported `KeyError: 'sweep_slot'` on `/api/linkedin/verification-pin/inbound` (the inbound reply-notification webhook, which routes reply+token mail into `_process_reply_inbound`).

`#951` patched the two call sites that were hitting it. This PR removes the class of bug: **28** tasks declare `once={'keys': [...]}`, and each one carried the same landmine at every present and future enqueue site.

`cqc_lem/app/queue_once.py` adds a `QueueOnce` subclass that fills a task's own parameter defaults for keys named in `once['keys']`, making an omitted `sweep_slot=0` and an explicit `sweep_slot=0` the **same** lock — which is what the key always meant. `run_automation.py` / `run_scheduler.py` import `QueueOnce` from there instead of from `celery_once`; no decorator site changed.

## Why it's safe

- **Only** keys named in `once['keys']` are filled. A task with no `keys` builds byte-for-byte the key celery-once always built, so no live lock changes identity across the deploy.
- A restricted key with no default is never invented (`Signature.bind` rejects a missing required argument first, so it cannot be absent).
- Distinct values still produce distinct locks — verified for `sweep_slot` (concurrent golden-hour sweeps must not collapse into one lock) and for `user_id`.
- The explicit `sweep_slot=0` at the existing call sites is left in place; it is still correct and still asserted by their tests.

## Testing

`tests/unit/app/test_queue_once_defaults.py` (6 cases): the exact enqueue shape that raised (`sweep_reply_comments.get_key(kwargs={"user_id": 7})`) no longer raises and equals the explicit-default key; distinct slots/users stay distinct; an unrestricted task keeps upstream's key shape; a restricted key without a default is left alone.

Confirmed the upstream base class still raises the exact reported error on the same shape, so the test is guarding a real regression:

```
$ python -c "... @app.task(base=celery_once.QueueOnce, once={'keys': ['user_id','sweep_slot']}) ..."
UPSTREAM KeyError: 'sweep_slot'
```

Full unit lane green locally: `10193 passed`.

Closes #989

### DIFF
--- a/src/cqc_lem/api/main.py
+++ b/src/cqc_lem/api/main.py
@@ -2383,8 +2383,9 @@ def _process_reply_inbound(form) -> ResponseModel:
     if not _reply_sweep_debounced(user_id):
         _log_inbound_verdict("debounced", form, user_id=user_id)
         return ResponseModel(status_code=200, detail="debounced")
-    # QueueOnce builds its dedup key from once['keys'] and does NOT apply function defaults — omitting
-    # sweep_slot here raised KeyError at enqueue and 500'd every forwarded notification.
+    # slot 0 is the single-shot trigger (the golden-hour amplifier owns the other slots). Omitting it
+    # once raised KeyError at enqueue and 500'd every forwarded notification; cqc_lem.app.queue_once
+    # now fills the default into the dedup key, so this stays explicit for meaning, not for safety.
     sweep_reply_comments.apply_async(kwargs={"user_id": user_id, "sweep_slot": 0}, countdown=120)
     _log_inbound_verdict("comment_accepted", form, user_id=user_id)
     log_info("Triggered reply sweep from comment notification", user_id=user_id)
--- a/src/cqc_lem/app/queue_once.py
+++ b/src/cqc_lem/app/queue_once.py
@@ -1,0 +1,30 @@
+from inspect import Parameter
+
+from celery_once import QueueOnce as _CeleryQueueOnce
+
+
+class QueueOnce(_CeleryQueueOnce):
+    """celery-once, but a task's OWN parameter defaults count toward its dedup key.
+
+    Upstream builds the key from `Signature.bind(...).arguments`, which never applies defaults, then
+    indexes that mapping with every name in `once={'keys': [...]}`. So enqueuing a task without
+    passing a defaulted key parameter raised `KeyError: '<key>'` inside `apply_async` — before the
+    message was ever sent — which surfaced as a 500 on the caller (issue #989: `sweep_slot` on the
+    inbound reply-notification webhook). Filling the default makes an omitted `sweep_slot=0` and an
+    explicit `sweep_slot=0` the SAME lock, which is what the key always meant.
+
+    Only keys named in `once['keys']` are filled, so the unrestricted case keeps upstream's key shape
+    byte for byte and no existing lock changes identity.
+    """
+
+    abstract = True
+
+    def _get_call_args(self, args, kwargs) -> dict:
+        call_args = super()._get_call_args(args, kwargs)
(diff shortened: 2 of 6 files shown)


gitchrisqueen:

Verified the core fix holds before hunting: audited all 43 `once={'keys': […]}` declarations against their signatures (AST) — every restricted key is a real parameter, and `sweep_slot` is the ONLY one with a default, so `Signature.bind`'s TypeError already covered the rest. Empirically confirmed key identity is unchanged on deploy: omitted / positional / explicit `sweep_slot=0` all build `qo_…sweep_reply_comments_sweep_slot-0_user_id-7`, `attempt` stays out of the key, distinct slots stay distinct. Both lock paths agree (`apply_async` and the `unlock_before_run` `__call__`/`after_return` clear go through the same `_get_call_args`), so an omitted key can't leak a lock. All four `sweep_reply_comments` enqueue sites pass a slot. `tests/unit/app` + `tests/unit/api/test_main_general.py`: 2129 passed.

- **Nothing stopped a future task from importing the upstream class** (test gap): the PR's thesis is that the bug class is gone, but that only holds while every task base is the shim — `from celery_once import QueueOnce` in a new module silently restores the KeyError with no runtime signal. Added `TestEveryTaskUsesTheShim` (source-text guard, same shape as `test_selenium_queue_routing`): no `src/cqc_lem` module may import `celery_once.QueueOnce`, and every module using `base=QueueOnce` must import `cqc_lem.app.queue_once`.
- **Three rationale comments now state the opposite of reality**: `run_scheduler.py:412`, `api/main.py:2386` and `tests/unit/api/test_main_general.py:392` all said QueueOnce "does not apply function defaults — omitting it raises KeyError at enqueue". After this PR it does. Rewrote them to the reason the call sites actually still pass it: slot 0 is the single-shot trigger, the golden-hour amplifier owns the other slots.
- **Missing type hints on `_get_call_args(self, args, kwargs)`** (CLAUDE.md requires them on all signatures) — annotated.

Close is honest: #989 is a single-exception auto-filed issue with no phased scope, and the diff + tests satisfy both acceptance criteria.

gitchrisqueen:

gitchrisqueen:

gitchrisqueen:

PR #1024, opened 2026-08-03
fix(api): gate the docs surface under /api, hide admin routes from the schema, trim /health/deep (closes #1020)

### LINKED ISSUE #1020: security(api): FastAPI docs + OpenAPI schema are fully public; trim worker topology from unauthenticated /health/deep
## Finding

The FastAPI docs surface is **not gated by anything**. `src/cqc_lem/api/main.py:165` is a bare
`app = FastAPI()`, so Swagger lives at `/docs`, ReDoc at `/redoc` and the schema at
`/openapi.json` — all **outside** `/api`. `_api_token_required()` (`main.py:254`) only fires on
paths starting with `/api/`, so none of the three ever reaches the bearer check.

Verified live through Cloudflare on 2026-08-03:

| Path | Result |
|---|---|
| `GET /docs` | **200**, 1944 bytes |
| `GET /redoc` | **200**, 1826 bytes |
| `GET /openapi.json` | **200**, 179,071 bytes |
| `GET /api/docs` | 401 `{"status_code":401,"detail":"Unauthorized"}` |
| `GET /api/openapi.json` | 401 (same) |

The edge access log independently confirms it: an external browser (Mac Chrome, client
`2601:346:c00:…`) pulled `/docs`, `/openapi.json` and `/redoc` successfully on 2026-08-03.

LEM is about to carry real users' LinkedIn sessions. This should close before that.

### Correcting the record: `ADMIN_SECRET` was never involved

`/api/docs` returns **401, not 404**, because `api_token_middleware` (`main.py:285-292`) runs
**before route matching** and rejects any non-public path under `/api/`, including paths that
don't exist. That 401 comes from the bearer check against `API_ACCESS_TOKENS` — `ADMIN_SECRET`
plays no part in it. A typo'd docs URL therefore *looks* like a credential wall when it is really
a prefix match on a route that isn't there.

---

### 1. Move the docs to the URL people actually guess

Set `docs_url="/api/docs"`, `redoc_url="/api/redoc"`, `openapi_url="/api/openapi.json"` on the
`FastAPI()` constructor (`main.py:165`), and add exactly those three **leaf** paths to
`_PUBLIC_API_PREFIXES` (`main.py:235-238`) — leaf entries, **not** a `/api/docs/` subtree. The
boundary logic at `main.py:249` (`path == prefix or path.startswith(prefix + "/")`) handles the
distinction.

Keep a **301 from `/docs`** so existing links keep working:
- `docs/TESTING_ENGAGEMENT_API.md:15,16,43,72`
- `run.sh:159,171,181`

**Gotcha, verified against the running image (FastAPI 0.140.0):**
`swagger_ui_oauth2_redirect_url` defaults to the fixed literal `/docs/oauth2-redirect` — it is
**not** derived from `docs_url`. Moving `docs_url` strands that helper outside `/api` and the
Swagger "Authorize" flow breaks silently. Set it explicitly to `/api/docs/oauth2-redirect`; the
leaf entry above already covers it, since the non-slash branch matches path-segment children.

### 2. Hide the admin routes from the public schema

Confirmed against the **live** `/openapi.json`: 146 paths / **170 operations**, of which **18** are
under `/api/admin` and fully published today:

```
POST /api/admin/automation-pause                POST /api/admin/regenerate-carousel
POST /api/admin/automation-resume               POST /api/admin/regenerate-video
GET  /api/admin/automation-status               GET  /api/admin/task-status/{task_id}
POST /api/admin/consolidate-duplicate-comments  POST /api/admin/test/comment
GET  /api/admin/feedback                        POST /api/admin/test/dm
POST /api/admin/feedback/{feedback_id}/review   POST /api/admin/test/dm-direct
POST /api/admin/fix-video-urls                  POST /api/admin/test/reply
POST /api/admin/generate-media-variants         POST /api/admin/user/location
                                                GET  /api/admin/youtube-status
                                                POST /api/admin/youtube-token
```

That is a targeting map for anyone probing `ADMIN_SECRET`. Use `include_in_schema=False`, or mount
a separate hidden sub-app. The routes themselves keep working exactly as now — this hides them from
the schema, it does not change their auth.

### 3. Stop naming the env vars in the security-scheme descriptions

`main.py:6400` literally says *"one of API_ACCESS_TOKENS"* and `main.py:6406` says *"the
ADMIN_SECRET env var"*. Describe the credential, not where it is stored.

### 4. Drop `lanes` from the unauthenticated `GET /health/deep`

`main.py:1804-1866`. **Decision made — implement this, do not re-open it.** Remove the `lanes` key
from the response dict at `main.py:1866`. **Keep `status`, `workers`, `consuming` and
`maintenance` exactly as they are.**

Current live body (348 bytes, fetched through Cloudflare):

```json
{"status":"healthy","workers":5,"consuming":5,"maintenance":false,
 "lanes":{"selenium-se_engage-worker@3571c22235c8":["se_engage"],
          "selenium-se_prepost-worker@f4aab4feea3c":["se_prepost"],
          "selenium-se_content-worker@1c93cd67e91e":["se_content"],
          "selenium-se_outreach-worker@440c229a3e3d":["se_outreach"],
          "main-worker@celery-worker-host":["celery"]}}
```

Target body:

```json
{"status":"healthy","workers":5,"consuming":5,"maintenance":false}
```

**Why `lanes` and only `lanes`.** It is the sole field leaking anything — container IDs and the
internal queue topology. `workers` and `consuming` are bare integers with no disclosure value, and
`consuming` is the field that *decides* `status`; removing it would make a `degraded` reading
unexplainable. `lanes` is computed internally either way, so `workers`/`consuming` are unaffected.

> ### ⛔ `status` and the literal `"status":"healthy"` must NOT change
>
> Per `docs/stack-watchdog.md:148-155` that literal is a **monitor contract**: renaming the field
> or the value silently disarms every configured monitor, and a monitor that can no longer match
> looks exactly like a monitor that is passing.
> `tests/unit/api/test_health_deep.py::test_healthy_keyword_is_stable_for_body_assertions` pins it
> and must keep passing untouched.
>
> This change is safe precisely because `status` is the **first** key and FastAPI preserves dict
> insertion order — removing `lanes`, the **last** key, leaves the literal `"status":"healthy"`
> **byte-identical**.

---

## VPS review findings

Read-only inspection of the production VPS (`srv1781491`), 2026-08-03. Nothing was changed,
restarted or deployed. Prod was on `v0.129.1`.

### The one real consumer: UptimeRobot (external, off-box)

The edge nginx (`web_app`) access log shows **54 requests** to `/health/deep` from
`Mozilla/5.0+(compatible; UptimeRobot/2.0; http://www.uptimerobot.com/)`, all **HTTP 200**, via
`https://lem.christopherqueenconsulting.com/health/deep`. This is layer 3 from
`docs/stack-watchdog.md` — it is wired up and live.

The only other hits were ad-hoc human checks: 2× `curl/8.20.0` and 2× iPhone Safari.

Which keyword that monitor asserts on is **not observable from the box** — UptimeRobot's config
lives in its own dashboard. That is the single owner sign-off this PR needs; see **Merge gate**
below.

### Everything that could have been a consumer, and is not

| Candidate | What it actually does |
|---|---|
| `lem-watchdog.service` / `.timer` (root, every 5 min) | Runs `/opt/lem/scripts/stack_watchdog.sh`, which uses **`docker ps` / `docker start`** — it never curls the API. Its only `curl` calls are the PostHog capture and the SendGrid alert. |
| `scripts/deploy.sh` | Health-gates on plain **`/health`** (lines 331, 403), not `/health/deep`. |
| `lem` crontab (7 jobs) | recovery-probe, model-check, agent-pipeline `tick.sh`, li-version-check, perf-tracking, error-issues-cron ×2 — greps over all of them return no `/health/deep` reference. |
| `deploy` crontab | one job: `scripts/backup.sh`. `root` crontab: empty. |
| `/etc/crontab`, `/etc/cron.d`, `/etc/cron.{hourly,daily,weekly,monthly}` | Stock Ubuntu only (e2scrub, sysstat, monarx-update, apt/dpkg/man-db/logrotate). |
| systemd timers (19) | Only LEM units are `lem-watchdog` (above), `lem-update-report` (unattended-upgrades summary) and `claude-update`. No health/monitor/uptime unit. |
| Monitoring agents | **None installed.** `dpkg -l` finds no netdata, telegraf, datadog, node_exporter, prometheus, zabbix, collectd, nagios or monit. Listening ports corroborate: sshd, tailscaled, `monarx-agent` (Hostinger "Security Scanner", localhost-bound), ollama, docker-proxy ports, and an agent-pipeline python on `127.0.0.1:8765`. |
| Docker healthchecks (9 defined) | None touches `/health` or `/health/deep`: `redis-cli ping`, `mysqladmin ping`, `celery inspect ping` ×4, a `/proc` scan for beat, flower `/metrics`, selenium `/status`, litellm `/health/liveliness`. |
| nginx edge (`web_app`, nginx:1.27-alpine) | A single catch-all `location /` proxying to the active color (`/opt/lem/deploy/nginx/default.conf`). **No location block, no keyword or log-based alerting, no rate limit on this path.** No fail2ban on the box. |
| Cloudflare Tunnel | Dashboard-managed — no config file on the box; the container runs `tunnel --no-autoupdate run`. Ingress read from its log: `lem→web_app:8000`, `lemflower→flower:8555`, `lemlitellm→litellm:4000`, `lemvnc→selenium:7900`, `lemgrid→selenium:4444`, else `http_status:404`. |

A box-wide grep for `health/deep`, `lanes` and `consuming` outside the repo returned **zero**
hand-written consumers — every hit is inside a git checkout of this same repo (agent-pipeline
worktrees, `/home/lem/{model-check,li-version-check,error-issues-cron,freezefix}`,
`/opt/lem/CHANGELOG.md`). The `freezefix` hits are Selenium **capacity** lanes, unrelated to this
response field.

### Cloudflare posture (for the docs half)

- **No Access policy or WAF rule is blocking `/docs`.** The 200s above came back with a normal
  `cf-ray` and no Access interstitial. The "maybe Cloudflare is already returning 403 in front of
  it" hypothesis is **disproven** — the docs are genuinely public.
- The Cloudflare **dashboard** (Zero Trust policies, Health Check monitors) is not inspectable
  from the box. No CF Health Check traffic appears in the edge log, but absence there is weak
  evidence — owner-verifiable-only.

### Env presence (values not read)

- `API_ACCESS_TOKENS` — **set** in `/opt/lem/.env`. So `_API_ACCESS_TOKEN_SET` is truthy and the
  middleware gate is live, which is why `/api/docs` 401s rather than 404s.
- `ADMIN_SECRET` — **set**.

### Options rejected for #4

- *Gate the whole endpoint behind `ADMIN_SECRET`* — breaks the UptimeRobot monitor outright and
  defeats the endpoint's purpose (an external dead-man's switch that needs no credential).
- *Keep it and do nothing* — leaves internal topology public as real user sessions arrive.
- *Split a detailed variant onto a separate authenticated path* — reasonable, but there is no
  consumer for it. `stack_watchdog.sh` already reads the same state via `docker ps` on-box.

---

## Merge gate (`risk:security`)

Build the whole scope, then **park the PR with `needs-human` + a Decision Comment** asking the
owner to confirm, before merge:

> The UptimeRobot monitor on `https://lem.christopherqueenconsulting.com/health/deep` — is its
> keyword `"status":"healthy"` with Keyword Type **"does not exist"**, as
> `docs/stack-watchdog.md:148-151` prescribes? If it is anything else, change it to that and
> confirm one green check **before** merging this PR.

Reconfigure first, ship second. The byte-identical argument above means a correctly configured
monitor is unaffected, but that is an assumption about off-box state the agent cannot verify.

## Acceptance criteria

- [x] `docs_url`/`redoc_url`/`openapi_url` moved to `/api/docs`, `/api/redoc`, `/api/openapi.json`;
      `swagger_ui_oauth2_redirect_url` set explicitly to `/api/docs/oauth2-redirect`.
- [x] Those paths added to `_PUBLIC_API_PREFIXES` as **leaf** entries (no trailing slash).
- [x] 301 redirect from `/docs` (and `/redoc`, `/openapi.json`) to the new paths.
- [x] `docs/TESTING_ENGAGEMENT_API.md:15,16,43,72` and `run.sh:159,171,181` updated to the new URLs.
- [x] All 18 `/api/admin/*` operations absent from the generated OpenAPI schema; their runtime auth
      and behaviour unchanged.
- [x] `main.py:6400` and `main.py:6406` no longer name `API_ACCESS_TOKENS` or `ADMIN_SECRET`.
- [x] `lanes` removed from the `/health/deep` response; `status`, `workers`, `consuming`,
      `maintenance` all still present and unchanged in name, type and order.
- [x] `tests/unit/api/test_health_deep.py::test_healthy_keyword_is_stable_for_body_assertions`
      passes **unmodified**.
- [x] `test_reports_each_worker_and_its_lanes` updated for the new shape. Note `out["workers"]` is
      asserted in **six** other tests in that file — those must keep passing as-is, which they do
      because `workers` is retained.
- [x] `docs/stack-watchdog.md:77-150` updated: the sample body, the field table, and the "Why
      `consuming`, not `workers`" section all currently show `lanes`.
- [x] New tests: Swagger/ReDoc/schema reachable at the `/api/*` paths; the old paths 301; **no**
      `/api/admin/*` operation appears in the generated schema. Without this last one the admin-route
      hiding regresses silently the next time a route is added.
- [ ] ≥80% patch coverage (Codecov).

gitchrisqueen:
## What & why

The FastAPI docs surface was at the framework defaults — `/docs`, `/redoc`, `/openapi.json` —
which is **outside `/api`**, the only prefix `_api_token_required()` inspects. All three were
served to anyone (verified live through Cloudflare on 2026-08-03), and the published schema listed
all **eighteen** `/api/admin/*` operations by method and path: a targeting map for anyone probing
the admin secret, on a document that needs no credential to fetch.

Four changes, per the issue's scope:

**1. Docs move under `/api`.** `docs_url="/api/docs"`, `redoc_url="/api/redoc"`,
`openapi_url="/api/openapi.json"`, added to `_PUBLIC_API_PREFIXES` as **leaf** entries (the
non-slash branch already matches path-segment children, so `/api/docs/oauth2-redirect` is covered
while a future `/api/docs-admin` is not). `swagger_ui_oauth2_redirect_url` is set **explicitly** —
FastAPI defaults it to the fixed literal `/docs/oauth2-redirect` and does *not* derive it from
`docs_url`, so moving `docs_url` alone strands the helper outside `/api` and breaks Swagger's
Authorize flow silently. The three old paths **301** to the new ones, so bookmarks, README links
and Postman imports keep working.

**2. Admin routes out of the schema.** `_hide_admin_routes_from_schema()` runs after
`include_router` and flips `include_in_schema` on every `/api/admin/*` route. It is **derived from
the route table**, not written as `include_in_schema=False` on eighteen decorators, because that
failure mode is silent — a nineteenth admin route added later would publish itself and nothing
would say so. `_ADMIN_ROUTES_HIDDEN` records how many it reached, so a walk that matches nothing
fails loudly rather than leaving the schema quietly unchanged (FastAPI ≥0.139 keeps an included
router as one opaque `_IncludedRouter` node — the first version of this loop hit exactly that and
hid zero).

**Hidden is not gated:** runtime auth is untouched. The existing 403-without-header /
200-with-header tests still pass unchanged.

**3. Descriptions name the credential, not the env var.** `API_ACCESS_TOKENS` and `ADMIN_SECRET`
no longer appear in the security-scheme descriptions.

**4. `lanes` dropped from `GET /health/deep`.** That endpoint is unauthenticated **by design** — an
external dead-man's switch cannot carry a credential — and the per-worker map was the one field
with disclosure value, naming container IDs and the internal queue topology.
`workers`/`consuming`/`maintenance` stay: bare counts name nothing, and `consuming` is the field
that *decides* `status`. `lanes` is still computed internally, so those counts are unaffected.

> `status` stays the **first** key and FastAPI preserves insertion order, so removing the **last**
> key leaves the literal `"status":"healthy"` — the monitor contract in `docs/stack-watchdog.md` —
> **byte-identical**. `test_healthy_keyword_is_stable_for_body_assertions` passes **unmodified**.

## Consequence worth knowing

Swagger's **Try it out** can no longer drive the `/api/admin/test/*` endpoints, because they are
not in the schema any more. With no visible operation referencing them, FastAPI also stops emitting
the two `securitySchemes` — so the Authorize dialog is gone. Nothing regresses for non-admin routes
(none of them ever declared `security`, so Swagger never sent the bearer for them). `curl` and
Postman are the documented paths and both already send both headers;
`docs/TESTING_ENGAGEMENT_API.md` §3 now leads with curl and says so explicitly.

## Testing

- New `tests/unit/api/test_docs_surface.py` (12 tests): the three paths answer on `/api/*`, the
  oauth2-redirect helper moved with them, the old paths 301, the public entries are leaves (and
  `/api/docs-admin` is *not* unlocked), **no** `/api/admin/*` path is in the generated schema
  (guarded against a vacuous pass on an empty schema), the hide count reached ≥18, the routes still
  exist and still 403, and neither scheme description names an env var.
- `test_health_deep.py`: `test_reports_each_worker_and_its_lanes` → `test_counts_workers_and_
  consumers_without_naming_them`, plus a shape test asserting the exact key set (so a future field
  re-adding the same disclosure under a different name also fails) and a key-order test. The six
  other tests asserting `out["workers"]` pass as-is.
- `TestAuthSchemeInOpenAPI` → `TestAdminCredentialsAreNotAdvertised`: the old test asserted the
  admin schemes *are* published, which is now the thing we are preventing. It pins the inverse plus
  runtime enforcement.
- Full unit lane green locally: **10292 passed**.

## Owner decision — `1A 2B 3A` (answered 2026-08-04)

- **1A — verify the UptimeRobot keyword first, then merge.** Owner-side; the dashboard is off-box
  state this branch cannot read. Made durable in `docs/stack-watchdog.md` §Layer 3: the
  `"status":"healthy"` literal is the contract, the **field set is not**, and #1020 changed it by
  dropping `lanes`. A monitor asserting on the whole body or a byte length breaks on a change that
  is safe by design — and breaks looking like a passing check, which is the one failure this layer
  cannot report about itself.
- **2B — `release:now`.** Ships at merge rather than waiting for the next 05/11/17/23 UTC batch.
  Rationale (per `docs/release-fast-lane.md`): `priority:high` + `risk:security`, and the published
  schema listing all 18 `/api/admin/*` operations is live and unauthenticated until this deploys.
  It skips the wait, never a check.
- **3A — accept losing Swagger for the admin endpoints.** No code change; `curl` is already §3 of
  `docs/TESTING_ENGAGEMENT_API.md` and the §4 Postman collection sends both headers. No second
  admin-gated docs app was mounted (that was 3B, declined).

Closes #1020


gitchrisqueen:

Held (`needs-human`, risk: security). Everything in #1020's scope is built and the unit lane is green — but one thing this PR depends on is **off-box state I cannot read**: how the UptimeRobot monitor on `https://lem.christopherqueenconsulting.com/health/deep` is actually configured. The edge log proves that monitor exists (54 × HTTP 200), not what it asserts on.

### 1. The UptimeRobot keyword — confirm before merge?

`docs/stack-watchdog.md:148-151` prescribes Keyword Type **"does not exist"**, keyword `"status":"healthy"`. If it is that, this PR cannot affect it: `status` is the first key, FastAPI preserves insertion order, and `lanes` was the last key — so the literal is byte-identical before and after. If it asserts on anything else (e.g. the whole body, a byte length, or `lanes`), the monitor breaks the moment this deploys, and a monitor that can no longer match looks exactly like one that is passing.

- **A. You check the dashboard, set it to keyword `"status":"healthy"` / "does not exist" if it isn't, confirm one green check, then merge** — reconfigure first, ship second; zero window where the dead-man's switch is silently disarmed  ✅ *recommended*
- **B. Merge now, check after** — if the keyword was something else, the monitor is quietly dead from the deploy until you notice
- **C. Keep `lanes` and close the rest of #1020** — monitor untouched, but the container IDs and queue topology stay public as real user sessions arrive, which is the thing the issue exists to fix

### 2. Ship in the next release window, or fast-lane it?

Releases batch 4× daily (median ~168 min). This is `priority:high` + `risk:security`, which qualifies for `release:now` under `docs/release-fast-lane.md` — but it also makes the docs URL change land sooner than you may want, and the fix closes an exposure that has been live for a while rather than a live breakage.

- **A. Normal release window** — ships at the next 05/11/17/23 UTC batch; no extra risk  ✅ *recommended*
- **B. `release:now`** — ships at merge; sensible if you'd rather not leave the schema public another few hours

### 3. Losing Swagger for the admin endpoints — accept?

Hiding all 18 `/api/admin/*` operations also removes the two `securitySchemes` from the schema (nothing visible references them any more), so Swagger's **Authorize** dialog and **Try it out** no longer drive the admin/test endpoints. Their auth and behaviour are unchanged; `docs/TESTING_ENGAGEMENT_API.md` §3 now leads with the curl recipe and §4's Postman collection already sends both headers.

- **A. Accept — use curl/Postman for admin test runs** — simplest, and the admin paths stay unpublished  ✅ *recommended*
- **B. Mount a second, admin-secret-gated docs app at e.g. `/api/admin/docs`** — keeps Try-it-out, but adds a new authenticated surface (and a new way to get its gate wrong); I'd file it as a follow-up rather than grow this PR

**My recommendation: `1A 2A 3A`.** 1A is the only option that removes the exposure without betting on an assumption about a dashboard I can't see; 2A because this is a long-standing exposure rather than an outage, so it doesn't need to jump the queue; 3A because curl and Postman are already documented and working, and 3B trades a real security win for convenience.

gitchrisqueen:
1A 2B 3A

gitchrisqueen:
Addressed your review — applied `1A 2B 3A`.

**2B — `release:now` applied.** Label is on the PR, so it ships at merge instead of the next 05/11/17/23 UTC batch. Rationale is now a line in the PR body so the call is auditable: `priority:high` + `risk:security`, and the schema listing all 18 `/api/admin/*` operations is public and unauthenticated until this deploys. It skips the wait, not a check — CI is green across all 23 required checks on the previous head, and this push re-runs them.

**1A — one commit, `1550b21b`.** The dashboard check itself is yours (it is off-box state I cannot read), so what I could do was make the *reason* for it durable rather than let it live in a decision comment that scrolls away. `docs/stack-watchdog.md` §Layer 3 now says: assert on the `"status":"healthy"` keyword **and nothing else** — that literal is the contract, the field set is not, and #1020 changed the field set by dropping `lanes`. A monitor configured against the whole body, a byte length, or any other field breaks on a change that is safe by design, and breaks in the direction that looks like a passing check.

> Ordering worth stating plainly, since 2B removes the buffer 1A assumed: with `release:now` the deploy follows the merge rather than waiting hours. If the keyword check is still outstanding on your side, do it before you let this merge — that is the sequence 1A picked ("reconfigure first, ship second"), and 2B just makes the second step immediate. If you have already confirmed it, nothing else is needed.

**3A — no code change.** Swagger no longer drives the admin/test endpoints and the Authorize dialog is gone with the two `securitySchemes`; `docs/TESTING_ENGAGEMENT_API.md` §3 already leads with the curl recipe and §4's Postman collection sends both headers. I did **not** mount an admin-secret-gated docs app (3B) and did not file a follow-up for it, since you declined it rather than deferring it — say the word if you want it tracked.

Nothing else from the decision comment was left undone. Handing back to the merge loop.

gitchrisqueen:

Reviewed the full diff, the seams it touches (`_is_public_api_path` / `api_token_middleware`, the session-scope and CSRF layers, the route table, the prod nginx edge, every `/health/deep` consumer on the box per #1020's own VPS sweep) and issue #1020's eleven acceptance criteria. The four scope items are all implemented and the close is honest — nothing is deferred to a later phase.

**What I checked and found sound** (recording it so the next reader doesn't re-derive it): the `lanes` drop has no consumer anywhere in the repo or on the box beyond UptimeRobot; `status` stays the first key so the `"status":"healthy"` literal is byte-identical; `_hide_admin_routes_from_schema` can't double-count on a duplicated route table (the `include_in_schema` guard makes it idempotent) and reaches all 18; the prod edge is a single catch-all `location /` proxy, so the `/docs` 301s do fire through Cloudflare; `/api/docs` is exempt from the CSRF and scope layers only because both are GET-side no-ops, not by accident; and no stale `/docs` or `/openapi.json` reference survives in `run.sh`, the SPA, CI, scripts or docs.

**Findings — both fixed in `0ad7c266`.** They are the same shape as #1020 itself: behaviour that only exists in the production config, so the green CI run said nothing about it.

- **The public-prefix entries were untested under the gate that makes them matter.** `_api_token_required` returns False for *every* path when `API_ACCESS_TOKENS` is unset — which is the CI config — so `test_reachable_at_the_new_path` and friends would have passed identically if `/api/docs`, `/api/redoc` and `/api/openapi.json` had never been added to `_PUBLIC_API_PREFIXES`; only `test_new_paths_are_public_leaf_entries_not_subtrees` touched the real logic, and it calls the predicate rather than the app. Added `test_the_surface_answers_with_no_credential_once_the_gate_is_LIVE`: with tokens configured, all three paths plus `/api/docs/oauth2-redirect` answer 200 uncredentialed while `/api/docs-admin` and `/api/admin/automation-status` 401. Confirmed non-vacuous — trimming the three entries makes all three 401.
- **The `/docs` 301 was never actually raced against the SPA catch-all.** `src/cqc_lem/ui/dist` does not exist in a checkout (the SPA is baked into the image), so `@app.get("/{full_path:path}")` is never registered under pytest — the redirect tests ran with nothing competing. In prod that catch-all matches `/docs`, and the redirects win *only* because Starlette matches in registration order, which is exactly the claim `main.py`'s own comment makes and nothing pinned. Added `test_the_legacy_redirect_beats_the_SPA_catch_all`, which registers a stand-in catch-all (reverted in `finally`) and asserts both the three 301s and that the stand-in was genuinely reachable. Without it, moving those three registrations below the SPA block would turn every legacy docs link into a 200 HTML page and no test would fail.
- *(Doc, folded into the same commit.)* `TESTING_ENGAGEMENT_API.md` advertised the OpenAPI schema as the Postman import path in §1 while §2 sends the reader to Postman for the now-hidden admin endpoints — importing the schema gives none of them. Now says so and points at the checked-in collection, which lists them explicitly and does not depend on the schema.

Full unit lane green on the new head: **10294 passed**.

Not a finding, flagged for the merge gate only: **1A is still owner-side.** The UptimeRobot keyword is off-box state; the byte-identical argument holds, but confirm the monitor asserts on `"status":"healthy"` / Keyword Type "does not exist" before letting this merge, since `release:now` means the deploy follows the merge immediately.

gitchrisqueen:
🧩 phase-guard — heads up: issue #1020 still has **boxes: 1 unchecked** acceptance box(es). Merging anyway (unticked boxes are common here), but if any of that scope did NOT ship, file a follow-up issue with `agent:ready` and link it, or drop the `Closes #1020` from this PR.

gitchrisqueen:

gitchrisqueen:

gitchrisqueen:

PR #1037, opened 2026-08-04
fix(feedback): make triage Approve actually file, and say what it did (closes #1036)

### LINKED ISSUE #1036: fix(ui): Fix approve button in feedback triage
## Why
Clicking approve on the feedback triage items does nothing and is not working.

Auto-filed from in-app feedback #39 by the feedback→auto-work loop (`docs/launch-and-marketing-plan.md` §B.2/B.3). Classifier: bug/high, risk `none`, confidence 1.00.

[Watch the session replay](https://us.posthog.com/project/475262/replay/019fc782-ec6f-7071-ab02-f7a1e2546ba1) — the browser session this report was filed from.

## Scope
- Address the reported bug in the `ui` area of the product.

## Files
- `src/cqc_lem/ui/src/`
- `tests/unit/`

## Acceptance
- All required CI gates pass.

gitchrisqueen:
## What was broken

Clicking **Approve** in the admin feedback triage panel did nothing visible.

`POST /api/admin/feedback/{id}/review` ran the *unattended* filer path unchanged, so an explicit
human approval was re-gated by holds that only exist because the batch pass runs with nobody
watching. Three filer outcomes leave the row **exactly as they found it**:

| outcome | status write | what the admin sees |
|---|---|---|
| `filed` / `deduped` / `dropped(noise)` / `faq` | yes | row moves ✅ |
| `rate_limited` — reporter over the per-user daily cap | **none** | row unchanged ❌ |
| `dropped(empty body)` | **none** | row unchanged ❌ |
| `error` — GitHub refused | **none** | row unchanged ❌ |

All three return **200**, the panel invalidates the list, and the row re-renders identically. The
only signal was a small grey `Last review: approved — filer: rate_limited` line **above** the table.

Two of those are also wrong on their own terms:

- **`rate_limited` is unrecoverable here.** The comment says "held for a later pass" — but the panel
  only ever holds **non-admin** rows, and `process_new_feedback` drains `get_unprocessed_feedback(admin_only=True)`.
  There is no later pass. Held meant held forever.
- **`needs_human` is a loop with no exit.** It parks the row in `triaged` because "a person should
  look at this before it reaches the backlog" — and a person just did. It leaves the queue, still with
  no issue, and no button that can ever produce one.

## The fix

**`file_feedback_issue(..., admin_approved=True)`** — the ONE path that is not unattended.

- Skips the per-user daily issue cap. An admin approving one row by hand is the opposite of the
  firehose that guard exists for.
- Files a `NEEDS_HUMAN` row instead of re-queueing it. Confidence still shapes the **labels**, so a
  shaky classification lands `needs-human` + assigned + a Decision Comment and never `agent:ready` —
  nothing gets built unattended off a low-confidence read.
- **NOISE and FAQ are unchanged.** Those settle the row, so the panel already shows the admin what
  happened, and forcing an issue for something classified as noise is not what Approve is for.
- **The batch pass keeps every hold** — covered by its own test.

**An empty body now settles as `dismissed`.** It can never become an issue, so leaving it `new` both
re-offered it to every later pass and made Approve a visible no-op.

**The response carries `filed`.** A 200 only says the review was recorded; a GitHub refusal changes
nothing else and is otherwise indistinguishable from a successful approve.

**The panel says what happened, at the button.** `FILING_OUTCOMES` maps each filer verb to a
sentence an admin can act on (`GitHub refused the filing — nothing changed. Try Approve again.`),
rendered in the acted row's Actions cell — the banner alone sits above a table the admin has
scrolled past — and amber, not grey, when nothing was filed. The button reads `Working…` while the
mutation is in flight.

## Testing

- `tests/unit/utilities/test_feedback_issue_service.py` — approve is not re-gated by either hold;
  low-confidence approve files but stays held/assigned/not-`agent:ready`; NOISE + FAQ verdicts still
  respected; the batch pass still gets every hold; empty body settles (and doesn't write without an id).
- `tests/unit/api/test_feedback_admin_api.py` — the filer is called with `admin_approved=True`;
  `filed` is true for filed/deduped and **false** for a GitHub error.
- `AdminFeedbackPage.test.tsx` — a no-issue approve is reported in the row's cell and not as a
  success; a successful approve names the issue; the click sends the clicked row's id.
- `poetry run pytest tests/unit -q` → **10464 passed**. `eslint` clean, `tsc --noEmit` clean.
  vitest could not run locally (node 18 on this box; the suite needs 22) — CI is the source of truth.

Docs: `docs/launch-and-marketing-plan.md` §B.2 now records that Approve is a human decision and which
holds therefore do not apply to it.

`release:now`: user-visible breakage on a `priority:high` bug — an admin surface that silently does
nothing is worth shipping in the next flip rather than the next window.

Closes #1036

### DIFF
--- a/docs/launch-and-marketing-plan.md
+++ b/docs/launch-and-marketing-plan.md
@@ -214,6 +214,11 @@ classifier/filer path and stamps `feedback.reviewed_by/reviewed_at`. Two consequ
   `agent:ready`. The NOISE and FAQ verdicts are unchanged: those settle the row, so the panel already shows
   the admin what happened. What can still file nothing is GitHub refusing — the response carries `filed:
   false` and the panel says so **at the button**, because that outcome changes nothing else on screen.
+  A low-confidence verdict and NO verdict are NOT the same thing: the fail-safe classification (`errors`
+  set — the LLM never answered) carries the RAW report as its `summary`, so approving it would publish the
+  feedback text §B.2 promises never reaches GitHub, under `issue_created` — which is terminal. That case
+  files nothing, leaves the row in `new`, and returns `reason: classification unavailable` so the panel
+  tells the admin to retry rather than blaming GitHub.
 `count_pending_admin_review()` reports the backlog on every filing pass so it can't grow silently.
 
 ### B.3 Deduplication / clustering (one recurring problem = one issue)
--- a/docs/launch-and-marketing-plan.md
+++ b/docs/launch-and-marketing-plan.md
@@ -205,6 +205,15 @@ classifier/filer path and stamps `feedback.reviewed_by/reviewed_at`. Two consequ
   re-running the filer on it would match it to itself at similarity 1.0 and post a false "+1 another report"
   on the very issue it created. The panel's buttons alone are not the guard: its list is cached, and the
   filing beat or a second admin can settle a row between render and click.
+- **Approve is a human decision, so the unattended holds do not re-apply to it (issue #1036).**
+  `file_feedback_issue(..., admin_approved=True)` skips the per-user daily issue cap and files a `NEEDS_HUMAN`
+  (low-confidence) row instead of parking it back in `triaged`. Both holds exist because the batch pass runs
+  with nobody watching; re-applying them to an explicit approve returned 200 and left the row in `new`, so the
+  panel re-rendered it unchanged — which is what "the approve button does nothing" was. Confidence still
+  shapes the LABELS, so a shaky classification lands `needs-human` + assigned + Decision Comment and never
+  `agent:ready`. The NOISE and FAQ verdicts are unchanged: those settle the row, so the panel already shows
+  the admin what happened. What can still file nothing is GitHub refusing — the response carries `filed:
+  false` and the panel says so **at the button**, because that outcome changes nothing else on screen.
 `count_pending_admin_review()` reports the backlog on every filing pass so it can't grow silently.
 
 ### B.3 Deduplication / clustering (one recurring problem = one issue)
--- a/src/cqc_lem/api/main.py
+++ b/src/cqc_lem/api/main.py
@@ -7002,7 +7002,7 @@ def admin_feedback_review(
 ) -> ResponseModel:
     """Approve a feedback row for auto-triage or dismiss it."""
     reviewer_user_id = _require_user_admin(request.session_token)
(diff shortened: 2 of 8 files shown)


gitchrisqueen:

- **An approve could file an issue whose classification never happened, terminally.** The new `admin_approved` fall-through filed *every* `NEEDS_HUMAN` row, but low-confidence and no-verdict are different things. `classify_feedback` returns `_unclassified` when the LLM call fails or answers off-contract — confidence 0.0, `errors` set, `summary` = the **raw report**, `title` = the reporter's first line. `build_issue_body` puts `summary` straight under `## Why`, so a LiteLLM blip during an approve published the raw feedback text that `issue_service`'s own docstring promises never reaches GitHub — and then wrote `issue_created`, which is in `_REVIEW_SETTLED_STATUSES`, so the row 409s forever and the report is spent on a content-free issue. Fixed: `classification.errors` is exactly that marker, so the approve path now files nothing, leaves the row in `new`, and returns `reason: classification unavailable`; the panel words it as the classifier being unreachable instead of `GitHub refused the filing`, which pointed the admin at the wrong system. `empty body` now reads as empty rather than as a noise verdict — that is what the `reason` field this PR added to `ReviewResult` (previously unread) is for. Tests: `test_admin_approval_will_not_file_a_classification_that_never_happened` + a vitest case asserting GitHub is not blamed. Docs §B.2 updated.

Checked and found sound: the `Closes #1036` is honest (no phases, no unchecked scope); `admin_approved` skipping the per-user cap and re-routing `NEEDS_HUMAN` is correct and the batch pass keeps both holds (covered by its own test); a low-confidence approve still can't reach `agent:ready` because `is_agent_ready` requires `route == AUTO_WORK`; the 409 self-dedup guard still stands; `record_feedback_review` after the filer does not clobber the status it wrote; `X-LEM-Client` is sent by the shared axios client, so the CSRF gate is not the dead-button cause.

`poetry run pytest tests/unit/utilities/test_feedback_issue_service.py tests/unit/api/test_feedback_admin_api.py` → 211 passed. vitest and tsc could not run on this box (node 18, no `node_modules` in the worktree) — CI is the source of truth for the SPA lanes.

gitchrisqueen:

gitchrisqueen:

gitchrisqueen:

Issue #1038, opened 2026-08-04
fix(errors): TimeoutException: Message: Finding Send Connection Button Stacktrace: #0 0x59a29113544a <unknown> #1  (25x)


gitchrisqueen:
## Why
PostHog error tracking grouped this exception into an issue that is still active: **TimeoutException**

> Message: Finding Send Connection Button
Stacktrace:
#0 0x59a29113544a <unknown>
#1 0x59a290b0d8c9 <unknown>
#2 0x59a290b62f32 <unknown>
#3 0x59a290b63181 <unknown>
#4 0x59a290bae374 <unknown>
#5 0x59a290bab55d <unknown>
#6 0x59a290b563ff <unknown>
#7 0x59

- Occurrences (last 24h): **25** across 1 distinct actor(s)
- First seen: `2026-08-03T14:31:06.234000Z` · last seen: `2026-08-03T14:54:31.166000Z`
- Celery task: `cqc_lem.app.run_automation.invite_to_connect`
- SDK: `posthog-python`

[Open the issue in PostHog](https://us.posthog.com/project/475262/error_tracking/019fc809-1768-70b1-b8c0-64576ba09a3f) — it has the stack trace, the grouped occurrences and the affected people.

## Scope
- Fix the root cause of the exception, not the symptom.
- If it is an EXPECTED best-effort failure that already degrades gracefully (a Selenium selector miss, a third-party timeout the caller retries), stop raising it into error tracking: catch it and `log_warning(...)` instead, or add a suppression rule in PostHog.
- Keep the change scoped to this exception — no unrelated refactors.

## Files
Start from the stack trace on the PostHog issue; add/extend the unit tests:
- `tests/unit/`

## Acceptance
- The exception no longer occurs (or is deliberately downgraded to a warning).
- A unit test covers the failing path (≥80% patch coverage).
- All required CI gates pass.

Auto-filed from PostHog Error Tracking. Dedup marker (do not remove): `posthog-issue-019fc809-1768-70b1-b8c0-64576ba09a3f`

PR #1045, opened 2026-08-04
fix(errors): a missing connect-note affordance is a no-op, not a defect (closes #1039)

### LINKED ISSUE #1039: fix(errors): TimeoutException: Message: Finding Add a Note Button Stacktrace: #0 0x6476025b344a <unknown> #1 0x6476 (1x)
## Why
PostHog error tracking grouped this exception into an issue that is still active: **TimeoutException**

> Message: Finding Add a Note Button
Stacktrace:
#0 0x6476025b344a <unknown>
#1 0x647601f8b8c9 <unknown>
#2 0x647601fe0f32 <unknown>
#3 0x647601fe1181 <unknown>
#4 0x64760202c374 <unknown>
#5 0x64760202955d <unknown>
#6 0x647601fd43ff <unknown>
#7 0x647601f

- First seen: `2026-08-03T14:32:00.257000Z` · last seen: `2026-08-03T14:32:00.257000Z`
- Celery task: `cqc_lem.app.run_automation.invite_to_connect`
- SDK: `posthog-python`

[Open the issue in PostHog](https://us.posthog.com/project/475262/error_tracking/019fc809-ea83-7f10-92e4-c8ec59ac7b67) — it has the stack trace, the grouped occurrences and the affected people.

## Scope

## Files
- `tests/unit/`

## Acceptance
- All required CI gates pass.

Auto-filed from PostHog Error Tracking. Dedup marker (do not remove): `posthog-issue-019fc809-ea83-7f10-92e4-c8ec59ac7b67`

gitchrisqueen:
## What

`_add_connect_note` warned with `exc=` on **any** failure to reach the note composer, so a routine
miss filed a fingerprinted `TimeoutException: Finding Add a Note Button` PostHog issue — and this
GitHub issue off it. Its own docstring calls that miss **expected**: LinkedIn hides the note
affordance once a free account's personalized-invite quota is spent, and #573 deliberately made it
non-fatal — the invite goes out bare. A fallback in hand is the test for DEBUG, not WARNING
(`src/cqc_lem/utilities/CLAUDE.md`).

The two ways to lose a note are now graded apart:

| Reading | Level | Why |
|---|---|---|
| No `Add a note` control, dialog still shows `Send without a note` | DEBUG | the documented quota-spent fallback — the invite goes bare |
| Dialog shows neither of its controls | DEBUG | the lost invite is what `_submit_connect_invite` logs as an ERROR; a warning here files a **second** grouped issue for one event (the #1038 fault) |
| Anything failing AFTER the affordance answered (no textarea, `send_keys` refusing emoji) | WARNING + `exc=` | a genuine degraded path — unchanged |

The affordance is resolved with `find_first(required=False, warn_on_miss=False)` (locator list, not a
single raising xpath), and the click that follows keeps `required=True` — the control answered a
moment ago, so failing to click it now is real and raises into the warning path rather than reading
as "not on offer". `_SEND_WITHOUT_NOTE_XPATH` is now shared with `_SEND_INVITE_XPATHS`, so the
cross-check can never disagree with the send that immediately follows it.

## Why not just drop `exc=`

Because the owner's question on the issue is the right one: the same miss is also how selector rot
would look, and **production cannot tell them apart** — a spent quota renders the dialog exactly the
same way. So that reading moved to where it can be attributed: `--connect-dialog` now reports
`note_affordance_present` / `bare_send_present` and names it in its verdict, deliberately **without**
grading it `drift` (the probe account's own quota state is not knowable from the page). Logged as a
tracked gap in `docs/sdui-probe-coverage.md` rather than left to be rediscovered.

No live LinkedIn run was needed to make this call: `_add_connect_note` only runs when
`_open_connect_invite_dialog` has already proven one of the dialog's own controls present, and both
labels (`Add a note` / `Send without a note`) were live-grounded 2026-08-03 in
`docs/sdui-selenium-notes.md`. The one thing a live run would add — what the dialog looks like with
the quota spent — is exactly what the probe field now captures on the next sweep against a target.

## Testing

- `tests/unit/app/test_invite_connect_note.py` — rebuilt around `find_first`/`click_first` stand-ins:
  a missing affordance never warns and logs the quota reading; the reading is grounded in the
  bare-send lookup (asserted by lookup order, not by the note button's absence alone); a dialog
  showing neither control produces exactly ONE signal (the send ERROR); a post-affordance failure
  still warns with `exc=`; happy path, non-BMP stripping, refinement cap and the send-failure error
  unchanged.
- `tests/unit/test_linkedin_live_validation.py` — the probe verdict reports a missing note affordance
  and still grades `ok`; a present one says nothing extra.
- Full lane green locally: `10478 passed`.

Closes #1039

### DIFF
--- a/docs/sdui-probe-coverage.md
+++ b/docs/sdui-probe-coverage.md
@@ -33,7 +33,7 @@ code is missing from this table.
 | Feed share-box composer | `_post_composer_for_card` | `--probe-composer` | yes | per-card miss is a DEBUG no-op by design (#876) |
 | Profile-views viewer list | `_PROFILE_VIEWER_ROWS_JS` | `--profile-views` | yes | zero rows vs the page's headline stat (#1009) |
 | Profile header scrape + degree badge | `parse_profile_header` / `_profile_is_first_degree` | `--profile-scrape` | yes | **none — see Gaps** |
-| Connect invite dialog | `_open_connect_invite_dialog` | `--connect-dialog` | no (needs a target) | dialog controls must be present before Send (#1012) |
+| Connect invite dialog | `_open_connect_invite_dialog` | `--connect-dialog` | no (needs a target) | dialog controls must be present before Send (#1012); a missing note affordance is graded against the bare-send control, never warned (#1039) |
 | Catch-up moment cards | `_CATCHUP_CARD_LOCATORS` | `--catchup-cards` | yes | zero cards vs `main div[role='listitem']` (#1013) |
 | Group share box / editor | `auto_post_to_group` | `--group-composer` | no (needs a group) | `_unpostable` rotates past the group (#858) |
 | Company-page invite modal | `automate_invitations` | `--company-invite` | yes | **none — see Gaps** |
@@ -111,8 +111,16 @@ composer and close it with Escape without typing.
 
 ## Gaps (tracked, not silent)
 
-Three production paths still lack a zero-result tripwire. They are named here rather than left to
+Four production paths still lack a zero-result tripwire. They are named here rather than left to
 be rediscovered:
+
+- **Connect-dialog note affordance** — a missing `Add a note` button is the expected quota-spent
+  fallback (`_add_connect_note` sends the invite bare and logs DEBUG, #1039) and is also what a
+  rotated label would look like, and production cannot tell them apart: the dialog renders exactly
+  the same way once a free account's personalized invites are spent. Warning on it filed a
+  fingerprinted defect per lost note, so the reading moved here — `--connect-dialog` reports
+  `note_affordance_present` / `bare_send_present` and says so in its verdict. It is deliberately
+  NOT graded `drift`: the probe account's own quota state is not knowable from the page.
 
 - **Profile header scrape** — `parse_profile_header` raises `ProfileUnavailableError` on an error
   page, but a page that renders with no `<h1>` and no usable `<title>` yields no name and no
--- a/docs/sdui-probe-coverage.md
+++ b/docs/sdui-probe-coverage.md
@@ -132,5 +132,7 @@ be rediscovered:
 - **Own post stats** — `_post_social_counts` scoring every signal 0 is indistinguishable from a post
   with no engagement.
 
-The probes above grade all three today, so the weekly sweep catches them; the tripwires would catch
-them in production, per run, which is stronger.
+The weekly sweep grades the last three, so it catches them there; the tripwires would catch them in
+production, per run, which is stronger. The connect-dialog note affordance is the odd one out and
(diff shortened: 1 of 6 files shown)


gitchrisqueen:

Reviewed the production change hard and it holds up: `_add_connect_note` is only reachable after `_open_connect_invite_dialog` proved one of the dialog's own controls present, so "no note button + bare-send present" really is the quota reading and "neither control" really is the same lost invite `_submit_connect_invite` is about to log as the one ERROR — the DEBUG on both is right, and `click_first(required=True)` correctly keeps a post-affordance failure in the WARNING + `exc=` path. `_SEND_WITHOUT_NOTE_XPATH` sharing is sound (`contains` is strictly broader than the exact-label dialog locator, so the cross-check can never be stricter than the send). Full unit lane green: **10480 passed**.

Three findings, all fixed on the branch:

- **`probe_connect_dialog` reported a note-affordance reading nobody took** (`scripts/linkedin_live_validation.py`): the two control lookups ran unconditionally, so a profile that rendered NO dialog — an already-pending invite, or genuine drift — landed `note_affordance_present: false` in the JSON a human reads to decide whether the label rotated. That is the exact confusion this PR exists to remove, one layer up. Both are now resolved only when the dialog rendered and are `None` otherwise; `connect_dialog_verdict` matches on `is False` so an unread affordance claims nothing. Side benefit: the probe no longer sits through two extra 10s waits on the run that already found nothing.
- **The doc's own closing claim contradicted the gap it just added** (`docs/sdui-probe-coverage.md`): the Gaps section still ended "The probes above grade all three today, so the weekly sweep catches them" after growing to four — and it is untrue of the new bullet on both counts, since `--connect-dialog` needs a target profile (not in the weekly sweep) and deliberately does not grade the affordance. Rewritten to say exactly that: this reading only exists when somebody runs the probe.
- **Test gap at the seam** (`tests/unit/test_linkedin_live_validation.py`): nothing exercised `probe_connect_dialog` at all, so the new reading keys were asserted only against hand-built dicts — renaming one in the probe would leave every verdict test green while the field silently vanished. Added two probe-level tests: the dialog case records `note_affordance_present is False` / `bare_send_present is True` and says so in its verdict; the no-dialog case records `None` and never even looks for the note button.

Not blocked on: the issue's suggestion to live-ground the quota-spent dialog first. The PR's reasoning stands — both labels were live-grounded 2026-08-03, the change only lowers a log level on a path that already had its fallback in hand, and the one thing a live run would add is precisely what the probe field now captures.

gitchrisqueen:

gitchrisqueen:

gitchrisqueen:

PR #1047, opened 2026-08-04
fix(errors): never warn on the condition a paging loop loops until (closes #1041)

### LINKED ISSUE #1041: fix(errors): RecurringWarning: Selector miss: Load more comments (1x)
## Why

> Selector miss: Load more comments

- First seen: `2026-08-03T13:02:39.225000Z` · last seen: `2026-08-03T13:02:39.225000Z`
- Celery task: `cqc_lem.app.run_automation.sweep_reply_comments`
- SDK: `posthog-python`

[Open the issue in PostHog](https://us.posthog.com/project/475262/error_tracking/019fc7b8-2044-7462-9c08-4812ed8f82bc) — it has the stack trace, the grouped occurrences and the affected people.

## Scope

## Files
- `tests/unit/`

## Acceptance
- All required CI gates pass.

Auto-filed from PostHog Error Tracking. Dedup marker (do not remove): `posthog-issue-019fc7b8-2044-7462-9c08-4812ed8f82bc`

gitchrisqueen:
## What

`_reply_to_comments_on_open_post` expands a post's comment thread by clicking **Load more comments**
up to five times, breaking the moment the control stops resolving. That lookup ran with the default
`warn_on_miss`, so the miss logged `Selector miss: Load more comments` at **WARNING** — and the miss
**is the loop's exit condition**:

- a post whose comments already fit on one page never renders the control at all;
- a post that does renders it until the last page is in, then stops.

Either way every sweep ends on a miss. Three in 24h escalated to ERROR, filed one grouped
`$exception`, and the daily error→issue cron turned "the loop terminated normally" into #1041.

## The change

- `src/cqc_lem/app/run_automation.py` — the `Load more comments` lookup now passes
  `warn_on_miss=False` (DEBUG per `utilities/CLAUDE.md`) and carries `user_id`/`post_id` so the DEBUG
  line still says whose sweep it was. **The `break` is untouched**, so expansion behaviour is
  unchanged.
- `src/cqc_lem/utilities/CLAUDE.md` — names the shape so the next paging loop doesn't repeat it:
  *never warn on the condition you loop until.*

No probe reading is lost here. Unlike the connect-note affordance (#1039), this miss is not
ambiguous evidence of anything — it is the normal terminal state of every run on every post, so
there was never a signal in it to relocate.

## Testing

`tests/unit/app/test_reply_sweep.py`, two new tests on `TestReplyToCommentsOnOpenPost`:

- `test_load_more_miss_never_warns` — the miss breaks the loop after ONE call, the call is made with
  `required=False` + `warn_on_miss=False`, and `log_warning` is never reached.
- `test_load_more_expands_until_the_control_is_gone` — a three-page thread is still clicked through
  to the end, so silencing the miss did not silence the expansion.

```
pytest tests/unit/app/test_reply_sweep.py tests/unit/app/test_lead_signals.py -q
71 passed
```

## Scope check

Issue #1041's acceptance criteria are: the exception no longer occurs (deliberately downgraded to
DEBUG here), a unit test covers the failing path, CI green. Nothing is deferred and no later phase
is declared — the close is honest.

Closes #1041

### DIFF
--- a/src/cqc_lem/app/run_automation.py
+++ b/src/cqc_lem/app/run_automation.py
@@ -4160,12 +4160,16 @@ def _reply_to_comments_on_open_post(driver, wait, user_id: int, post_id: int, my
 
     # SDUI: expand more replies where available, then collect comment items from the
     # new data-testid comment list (comments are no longer article.comments-comment-entity).
+    # The miss IS this loop's exit condition, so it never warns (utilities/CLAUDE.md): a post whose
+    # comments already fit on one page never renders the control, and one that does stops rendering
+    # it once the last page is in — every sweep ends on a miss by design.
     for _ in range(5):
         more = click_first(driver, wait,
                            [(By.XPATH, "//button[contains(@aria-label,'more comment') or "
                                        "contains(normalize-space(),'Load more') or "
                                        "contains(normalize-space(),'more repl')]")],
-                           "Load more comments", required=False)
+                           "Load more comments", required=False, warn_on_miss=False,
+                           user_id=user_id, post_id=post_id)
         if not more:
             break
         time.sleep(2)
--- a/src/cqc_lem/utilities/CLAUDE.md
+++ b/src/cqc_lem/utilities/CLAUDE.md
@@ -85,6 +85,12 @@ Two things follow for anyone writing a call site here:
    readings are identical in the DOM — so it moved to the `--connect-dialog` probe, which reports the
    affordance without grading it (`docs/sdui-probe-coverage.md`). **A miss you cannot attribute
    belongs in a probe, not in a warning.**
+   A **paging loop** is the same test at its most obvious: `_reply_to_comments_on_open_post` clicks
+   'Load more comments' until the control stops resolving, so the miss IS the loop's exit condition
+   and fires on EVERY sweep — a post whose comments fit on one page never renders the control, and
+   one that does stops rendering it once the last page is in. It logged `Selector miss: Load more
+   comments` at WARNING and filed a defect for the loop terminating (issue #1041). **Never warn on
+   the condition you loop until.**
    The rule reads sideways for a **state-setter**: doing what you were asked is not a degraded path,
    however serious the state is. `pause_automation` warned every time it stored the global Selenium
    kill-switch, and maintenance mode sets one on EVERY release, so a routine deploy filed
--- a/tests/unit/app/test_reply_sweep.py
+++ b/tests/unit/app/test_reply_sweep.py
@@ -312,6 +312,39 @@ class TestReplyToCommentsOnOpenPost:
         assert result == {"status": "ok", "summary": "Replied to 1 comments",
                           "comments_found": 1, "replies_sent": 1}
 
+    def test_load_more_miss_never_warns(self):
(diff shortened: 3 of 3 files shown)


gitchrisqueen:

Reviewed the full diff plus the seams it touches: `click_first`/`find_first` in `selenium_util.py` (both miss paths and the `warn_on_miss` threading), the surrounding `_reply_to_comments_on_open_post` body, `utilities/CLAUDE.md`'s escalation contract, and issue #1041's acceptance criteria.

What I checked and could not break:
- **The claim that the miss is unconditional holds.** `if not more: break` is the only exit, and `click_first` returns `None` on both `Selector miss:` and `Click miss:` — so every sweep on every post terminates on a miss. There was no signal in the warning to relocate to a probe (unlike #1039), so nothing is lost by the downgrade.
- **`warn_on_miss=False` reaches both miss paths.** `click_first` forwards it to `find_first` and re-uses it in its own `Click miss:` branch (selenium_util.py:655, :671), so the flag can't be half-applied. `user_id`/`post_id` are supported structured-context fields, so the DEBUG line still attributes the sweep.
- **Only one call site emits this label.** Grepped `src/` for `Load more` / `more comment` / `more repl` — `run_automation.py:4168` is the sole producer, so the grouped `$exception` has no second source left un-fixed. It is also the only paging loop of this shape in `run_automation` / `utilities/linkedin`.
- **Nothing depended on the warning.** No test, script, or `LOG_ESCALATE_EXCLUDE` entry referenced it.
- **The expansion behaviour is genuinely unchanged.** The `break` and the `range(5)` cap are untouched.
- **The tests are load-bearing, not decorative.** I reverted the source change and re-ran: `test_load_more_miss_never_warns` FAILS without it (1 failed, 1 passed), then restored. Full file green on the fix: `pytest tests/unit/app/test_reply_sweep.py -q` → **41 passed**.
- **The close is honest.** #1041's three acceptance criteria (exception gone/downgraded, unit test on the failing path, CI green) are all met; the issue declares no later phase and has no unchecked boxes, so `Closes #1041` is correct with no follow-up owed.

Two non-blocking observations, both **pre-existing and deliberately not changed here** (neither is caused by this diff, and fixing either inside a 5-line error-tracking fix would add regression risk):
- The terminal miss costs ~35s of dead Chrome-session time per post (`MAX_WAIT_RETRY=2` × `WAIT_DEFAULT_TIMEOUT=15s` + a 5s inter-retry sleep in `find_first`), and by this PR's own reasoning it is paid on *every* sweep. `max_try=1` would reclaim it, but it would also shorten the window for a slow-hydrating control on the first iteration — i.e. trade guaranteed latency for a chance of silently skipping expansion. Not worth taking on a fix whose whole virtue is that it changes no behaviour.
- The 'Load more comments' control has no entry in `docs/sdui-probe-coverage.md` — `comment_outcome` covers `_comment_items` / `_switch_comment_sort`, not this expansion. So if the selector rots, reply sweeps quietly see only page one. Worth noting that this is **not a regression**: a warning that fires on every run detected nothing, so no detection existed to lose. Flagging rather than filing an issue, since probe coverage is outside #1041's scope and is the owner's call.

No code changes pushed — the head commit stands as-is.

gitchrisqueen:

gitchrisqueen:

gitchrisqueen:

